PR #590, opened 2026-07-25
Cut PR CI from ~39 min to ~12 min by sharding the backend gate

lindkvis:
## Why

The backend gate ran 39 test files as **22 sequential steps inside one job**. 35.4 of `build`'s 39.3 minutes were tests waiting on each other, and one step — the authz/BOLA gate — was 20.4 minutes on its own.

## Measured before / after

Baselines from run [`30148623406`](https://github.com/Pikselkroken/pixlstash/actions/runs/30148623406) (regular PR) and [`29645768923`](https://github.com/Pikselkroken/pixlstash/actions/runs/29645768923) (rc-prep):

| | before | after (projected) |
|---|---|---|
| `build` | **39.3 min** (authz gate 20.4, `test_server.py` 5.5) | ~12–13 min |
| `build-windows` | **29.8 min** (`test_server.py` alone 13.3) | ~10 min |
| `e2e` | 7.3 min | unchanged |
| **PR critical path** | **~39 min** | **~12–13 min** |

Sizing input: the whole suite single-process measures **65.9 min / 863 passed, 1 skipped**.

## What changed

- **8-way shard matrix.** `--ci-shard i/N` (new hook in `tests/conftest.py`) deals *collected tests* round-robin, so the 20-minute authz block and `test_server.py` get split across runners instead of pinned to one. Windows is sharded 4 ways the same way.
- **`checks` split out** (ruff, arch-doc, wheel, frontend) so lint failures surface in ~4 min instead of behind the test suite. It carries no `HF_TOKEN` — it runs no pytest.
- **Shared setup extracted** to `.github/actions/setup-backend` so the shards cannot drift apart on cache keys.
- **`build` / `build-windows` are now aggregator jobs**, keeping those exact check names stable while the real work runs as `backend shard N`.
- **Log flags** `-q -ra --log-level=INFO` (was `-s -vvv --log-cli-level=INFO`). Capture is on, so a *failing* test still prints its full INFO log; only passing-test noise is gone.
- **Duplicate runs removed.** `push` is filtered to `main`/`develop`/`release-**`/`rc-prep-**`/`v*` tags — every PR-branch commit previously started the workflow twice on the same SHA. The org is on the free plan (**20 concurrent jobs**), so without this the matrix would queue behind a redundant copy of itself. Superseded runs are now cancelled too.
- **One flaky test fixed at the cause.** `test_rate_limit_window_resets` pinned the sliding window to 1s, but two bcrypt logins can exceed that, evicting the first hit before the third arrives (200 instead of 429). Window widened; assertions unchanged.

## Coverage

**Unchanged versus today**, verified mechanically:

```
OLD blocking=39  NEW blocking=40
dropped = EMPTY
added   = ['tests/test_ci_shards.py']
WINDOWS dropped=EMPTY added=EMPTY
partition: gated(40) + deferred(59) = 99 == discovered(99): True
```

The gate is still an allowlist, so drift is stopped a different way: **`tests/` must be exactly the gated set plus an explicit `DEFERRED_FROM_GATE` list**, and `tests/test_ci_shards.py` fails if a file is in neither. A new test file now forces a decision instead of silently missing CI — the bug that would have left PR #588's new test ungated. Deferred files still run in the informational release-prep sweep, and the guardrail also fails if that sweep stops covering the whole suite.

All five invariants were verified non-vacuous: each was broken, watched fail, and restored.

## Security review

**CSO verdict: APPROVE WITH CONDITIONS — condition F2 is fixed in this PR.**

F2: the aggregators accepted `success|skipped` for every dependency. Now discriminated per job — `success` required for `checks`, `backend`, `backend_windows`; `skipped` permitted only for `backend_release_sweep`. Otherwise a future `paths:`/`if:` filter would let a required check pass having run zero tests.

## Follow-ups

- **F1 (do this first): there are no required status checks on `develop`/`main` today.** CI currently blocks nothing on merge. Enable `build` + `build-windows` as required — without that, none of the above gates anything.
- **F3: on the first full run, compare the summed per-shard test counts against the single-process total** (863) to confirm the partition loses nothing in practice.
- F5/F6/F7: minor hygiene noted in review.
- **Deferred: flip the gate to full discovery (`pytest tests/`, no list).** Blocked on the deferred set going green. `tests/test_smart_score_invalidation.py::test_penalised_tag_change_invalidates_atomically_no_second_task` **fails on the current baseline** (`assert 2 == 1`, serial, no sharding — pre-existing, unrelated to this PR) and **has no tracking issue; one should be opened.** ~45% of the deferred set is also still unverified as blocking.

## Not used: pytest-xdist

Added to the test extra for local `-n auto`, but **CI does not use it**. Two attempts to validate the ML-heavy half under `-n auto` ended with every worker aborting in teardown — `terminate called without an active exception`, a C++ thread-pool destructor firing during conftest's model release. A crash with no failing test attached is the worst flake class to ship, so the parallelism is across runners only. `-n 2` is a one-line upgrade once there is a green measured run.

### DIFF
--- a/.github/actions/setup-backend/action.yml
+++ b/.github/actions/setup-backend/action.yml
@@ -1,0 +1,59 @@
+name: Set up the Python backend
+description: >-
+  Reclaim runner disk space, install Python 3.12, restore the pip and
+  HuggingFace model caches, and install pixlstash with its test and dev extras.
+  Shared by every Linux backend job so the matrix shards cannot drift apart —
+  in particular so they all compute the same cache keys and therefore all hit
+  the same warm caches.
+
+runs:
+  using: composite
+  steps:
+    - name: Remove unused tools
+      uses: jlumbroso/free-disk-space@54081f138730dfa15788a46383842cd2f914a1be # v1.3.1
+      with:
+        tool-cache: false
+        android: true
+        dotnet: true
+        haskell: true
+        large-packages: false
+
+    - name: Set up Python
+      uses: actions/setup-python@v6
+      with:
+        python-version: '3.12'
+
+    - name: Cache pip packages
+      uses: actions/cache@v5
+      with:
+        path: ~/.cache/pip
+        key: pip-${{ runner.os }}-py3.12-${{ hashFiles('setup.py', 'requirements.txt') }}
+        restore-keys: pip-${{ runner.os }}-py3.12-
+
+    - name: Install dependencies
+      shell: bash
+      run: |
+        python -m pip install --upgrade pip
+        pip install .[test,dev]
+
+    - name: Cache HuggingFace models
+      uses: actions/cache@v4
+      with:
+        # Two distinct on-disk locations, both required:
+        #  * ~/.cache/huggingface — the HF hub cache used by CLIP/SBert/
+        #    Florence via from_pretrained_local_first.
+        #  * ~/.local/share/pixlstash/downloaded_models — platformdirs
+        #    user_data_dir, where the WD14 and PixlStash taggers land because
+        #    they call hf_hub_download(local_dir=...), which bypasses the hub
+        #    cache entirely. The old key cached only the hub dir and was thus
+        #    a no-op for the taggers that actually 429'd.
+        # A warm cache makes needs_download() (a pure file-existence check)
+        # return False, so download() — and its HEAD revalidation, the call
+        # that returns 429 — is never issued. Key off the tagger plugins so a
+        # model/revision bump busts the cache; restore-keys keeps it warm
+        # across unrelated commits.
+        path: |
+          ~/.cache/huggingface
+          ~/.local/share/pixlstash/downloaded_models
+        key: hf-models-${{ runner.os }}-${{ hashFiles('pixlstash/tagger_plugins/wd14.py', 'pixlstash/tagger_plugins/pixlstash_tagger.py', 'pixlstash/tagger_plugins/florence2.py') }}
+        restore-keys: hf-models-${{ runner.os }}-
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,75 +4,60 @@ name: CI
 permissions:
   contents: read
 
+# A superseded run has nothing left to prove, and its runners are exactly what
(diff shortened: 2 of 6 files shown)


lindkvis:
## First real run: green, and ~3x faster as projected

Run [`30153127026`](https://github.com/Pikselkroken/pixlstash/actions/runs/30153127026) — **all checks pass.**

| job | duration |
|---|---|
| `backend shard 1..8` | 9m03s – **10m52s** |
| `backend-windows shard 1..4` | 10m27s – **13m10s** |
| `checks` | 5m27s |
| `e2e` | 8m03s |
| `build` / `build-windows` (aggregators) | 2s / 4s |
| `backend-full` | `skipping` (correct — not release-prep) |

**PR critical path: ~39 min → 13m10s.** Projection was ~12–13 min.

## F3 closed — the partition is exact, measured not argued

Per-shard collection from the logs:

```
shard 1: 62 passed, 430 deselected     shard 5: 61 passed, 431 deselected
shard 2: 62 passed, 430 deselected     shard 6: 61 passed, 431 deselected
shard 3: 62 passed, 430 deselected     shard 7: 61 passed, 431 deselected
shard 4: 62 passed, 430 deselected     shard 8: 61 passed, 431 deselected
```

- Every shard collected **492** tests (62+430 = 61+431 = 492) — identical collection across all 8, so the sharding is deterministic.
- Tests actually run: 4x62 + 4x61 = **492**.
- **492 run == 492 collected.** Nothing dropped, nothing run twice.
- Shard sizes differ by at most 1 test, which is what round-robin promises.

## Two observations for later

1. **Windows is now the critical path** (13m10s vs Linux 10m52s) and is the less balanced of the two (10m27s–13m10s). A 5th Windows shard is the cheap next win.
2. **Setup is now ~40% of a shard.** Test execution is 4m58s–6m36s inside ~10m jobs, so the remaining ~4 min is checkout/install/cache — `Remove unused tools` alone was ~1.4 min historically. Further gains come from trimming setup, **not** from adding shards; past ~8 shards the fixed cost dominates.

F1 (no required status checks on `develop`/`main`) is still the one that matters most — none of this gates anything until `build` + `build-windows` are marked required.

PR #626, opened 2026-07-28
Operation log with metadata undo/redo (DAM 1.2)

lindkvis:
Lane B of v1.9.0. Append-only `operation` table (migration 0086, conditional creation) with `batch_id` from day one so any bulk action is one undoable operation. Undo scope = DAM 1.2 metadata facets only (tags, description/caption, score, set/project membership, character assignment, stacking).

- State-based design: records before/after metadata of affected pictures, not per-endpoint inverses; capture → mutate → capture → record runs in one queued DB task for atomicity.
- 7 new routes (list, get, undo-state, undo/redo, per-op undo, per-batch undo), all owner-only in ROUTE_POLICIES with justifications; a test pins that the module has no handler-level authz.
- Origin discipline: undo emits origin in the event `data`, never from the contextvar — pinned by an AST test and a new ws-broadcaster producer-side test.
- Registry arithmetic re-verified (0 undeclared / 0 dead); coverage matrix updated.

252 targeted tests passed across op-log, migrations, authz gate, ws broadcaster, and touched mutation modules; ruff clean.

Known judgement calls (see review notes): locked-set undo returns 423 refuse-and-report; sentinel-tag round-trip on undo re-enters the retag queue; redo invalidation is global (single-owner correct).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -227,6 +227,7 @@ jobs:
           tests/test_metadata_hash.py
           tests/test_migrations.py
           tests/test_openapi_response_schemas.py
+          tests/test_operation_log.py
           tests/test_picture_mutation_scope.py
           tests/test_picture_plugins.py
           tests/test_picture_set_locking.py
--- a/docs/backend_architecture.md
+++ b/docs/backend_architecture.md
@@ -31,6 +31,7 @@
 18. [Snapshots & Restore](#18-snapshots--restore)
 19. [Mermaid Diagrams](#19-mermaid-diagrams)
 20. [Architectural Patterns](#20-architectural-patterns)
+21. [Operation Log](#21-operation-log--undoredo-and-the-audit-trail-dam-12)
 
 ---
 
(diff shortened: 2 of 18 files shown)


lindkvis:
CSO sign-off (PR #630): **APPROVE.** X1 applies: this branch and #629 both mint migration 0086 off head 0085 — whichever merges second must renumber or boot fails with two Alembic heads. Informational: N1 (undo can restore set/project membership a share token was revoked from via entity-move — document that revocation = token deletion) and N2 (unscoped READ tokens can read /operations*, consistent with existing precedent).

lindkvis:
CSO final certification: **APPROVE, unconditional.** Merge-order note stands: renumber migration 0086 if #629 merges first; last merger re-derives the coverage-matrix totals.

lindkvis:
CI fixed: the shard-1 failure was the test-classification guardrail — new test files weren't registered in the backend gate's file list. Registered (CSO-cleared workflow edit); test_ci_shards green locally.

PR #638, opened 2026-07-29
Dedup → Stacks backend: tiered detection, verdicts, queue APIs

lindkvis:
Lane 1A of the Dedup → Stacks feature (design: `design/1.9-dedup/`, committed here). Implements the owner's design contract:

- **Tier 1 exact**: groups on existing indexed `(pixel_sha, size_bytes)` — documented honestly as a sampled digest with the consent-dialog mitigation; `MissingPixelShaFinder` backfills NULLs. **Tier 2**: bucketed dHash XOR/popcount (size-bin / capture-minute / import-batch / folder buckets, sharded at 4000, never dropped), streamed via the task system. **Tier 3**: existing likeness edges, weakest-link confidence.
- `TierPolicy` (exact locked on, opt-in descent, 0.90 default, 0.65 hard floor) + the design's cover formula with why-pills in both directions.
- Verdict memory (`dedupverdict` keyed on group signature) + group cache + scan rows; migration `0088` (chain 0086 → 0087 → 0088, verified linear).
- Verdicts apply the metadata union (tags, set/project membership, score lifted; characters only in the unambiguous single-character case — documented partial) and record ONE op-log operation per verdict; bulk auto-stack = one batch id = one Ctrl+Z, proven by tests. Keep-separate/reopen deliberately record nothing (no reversible facet changed; a no-op row would eat a Ctrl+Z).
- Undo snapshots the stack-expanded set — fixes a real stranded-sibling bug found during integration, pinned by a deliberately de-vacuized two-stack-fold test.

8 new routes, all `owner_only` with content-identity justifications; coverage matrix re-derived to 236 declared / 235 mounted. 190 targeted tests green; ruff clean; new test files registered in the CI gate.

**Merge gated on independent CSO sign-off** (author cannot certify own authz declarations) — running now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -218,6 +218,9 @@ jobs:
           tests/test_comfyui_workflow_extraction.py
           tests/test_dedup_sweep_api.py
           tests/test_dedup_sweep_service.py
+          tests/test_dedup_tier_service.py
+          tests/test_dedup_tiers_api.py
+          tests/test_dedup_verdict_service.py
           tests/test_detections_scope.py
           tests/test_facial_features_worker.py
           tests/test_find_orphaned_files.py
--- a/design/1.9-dedup/CONTRACT.md
+++ b/design/1.9-dedup/CONTRACT.md
@@ -26,9 +26,11 @@ with a one-time notice pointing at the new sidebar entry.
 ## Design-over-shipped-backend deltas (backend adapts)
 
 1. **Cover selection formula** replaces the shipped planner's keeper order
-   (score → smart score → recency): `pixels×4 + tags×3 + userScore×2 + RAW
-   bonus`, ties break to **oldest capture time**. Always a visible
-   preselection the user can override (1–9); never silent.
+   (score → smart score → recency): `megapixels×4 + tags×3 + userScore×2 +
+   RAW bonus`, ties break to **oldest capture time**. Always a visible
+   preselection the user can override (1–9); never silent. (The pixel term is
+   **megapixels**, not raw pixel count: at raw counts it would dwarf every
+   other term and the formula would be "biggest file wins" with decoration.)
 2. **Policy model** replaces `SweepPolicy`'s auto/review split semantics with
    tier gating: Tier 1 exact is always included and cannot be switched off;
    each looser tier is a separate opt-in with its own live count and enabling
@@ -49,6 +51,14 @@ with a one-time notice pointing at the new sidebar entry.
    membership onto the stack and takes the highest score. Nothing is
    overwritten or lost. (Reconcile with existing stack semantics; if current
    "Stack groups" behaviour differs, the design's union wins.)
+   **Characters carry one carve-out.** A face carries a bbox and an embedding
+   that belong to one specific picture, so a true face-to-character union
+   would mean fabricating `Face` rows. When the group's members between them
+   reference exactly **one** character, members that lack it get
+   `pending_character_id` (the shipped deferred-assignment path). A group
+   spanning **several** characters is left alone and logged — inventing
+   detection data is worse than not unioning. Tags, sets, projects and score
+   union unconditionally.
 
 ## Performance rules (design §1, binding)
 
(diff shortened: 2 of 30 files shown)


lindkvis:
CSO sign-off (docs/reviews/dedup-tiers-authz-signoff.md): **APPROVE-WITH-REQUIRED-CHANGES** — the authz surface held (all 8 routes fail closed for scoped, unscoped-READ and query-param tokens; no sibling leaks; matrix reproduces at 236 declared / 235 mounted; the two-stack-fold undo test verified non-vacuous by independent mutation). Required before merge: **R1** `content_key` must include `size_bytes` (signature currently non-injective — two distinct exact groups collapse to one signature, and keep-separate silences both); **R2** `bulk_auto_stack` must catch the 423 lock path and always return the minted `batch_id` (currently a partial bulk apply loses its undo handle); **R3** verdict handlers must pass `request_context(request)` so operations carry the actor/source. Plus §22.8 share-token-widening doc note (accepted risk A1, pre-existing class) and stale §22.9. Also to fix from the resource pass: non-numeric `scope_id` 500s (and `POST /dedup/scan` persists the poison row) → validate to 400; Tier-2 pair accumulation is memory-unbounded within a bucket (O(k²), ~582MB worst case at the cap) — bound it. Fixes dispatched; CSO re-verifies before merge.

lindkvis:
Heads-up: the branch was force-pushed with a rebased history this morning (migration renumbered to 0088 + migration-table row — duplicating what the merge-based update had also done). The CSO fix commit has been transplanted onto that rebased head as 029525a0 (conflicts were stale-vs-fixed content only; the rebase's migration-table rows are preserved; 63 verification tests + ruff green). **CSO required changes applied**: R1 signature keys on (pixel_sha, size_bytes) with the reviewer's E1 repro as a durable test + 0088 amended to clear stale rows; R2 bulk auto-stack skip-and-report with per-group outcomes, always returns batch_id, plus a rollback fixing a partial-write hazard the abort had masked; R3 actor/origin on the two recording handlers (keep-separate/reopen record nothing — flagged for reviewer adjudication). Resource hardening: scope validation to 400 (no poison rows), LIKE wildcards escaped, Tier-2 memory bounded (50k/bucket, 400k/scan, honest warnings), incremental persist. Punch-list: thumbnail_version via shared cache-token helper; race-free dry_run_summary. CSO re-verification running.

lindkvis:
QA cross-layer verdict (live server, real repros): **two merge blockers** — (1) CRITICAL: undo/batch-undo reverts picture state but never reopens the DedupVerdict/DedupGroup rows, so the group is permanently lost from the queue (survives rescan); (2) HIGH: exact-tier groups silently downgrade to 'near' when a near-enabled scan re-detects them (unconditional tier overwrite on upsert), starving the exact-only view and auto-stack. Also: **keyset cursor now REQUIRED** — QA reproduced a deterministic single-request skip under offset paging. Fix round running (also folds in CSO M1/W1/W2 + doc lows). CSO status: APPROVE unconditional on the prior scope; final confirm after this round.

lindkvis:
QA blockers + CSO residuals fixed (5 commits): undo/redo now reopen/re-resolve the verdict via a generic post-restore hook in the op-log (one transaction, one hook call per batch — a raising hook aborts the restore); tier precedence on upsert (exact never downgraded); keyset cursor (`cursor`/`next_cursor`, versioned encoding, exact float round-trip for the tie-break; offset deprecated, cursor+offset=400); M1 batch-id namespacing unified (srv- minting shared with the op-log, cli- validation on all four verdict bodies); W1 comment honesty (the cap CAN split membership in dense non-uniform blocks); W2 slash-only folder scopes refused. Every fix reverted-and-watched-fail before restore. 258 targeted tests green. CSO final confirm + QA blocker re-verification running.

lindkvis:
CSO final certification (review closed, docs/reviews/dedup-tiers-authz-signoff.md @ 318dea99): **APPROVE at b6b8091d, no open required changes.** R5 verified with both defenses tested independently (incl. forcing a batch id into the DB to defeat defense 1 — the verdict filter alone still held, which also covers legacy rows); R6 sweep extended to 1e400-style overflow literals; W1/A1 round-3 'still open' claims corrected as reviewer error (both were already fixed — correction recorded in the doc). Accepted risks A1/A2 documented with revisit triggers. **Merge gate released.** Merge order: #638 → #640.

PR #642, opened 2026-07-29
Scrapheap undo: restored events, no NEW badge, ghost tiles

lindkvis:
Fixes the three owner-reported issues in the Scrapheap-undo flow:

1. **Grid refresh**: new `change_kind: "restored"` (backend gate + SPA mirror moved together — both ends silently degrade unknown kinds, so all four consumer sites changed in one commit). Own-client restores reinsert immediately; external clients keep the change-pill flow.
2. **Sidebar NEW badge**: restores no longer masquerade as `"added"`, so `markCountNew` and the grid's new-image highlight ignore them; true imports unchanged.
3. **Ghost tiles per the design**: deleted tiles stay in place, dimmed + two-tone hatched (all existing tokens, zero new DOM elements), for exactly the receipt's 8s destructive dwell — one clock, owned by the operation store, so hover/focus/hidden-tab pauses apply to the tiles for free. Receipt expires or is replaced → tiles collapse with scroll anchoring; undo while live → un-ghost in place, no flash; undo after collapse → the restored-reinsert path; lightbox undo un-ghosts too; redo re-ghosts symmetrically.

ui-ux pass: cursor skips ghosts (inert cells with a cursor make following keys silently dead), `inert` semantics, no context menu/lightbox/drag on ghosts, collapse-with-scroll-compensation as a ship condition. Lead-designer pass: hatch placement + two-tone self-contrast, no new tokens, ruled the text chip unnecessary.

Also fixed en route: `buildReceipt` dropped `undoable`, which would have committed every ghost set instantly.

Backend 178 targeted tests, frontend 1111 (28 new), e2e pins 10 passed twice, ruff/lint/build clean.


lindkvis:
Owner rulings on the three review-pass questions: (1) timed collapse at receipt expiry stands, definitively — no interaction-based alternative; (2) ghost/receipt hover-linkage parked for later consideration; (3) instantaneous collapse confirmed, no exit fade. Shipped behavior matches all three; no changes needed.

PR #644, opened 2026-07-29
Group one gesture's operations with a client batch id

lindkvis:
Stacked on #641. Completes the tag-undo fix: a compound gesture (chip delete = remove_all + reject; bulk confirm fan-out) now shares one `X-Operation-Batch-Id`, so a single Ctrl+Z restores the tag AND the ledger together — pinned by an end-to-end test asserting both rows revert in one undo call.

- Header validated in the origin middleware exactly like `X-Client-Id`: strict `cli-` namespace + URL-safe charset, anything else dropped with a debug log (never a 500, never truncated). Server-minted ids now carry `srv-`, so a client can never name or join a server batch (pinned by a test that feeds the real minting function into the real validator).
- `request_context` carries the batch id; the scrapheap sites keep their behavior via `fallback_batch_id`.
- Frontend: one `newOperationBatchId()` per gesture across the four fan-out sites; real-DOM tests assert 4 requests / 1 id on drag-to-rejected and distinct ids across separate gestures; the receipt narrates a compound gesture as one entry.
- §21.2 documents the A2 stance: a client batch id is a grouping hint scoped to the owner's own history — grouping never widens what an operation may touch.

Backend 101 targeted tests, frontend 1092, ruff/lint/build clean. Independent adversarial review of the namespace guard is folded into the running CSO pass (author does not self-certify).


lindkvis:
CSO G1 applied: the sanitizer uses fullmatch, so `cli-abcd\n` (and any trailing control character) is rejected — pinned in the malformed-header sweep. CSO verdicts so far: #638 APPROVE unconditional; #644 approved pending this fix; M1 (cross-branch batch-id namespace unification in the dedup verdict body path) lands in the dedup fix round now being dispatched.

lindkvis:
CSO final certification: **APPROVE at cf3786cc.** G1 verified directly (fullmatch rejects trailing/mid-string control characters; srv- still refused; legit ids and the 76-char max unaffected). M1 unification onto request_origin.py stays a follow-up, not a gate — two copies that agree are debt, not a hole. Merge gate released.

PR #649, opened 2026-07-30
Mark undecodable images in thumbnail and quality tasks too (#585)

### LINKED ISSUE #585: Invalid Images break tasks
**Describe the bug**
If a reference folder contains invalid/corrupt/etc images, the tagging (and other task) systems seem to endlessly retry opening the invalid image file(s). This spams the console with warning messages and may be stopping other background tasks from starting.

**To Reproduce**
1. Create an empty file named "image.jpg" (or use a corrupt jpeg file) in a new folder.
2. Add the new folder as a reference folder.
3. Observe the server console log.

**Expected behavior**
After an image fails to be opened, it should be ignored for the remaining duration of the server processes lifetime, or until it is modified (to handle cases where the image was read while it was still being written to.)

lindkvis:
Fixes #585 (follow-up reported against v1.8.0).

## Problem

The v1.8.0 fix added the `UnprocessableImageRegistry` and made every batch finder skip suppressed pictures at the shared `_filter_and_claim` choke point — but only `ImageEmbeddingTask` ever **marked** a picture as unprocessable.

`ThumbnailGenerationTask` logged a warning on a failed decode and returned without writing any columns, so `thumbnail_width` stayed NULL and `MissingThumbnailFinder` re-selected the same corrupt picture on every sweep — the original endless-retry bug. It surfaces prominently in 1.8.0 because the upgrade resets the thumbnail columns for a one-time regeneration sweep, pushing every corrupt image in the library through this finder. Suppression via the embedding task only masks it until the next restart (the registry is in-memory by design).

`QualityTask._backfill_missing_picture_metadata` had the same shape one step worse: it **raised** on the first undecodable file, aborting the whole batch — good pictures included — and leaving every column unset for the next sweep.

## Fix

- `ThumbnailGenerationTask._resolve_columns` marks the picture in the registry when the source exists but cannot be decoded, matching the embedding task's semantics (only genuine decode failures; missing files stay with the missing-file purge finder). The registry's own one-per-file-version warning replaces the per-sweep log spam.
- `QualityTask._backfill_missing_picture_metadata` marks the picture and continues instead of raising, so the rest of the batch is processed and `_compute` writes its usual sentinel quality values for the unloadable image.

`TextScoreTask` was checked and already terminates on its own (writes a `-1.0` sentinel).

## Tests

`tests/test_invalid_image_endless_retry.py` (the #585 regression file) previously covered only the embedding path — which is how this was missed. Added:

- `test_585_thumbnail_task_suppresses_corrupt_image` — drives the real `MissingThumbnailFinder` → `ThumbnailGenerationTask` → finder loop against a corrupt file and asserts the finder stops re-selecting it.
- `test_585_quality_metadata_backfill_marks_instead_of_raising` — corrupt picture first in the batch; asserts no raise, the corrupt picture is suppressed, and the valid picture behind it still gets its metadata backfilled.

All 5 tests in the file pass, plus the thumbnail/quality-adjacent suites (49 tests) and ruff.

## Follow-up: stuck progress bar after restart

Manual testing found that after a restart the "Upgrading thumbnails" progress bar never completed: the corrupt picture was correctly re-marked and skipped, but `Vault._count_missing_thumbnails` still counted every `thumbnail_width IS NULL` row, so `remaining` stalled at the number of undecodable files. It now excludes active suppressed ids, the same way `_count_missing_image_embeddings` already did.

`MissingThumbnailFinder`'s candidate query got the same exclusion so a large run of low-id corrupt pictures cannot fill the fetch window and starve real work behind it — mirroring the embedding finder.

The thumbnail regression test now also asserts the progress count goes 1 → 0 across the suppression. Other progress counts were checked: quality/text-score terminate via sentinel values, face extraction inserts a sentinel `Face` row even for unloadable images, and embeddings were already covered.

### DIFF
--- a/frontend/src/components/views/ImageGrid.vue
+++ b/frontend/src/components/views/ImageGrid.vue
@@ -218,8 +218,8 @@
         </v-card-title>
         <v-card-text style="padding: 0 20px 4px">
           <div style="font-size: 0.875rem; opacity: 0.85; margin-bottom: 10px">
-            Leave the label empty for dense object detection, or type a phrase to
-            detect only that (e.g. "dog").
+            Leave the label empty for dense object detection, or type a phrase
+            to detect only that (e.g. "dog").
           </div>
           <v-text-field
             v-model="segmentPrompt"
@@ -231,7 +231,9 @@
           />
         </v-card-text>
         <v-card-actions style="padding: 8px 16px 16px">
-          <v-btn variant="text" @click="segmentDialogOpen = false">Cancel</v-btn>
+          <v-btn variant="text" @click="segmentDialogOpen = false"
+            >Cancel</v-btn
+          >
           <v-spacer />
           <v-btn color="primary" variant="tonal" @click="confirmSegment">
             Detect
@@ -266,7 +268,9 @@
       v-model="snapshotsWithDeletedOpen"
       :snapshots="snapshotsWithDeleted"
       :dont-show-again="userPrefsStore.hidePurgeSnapshotWarning"
-      @update:dont-show-again="userPrefsStore.setHidePurgeSnapshotWarning($event)"
+      @update:dont-show-again="
+        userPrefsStore.setHidePurgeSnapshotWarning($event)
+      "
     />
     <DeleteForeverDialog
       v-model:open="deleteForeverOpen"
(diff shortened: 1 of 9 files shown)


lindkvis:
I've tested it with both a broken and an empty jpg file. It really should work now.

PR #682, opened 2026-07-31
chore: release v1.8.2rc1 website update

lindkvis:
Automated website update triggered by release **v1.8.2rc1**.

Changes included:
- `website/api/v{major}.{minor}/` — versioned OpenAPI docs (Scalar UI)
  generated from the release tag's own backend code. The generator
  also heals any older version pages still on ReDoc and drops the
  legacy top-level `openapi.json`.
- `website/latest-version.json` — updated to this release (if it is
  strictly newer than the currently published version; skipped otherwise).
  The `[Security: ...]` tag was taken from the release tag's own `CHANGELOG.md`.
- `website/latest-version/{docker,pip,electron,other}.json` — per-install-type
  manifests carrying the **identical** payload as the legacy file. These
  back the PATH-encoded version-check URLs so Cloudflare zone analytics can
  bucket active installs by type (docker vs pip vs electron vs other).
  Updated/skipped in lock-step with the legacy `latest-version.json` above.


lindkvis:
Prerelease rehearsal for the Certum signing pipeline — the public v1.8 API docs should keep advertising 1.8.1, not an rc. Same handling as #657.

PR #765, opened 2026-08-08
Stop Tab trapping focus in the tag panel (#755)

### LINKED ISSUE #755: TbTagPanel: Tab is a keyboard trap after a failed tag write (WCAG 2.1.2, Level A)
Found during the v1.10.0 model shelf design review. This is a **WCAG 2.1.2 (No Keyboard Trap), Level A** failure in shipped code.

## Mechanism

`frontend/src/components/panels/TbTagPanel.vue`:

- `handleTagKey` (L902-907) calls `preventDefault()` on `Tab` whenever `tagSuggestions.length > 0`. There is no `else` branch and no second-Tab escape.
- `selectTagSuggestion` (L882-886) sets `tagInput` to the tag text and calls `applyTag()` on nextTick, which writes to **every selected picture**.
- The suggestion filter is `startsWith` (L767), so an exact tag still matches itself and the list stays non-empty after selection.
- `applyTag` only clears `tagInput` on the success path (L945). On the error path (catch, L949-951) it does not.

So after a failed write: the field never clears, the list never empties, every `Tab` re-fires the failing request, and **focus can never leave the field by keyboard**.

## Also on the same control

- `Escape` (L908-915) closes the entire panel rather than dismissing the suggestion list first.
- No `role="combobox"`, `aria-expanded`, `aria-activedescendant`, `aria-autocomplete` or live region anywhere on it.

## Fix

Separate fill from commit:

- `Tab` fills the field from the highlighted suggestion and does **not** commit.
- `Enter` commits.
- First `Escape` dismisses the list, second `Escape` leaves the field or closes the panel.
- `Tab` with the list dismissed always moves focus.
- Clear `tagInput` in a `finally`, not only on success.

## Why it matters beyond the tag panel

The v1.10.0 model shelf plans to extract this control as `AppCombobox` and use it for `Set base model for 40 adapters`. Commit-on-Tab there is a 40-row bulk write with no review step. The extraction must change these semantics, so fixing it here is a prerequisite rather than a parallel task.

## Verification

Keyboard-only: type a tag prefix, force the write to fail, then press `Tab` repeatedly. Focus must leave the field.

lindkvis:
Fixes #755.

`handleTagKey` called `preventDefault()` on every `Tab` while the suggestion list was non-empty, and `Tab` committed the highlighted tag to every selected picture. Because the filter is `startsWith`, the completed tag still matched itself and the list never emptied, so `Tab` could never move focus. On a failing write the field also never cleared, so each `Tab` re-fired the failing request. WCAG 2.1.2 (No Keyboard Trap), Level A.

## Change

- `Tab` completes the field from the highlighted suggestion and dismisses the list. It no longer commits.
- `Enter` commits. Clicking a suggestion still fills and commits, since that is an explicit choice.
- First `Escape` dismisses the list, second closes the panel. `ArrowDown` reopens a dismissed list.
- `Tab` with the list dismissed falls through, so focus always leaves the field.
- The control now exposes `role="combobox"`, `aria-expanded`, `aria-controls`, `aria-autocomplete`, `aria-activedescendant`, a `role="listbox"` popup with `role="option"` children, and a polite status region. The error and success lines got `role="alert"` / `role="status"`.

Dismissal is one ref: the input value the list was last dismissed for. Typing changes the value and reopens the list with no extra watcher, while completing the field leaves it dismissed even though the completed tag still matches the filter.

## Not done

The issue asked for `tagInput` to be cleared in a `finally` rather than only on success. Left as is on purpose: clearing on failure discards the tag the user typed and leaves an error message naming a tag no longer on screen. The trap is closed by the `Tab` semantics, so the clear is not load-bearing.

## Verification

`frontend/src/components/panels/TagPanelKeyboardTrap.test.js`, 6 cases. Confirmed to fail 5/6 against the unfixed component. Full panel suite green (122 tests).

### DIFF
--- a/docs/frontend_architecture.md
+++ b/docs/frontend_architecture.md
@@ -409,6 +409,28 @@ The tab/category switch is **stateless** (see Key Design Principles). Concretely
   pictures on a global view → switch to the Projects/People/Sets tab → drag the
   selection onto a project or character to add them, with the global view still
   intact underneath.
+- **A drop target judges the payload KIND during dragover, from `types` alone.**
+  The JSON body is protected while a drag is in flight (`getData()` returns `""`
+  in Chrome and Firefox), so the kind travels as the *key*: every internal drag
+  writes `application/json` **plus** a marker type —
+  `application/x-pixlstash-pictures` or `application/x-pixlstash-faces`
+  (`utils/media.js`: `setInternalDragPayload`, `isPictureDrag`, `isFaceDrag`).
+  A new payload kind adds a marker; it does not add a field to the body.
+  Two rules follow, and both were once broken (issue #757):
+  - **Never `@dragover.prevent` on a drop target.** The modifier calls
+    `preventDefault()` before the handler body and regardless of what it
+    decides, which accepts every drag on the page. `preventDefault()` belongs
+    inside the handler, only for the kinds that row takes — `SideBar.acceptDrop`
+    returns `accept` / `reject` / `ignore` (`ignore` = an external file drag the
+    window-level importer still owns, so the row stays unpainted rather than
+    promising a refusal it will not perform).
+  - **A drop handler keys off `data.type`, never off the presence of
+    `imageIds`.** A face payload carries `imageIds` too (the pictures the faces
+    were found in), which is how face drags used to file themselves into sets.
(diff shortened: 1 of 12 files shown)


lindkvis:
Reopening from `fix/issue-755` to match the requested branch name.

PR #772, opened 2026-08-08
Re-land: suspend SQLite FK enforcement during migrations

lindkvis:
Straight cherry-pick of `b9fc9ca9`, unchanged. It is the fix reviewed in #771, which never reached `develop`.

## What happened

#771 was stacked on #767's branch. #767 merged into `develop` at commit `fa5502e4`; #771 then merged into `fix/grid-backend-url-default` **after** that, producing `132a24ca` on a branch nobody merges again. GitHub marks #771 "merged", so it looks landed, but `origin/develop:pixlstash/migrations/env.py` still has no `_foreign_keys_suspended`. That was my sequencing error in stacking it.

## Why it still matters

Without it the `e2e` lane cannot start its backend at all:

```
sqlalchemy.exc.IntegrityError: (sqlite3.IntegrityError) FOREIGN KEY constraint failed
[SQL: DROP TABLE picture]
Error: Process from config.webServer was not able to start. Exit code: 1
```

`op.batch_alter_table` rebuilds a table by copy-drop-rename, and the vault engine runs `PRAGMA foreign_keys=ON`, so migration 0080 fails on any database that holds pictures — the committed e2e fixture and every real library. Fixed in `migrations/env.py`, the one chokepoint, per SQLite's documented table-rebuild procedure, with `PRAGMA foreign_key_check` afterwards so it cannot mask real damage.

Re-verified on top of the current `develop`: `test_a_populated_library_upgrades_through_the_picture_table_rebuild` passes, and fails without the `env.py` change.

### DIFF
--- a/pixlstash/migrations/env.py
+++ b/pixlstash/migrations/env.py
@@ -3,11 +3,12 @@ from __future__ import annotations
 import logging
 import os
 import sys
+from contextlib import contextmanager
 from logging.config import fileConfig
 from pathlib import Path
 
 from alembic import context
-from sqlalchemy import engine_from_config, pool
+from sqlalchemy import engine_from_config, pool, text
 from sqlmodel import SQLModel
 
 # Import models to register SQLModel metadata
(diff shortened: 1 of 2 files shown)


lindkvis:
Superseded by #775, which carries this commit unchanged. None of the three develop failures can go green in isolation, so they are verified together there.

Issue #757, opened 2026-08-08
Sidebar drop targets accept any drag payload; face-bbox drags already leak into sets


lindkvis:
Found during the v1.10.0 model shelf design review. Two related defects in `frontend/src/components/panels/SideBar.vue`.

## 1. Rows accept any drag on the page

Character rows (L5476, L5968), set rows (L5623, L6176) and project rows (L5753) bind `@dragover.prevent`.

The `.prevent` modifier calls `preventDefault()` **before the handler body runs and regardless of what the handler decides**. Those rows therefore accept any drag payload and paint the full `droppable` highlight (`SideBar.css:1270`, primary fill) for payloads they will do nothing with.

`handleReferenceFolderDragOver` (L2993) already does this correctly: it calls `preventDefault()` inside the handler, only for accepted types.

## 2. Face-bbox payloads already leak into sets

All internal payloads share one MIME key, `application/json`. The discriminator (`data.type`, values `image-ids` / `face-bbox`) lives inside the JSON body, and `getData()` returns `""` during dragover under Chrome and Firefox protected mode. The file documents this itself at L3605-3607; only `types` is readable, which is what `isInternalImageDrag` (L2977) uses.

`handleDropOnSet`, `onProjectDrop` and `handleReferenceFolderDrop` ignore `data.type` entirely and read `data.imageIds`. The face-bbox payload also carries `imageIds`, so a face drag currently drops into a set as if it were a picture drag.

## 3. There is no invalid-drop state in the product

No class, no `dropEffect = "none"`, no cursor change. Rejection is a post-drop toast (`showNotice` L2947, `noticeStore.error` L3035). A user sees a valid-looking highlight, drops, and gets an error afterwards.

## 4. No enter/leave depth counter on sidebar rows

The grid has `dragOverlayDepth` in `useGridDragDrop.js`; sidebar rows do not. Moving the pointer from a row's label into a child element flickers the highlight off.

## Fix

- Remove `.prevent` from `@dragover` on L5476, L5623, L5753, L5968, L6176; call `preventDefault()` inside each handler only for accepted types.
- Give each payload a distinct MIME key so `types` alone is a sufficient discriminator during dragover.
- Check `data.type` in every drop handler rather than assuming `imageIds` implies pictures.
- Add a `.not-droppable` state using a glyph plus reduced opacity, not colour alone.
- Add a shared enter/leave depth counter to sidebar rows.

## Why it is worth fixing now

[...]

PR #775, opened 2026-08-08
Make develop green: three independent CI failures

lindkvis:
One PR because none of the three can be proved in isolation. `develop` has three unrelated failures at once, so a PR that fixes any one of them still shows red for the other two, and merging on a red check proves nothing. Combined, the run can actually go green.

Each commit stands alone and is separately reviewable. Previously opened as #772, #773 and #774, now closed in favour of this.

---

### 1. `e2e` — the backend cannot start (`5de4d7ae`)

```
sqlalchemy.exc.IntegrityError: (sqlite3.IntegrityError) FOREIGN KEY constraint failed
[SQL: DROP TABLE picture]
Error: Process from config.webServer was not able to start. Exit code: 1
```

`op.batch_alter_table` rebuilds a table by copy-drop-rename, and the vault engine runs `PRAGMA foreign_keys=ON`, so migration 0080 fails on any database holding pictures — the committed e2e fixture (`0049_snapshots`, 111 pictures) and every real library. A **fresh** database has nothing referencing `picture`, which is why `test_alembic_upgrade_head_fresh_db` stayed green throughout. 36 migrations use batch mode, so this is a class, not one migration.

Fixed in `migrations/env.py`, the chokepoint every caller routes through, per SQLite's documented table-rebuild procedure, with `PRAGMA foreign_key_check` afterwards so suspending enforcement cannot mask real damage.

*(This is the fix reviewed in #771. It never reached `develop`: #771 was stacked on #767's branch and merged into it after that branch had already gone to `develop`. Marked merged, never landed. My sequencing error.)*

### 2. `backend shard 1` — `DID NOT RAISE HubPermissionError` (`62316d80`)

The test builds its "loose" directory with `mkdir(mode=0o775)`. That mode is masked by the process umask; runners default to 022, so it lands at **0755**, isn't group-writable, and the code correctly declines to raise. The production code was never wrong — the fixture failed to create the condition it asserts on. It passes on an Ubuntu workstation because the default umask there is 002.

`chmod` isn't umask-masked. Verified: `tests/test_hub_bootstrap.py` is 46 passed under **both** `umask 022` and `umask 002`; before the change it is red under 022 and green under 002, which is exactly the CI-versus-local split.

### 3. `backend shard 3` — `assert [True, True, …] == []` (`948fa47a`)

The production behaviour is correct; the assertion was. The test claimed **nothing in the process** read the vault during the request. Instrumenting the observer with the callable and thread shows all 15 reads come from the WorkPlanner thread probing for work:

```
DURING REQUEST: []
IN THE NEXT 2s: [('WorkPlanner', 'QualityTask.count_missing_quality'),
                 ('WorkPlanner', 'MissingLikenessFinder._likeness_state'), …]
```

The runner's call took 0.63 s, long enough to catch probes; this machine finishes first. Timing-dependent from the day it was written.

Now it observes the guest lookup by name — the claim in the test's own name, and the same `_lookup_by_token` sentinel `test_writer_waits_for_request_paused_in_guest_lookup` waits on, so a rename cannot quietly defang it. Checked it is not vacuous: on the positive path (token matching the **active** library) the same request returns 200 and the observer does record `_lookup_by_token`.

---

## Verification

Each fix was reproduced first and each fails without its change. The full backend gate list is running locally under `umask 022` to match the runner; I will report the result on this PR. `#764` (the #757 sidebar work) is deliberately **not** in here — it is a behaviour change, and it should be validated against a green base rather than bundled into the repair.

### DIFF
--- a/electron/package.json
+++ b/electron/package.json
@@ -23,7 +23,7 @@
     "dist": "npm run build && electron-builder",
     "dist:dir": "npm run build && electron-builder --dir",
     "lint": "tsc --noEmit",
-    "pretest": "node -e \"require('node:fs').rmSync('dist-test',{recursive:true,force:true})\"",
+    "pretest": "node -e \"require('node:fs').rmSync('dist-test',{recursive:true,force:true})\" && node -e \"require('electron')\"",
     "test": "tsc -p tsconfig.test.json && node --test \"dist-test/test/*.test.js\""
   },
   "dependencies": {},
--- a/pixlstash/database.py
+++ b/pixlstash/database.py
@@ -864,6 +864,19 @@ class VaultDatabase:
 
         self._engine = create_configured_engine(self._db_path)
 
+        # The guard holds a raw fd on vault.db. POSIX advisory locks are
+        # per-process, per-inode: closing ANY fd a process has on a file
+        # releases EVERY fcntl lock that process holds on it — including the
+        # ones SQLite took for the engine's live connections
+        # (sqlite.org/howtocorrupt.html §2.2). Closing the guard here, while
+        # pooled connections were open, therefore stripped the server of its
+        # kernel locks on the vault; a second process (CLI, the e2e dedup
+        # seeder, another instance) closing its own connection then saw the
+        # database as unused, checkpointed, and DELETED the live -wal/-shm,
+        # split-braining every open connection ("disk I/O error", "database
+        # disk image is malformed"). The guard must outlive the engine: it is
+        # retained and closed in close(), after dispose().
+        self._location_guard = location_guard
         try:
             with self._engine.connect() as initial_connection:
                 if location_guard is not None:
@@ -877,10 +890,10 @@ class VaultDatabase:
         except Exception:
             self._engine.dispose()
             self._engine = None
-            raise
-        finally:
+            # All connections are gone, so closing the guard fd is safe now.
             if location_guard is not None:
                 location_guard.close()
+            raise
 
         # Write queue and worker
         self._task_queue = queue.PriorityQueue()
@@ -944,6 +957,19 @@ class VaultDatabase:
                     logger.warning(
                         f"VaultDatabase: Exception during engine dispose: {e}"
                     )
+
+            # Only after every SQLite connection is disposed may the guard fd
+            # be closed: closing it earlier releases the process's POSIX locks
+            # on the vault file out from under the live connections (see the
+            # comment in __init__).
+            if getattr(self, "_location_guard", None) is not None:
+                try:
+                    self._location_guard.close()
+                except OSError as e:
+                    logger.warning(
+                        f"VaultDatabase: Exception closing location guard: {e}"
+                    )
+                self._location_guard = None
 
         gc.collect()
         logger.info("VaultDatabase.close called, resources released.")
--- a/pixlstash/database.py
+++ b/pixlstash/database.py
@@ -852,6 +852,16 @@ class VaultDatabase:
         db_exists = os.path.exists(self._db_path)
         logger.debug(f"Vault init, db_path={self._db_path}, db_exists={db_exists}")
 
+        if not db_exists:
+            # Pre-create the database file 0600. Left to SQLite, a missing
+            # database is created at 0644 & ~umask — group/world-readable
+            # under the Debian/Ubuntu default umask 002. Doing it here covers
+            # every construction site, guarded or not. Without O_EXCL a lost
+            # race simply opens the other creator's file; the mode argument is
+            # ignored for an existing file.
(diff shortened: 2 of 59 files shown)


lindkvis:
## Fourth failure, and a security review of the fix for it

`backend-windows shard 1` came back red on this PR. It is not one of the three above — it is a **fourth** breakage that was invisible until now, because `os.fchmod` crashed every Windows test before anything else could run.

```
HubPermissionError: SQLite path C:\Users\RUNNER~1\AppData\Local\Temp\tmp0eb0z7j9\hub.db
                    contains a symlink; refusing to open it.
```

There is no symlink. `RUNNER~1` is an 8.3 short name: `abspath` keeps the short form it got from `%TEMP%`, `realpath` expands it, and the guard treated *any* `absolute != canonical` as proof of a symlink. On Windows that also fires for short names and case normalisation.

`c754ac1c` asks the question that was meant — is any component a symlink — instead of inferring it from a string difference, and names the offending component when it refuses.

`d4aa7fa1` adds what the review below turned up.

### Adversarial review of `c754ac1c`

Question asked: **is there a path the new check accepts that the old one refused?**

- **POSIX equivalence.** Old raises iff `realpath(absolute) != absolute`; new raises iff some component of `absolute` is a symlink. On POSIX `realpath` differs from `abspath` only where a symlink was resolved, so these coincide.
- **The `..`-through-a-symlink escape — checked, does not exist.** `/tmp/x/link/../y` looked like a way to hide a symlink from a per-component walk, but `abspath` normalises `..` *lexically before either check runs*, so both see the same input:
  ```
  abspath : /tmp/tmpuztaspvo/x
  realpath: /tmp/tmpuztaspvo/x
  old check would raise: False
  components symlinked  : False
  ```
- **Symlink loops: the new check is stricter.** `realpath(strict=False)` returns an ELOOP path unchanged, so `absolute == canonical` and the old check *accepted* it. `islink` refuses.
- **Windows `subst` drives and non-name-surrogate reparse points** are resolved by `realpath` but not flagged by `islink`, so the new check is marginally more permissive there. Not a redirection primitive available to another OS principal, and the fail-closed config-dir gate sits behind it. Recorded as accepted risk, not a blocker.
- **TOCTOU** is unchanged: both versions are non-atomic, and `_validate_namespace` + `_validate_file` + `verify_after_open` re-check identity after the open.

### The one finding, now fixed

**The Windows fail-closed policy had no test at all.** That mattered more once this PR adds a session fixture pointing `user_config_dir` at the system temp dir so Windows tests can open a hub under `tmp_path` — scaffolding that widens a security boundary, with nothing asserting the boundary. `d4aa7fa1` pins it in both directions (outside the config dir is refused; inside is allowed, since over-blocking would lock every Windows user out), patching `os.name` rather than skipping off-Windows, so it runs on the Linux lane too.

### Caveat on this review

I wrote the fix, so this is not the independent sign-off `CLAUDE.md` requires. It should have a second pair of eyes on `c754ac1c` before merge, or a `/security-review` run by someone other than me.

lindkvis:
## The Electron flake (`75fa4107`)

Not a flake in the tests — a race with a deterministic cause.

`electron@42` **removed its postinstall**. The binary is now fetched lazily, by `node_modules/electron/index.js`, at require time:

```js
function getElectronPath () {
  let executablePath;
  if (fs.existsSync(pathFile)) { ... }   // path.txt missing → downloadElectron()
```

Nine test files import `../src/config`, which does `import { app } from 'electron'`. `node --test` runs files in **parallel processes**, so each one finds no `path.txt` and starts its own `install.js` into the same `node_modules/electron/dist`. Whoever loses:

```
Error: failed to create '.../electron/dist/LICENSES.chromium.html': File exists (os error 17)
Error: Electron failed to install correctly.
```

### Measured, not assumed

Reproduced locally by installing `electron/node_modules` and counting the download banner inside one test run:

| | downloads inside `npm test` |
|---|---|
| before | **5** (concurrent, racing) |
| after, cold binary | **1** (serial, in `pretest`) |
| after, warm binary | **0** |

53 tests pass, 136 ms warm.

### Why this shape

The fix primes the binary once in `pretest`, before `node --test` forks. Of the two options on the table:

- **Serialising the tests** (`--test-concurrency=1`) also works, but pays wall-clock on every run forever to fix a one-time setup problem.
- **Separate dist folders per worker** removes the collision by giving each process its own ~100 MB download. Worse on every axis.

Priming serialises *the download* and leaves the tests parallel. It fixes local `npm test` too, and a download that fails now fails in `pretest` with its own error instead of surfacing as a mystery race.

### For the CI/security gate

`electron/package.json` only; no workflow file touched. `node -e "require('electron')"` walks exactly the code path the tests already walk — same source, same `@electron/get` checksum verification, no new network origin or trust. Flagging it rather than claiming sign-off, per the rule that CI changes clear security first.

PR #777, opened 2026-08-08
Contain Picture.file_path at the resolver (#776)

### LINKED ISSUE #776: Contain Picture.file_path before it reaches the filesystem
Found during an independent security review of #775 (the Windows trusted-location work), but this is **not Windows-specific** — it applies on every platform.

## The gap

`Picture.file_path` is handed to the filesystem without containment. The single resolver returns an absolute path verbatim:

`pixlstash/utils/image_processing/image_utils.py:147`
```python
if os.path.isabs(file_path):
    return file_path          # no root check
```

`resolve_path_within` exists in `pixlstash/utils/path_utils.py` and is never applied to `Picture.file_path`.

Several sinks then also use `os.path.join(vault_root, rel_path)`, which **discards the base entirely** when `rel_path` is absolute — the standard `os.path.join` footgun.

## Where it lands

| Sink | Location | Trigger |
|---|---|---|
| File **delete** | `services/snapshot_service.py:689-692` → `os.remove(abs_path)` | Unattended. `_apply_gfs_retention` runs on every snapshot creation (`snapshot_service.py:183`), and snapshots are scheduled. |
| File **delete** | `services/scrapheap_service.py:1327-1338` → `os.remove(resolve_picture_path(...))` | Unattended, via `ScrapheapRetentionPurgeTask` (`tasks/scrapheap_retention_purge_task.py:81-88`). The confirm-token gate covers only the manual endpoint. |
| File **write**, attacker-chosen name and suffix | `utils/caption_file_utils.py:197` — `writeback_path` returns `existing_path` verbatim, bypassing `is_safe_sidecar_suffix`; `sidecar_path` is `splitext(path)[0] + suffix` and `_SAFE_SUFFIX_RE` accepts `.bat` / `.ps1` | Unattended, from `tasks/reference_folder_scan_task.py:501-504`. |
| File **read** over HTTP | `routes/pictures/_serving.py:60-62,193` (`FileResponse` on the resolved path; the sibling check `pic.format == ext` is also a DB column), and the token-only `routes/share.py` | Any request. |

I verified the resolver, the snapshot delete and the caption write-back directly; the rest is from the review and worth re-checking as part of the fix.

## How a hostile `file_path` gets in

It has to come from somewhere, and normally the app writes these itself from real filesystem paths. Two routes that do not assume a local attacker:

1. **A substituted vault DB.** This is the Windows case #775 documents, but it is also any POSIX setup where the library directory is writable by another account — `unlink`/`rename` are governed by the *directory's* write bit, not the file's mode, so a 0600 `vault.db` in a shared directory can be swapped wholesale.
2. **Restoring a snapshot or backup archive from an untrusted source.** `services/restore/full_restore.py` has the same `os.path.join(vault_root, ...)` shape and deserves an audit in the same pass.

## Why this is the fix that matters

It shrinks the accepted risk recorded in `trusted_sqlite.py` on **every** platform, rather than trading permission checks against portability. The callers funnel through a small number of functions, so containment is cheap:

- guard inside `resolve_picture_path` (`image_utils.py:147`), and
- replace the absolute-join sites with the existing `resolve_path_within`.

## Tests

Both directions, per the repo's rule: a row whose `file_path` escapes the library root is refused at every sink above, and an ordinary in-root path still resolves, is still served, and is still cleaned up by retention — over-blocking would break normal snapshots and scrapheap purges.

lindkvis:
Closes #776.

Stored paths (`Picture.file_path`, sidecar columns, `Snapshot.relative_path`) are now contained before they reach the filesystem. First and foremost this stops the accident class: a wrong absolute path written into `file_path` by an import or a bug no longer reaches an unattended retention delete, a sidecar write, or the serving route — it is logged and skipped. It also raises the bar against a tampered database row, without claiming to defend against whole-vault substitution (out of threat model per the accepted-risk record).

- Root set: `image_root` ∪ reference-folder roots (host and path-mapped forms), supplied by a vault-registered provider and refreshed on a containment miss, so a just-added folder works immediately.
- Over-blocking regression test: a picture under a reference folder outside `image_root` is still served 200 end-to-end.
- Revert-check done per sink: reverting only the resolver, only the snapshot service, or only the sidecar write-back each fails its corresponding new tests.
- Suites green: new `tests/test_path_containment.py` (14 tests, CI-gated) plus snapshots/restore/scrapheap/server/reference-folder/guardrail suites — 460+ tests, ruff clean.

https://claude.ai/code/session_01VGZQAuSvj8vSFH5NcXBywF

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -357,6 +357,7 @@ jobs:
           tests/test_multi_project_membership_authz.py
           tests/test_openapi_response_schemas.py
           tests/test_operation_log.py
+          tests/test_path_containment.py
           tests/test_pending_score_invalidation.py
           tests/test_picture_is_video.py
           tests/test_picture_mutation_scope.py
--- a/docs/backend_architecture.md
+++ b/docs/backend_architecture.md
@@ -1128,6 +1128,36 @@ The fix is an additive column, `public_id`: 128 bits of randomness as lowercase 
 - File-based **SQLite** at `{image_root}/vault.db`
 - All writes are serialised through `VaultDatabase`'s task queue (single writer); reads run in parallel.
 
+### Stored path containment (#776)
+
+Paths stored in the database (`Picture.file_path`, sidecar columns,
+`Snapshot.relative_path`) are treated as untrusted — a substituted `vault.db`
+or a restored archive can carry arbitrary values — and are contained before
+they reach the filesystem:
+
+- **`ImageUtils.resolve_picture_path` is the chokepoint for picture paths.**
+  The legitimate root set is `image_root` **plus every configured reference
+  folder** (host-side and path-mapped forms). Each `Vault` registers a lazy
+  provider for its reference-folder roots via
+  `path_utils.register_reference_roots_provider` (cached; refreshed once on a
+  containment miss, so a just-added folder is honoured immediately). A path
+  outside the set resolves to `None` with a logged warning — the value callers
+  already handle for an empty path — so unattended tasks skip the row instead
+  of crashing. Containing against `image_root` alone would refuse real
+  reference-folder libraries; never do that.
+- **Snapshot file paths** (`relative_path` / manifest / hashes) go through
+  `resolve_path_within(vault_root, …)` at every read/delete site (retention
+  prune, snapshot delete, restore, identity scrub); `os.path.join` silently
+  discards the base for an absolute component and must not be used there.
+- **Sidecar write-backs** funnel through `caption_file_utils.writeback_path`,
+  which refuses an image path outside the root set and only honours a recorded
+  `tags_file`/`description_file` value that is exactly the image stem plus a
+  safe suffix.
+
+Both directions are asserted in `tests/test_path_containment.py`: escape is
+refused at each sink, and in-root plus reference-folder pictures keep working
+(over-blocking is its own regression).
+
 ### Hub and library identity
 
 `hub.db` sits beside `server-config.json`, outside every image library. It owns
--- a/pixlstash/routes/reference_folders.py
+++ b/pixlstash/routes/reference_folders.py
@@ -1356,6 +1356,7 @@ def create_router(server) -> APIRouter:
                             SIDECAR_TYPE_TAGS,
                             rf.tags_suffix,
                             pic.tags_file,
+                            image_root=server.vault.image_root,
                         )
                         mtime = (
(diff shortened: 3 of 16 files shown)


lindkvis:
Parked as draft by owner decision — do not merge as-is.

1. **Never validated on a working Windows lane.** This branch's CI run predates every Windows fix on #775; its containment predicates are path-string comparisons (normpath/commonpath/realpath), the exact class that produced three separate Windows failures today (8.3 short names, case, path forms). A false refusal at the resolver returns None, which presents to a desktop user as their pictures failing to load.
2. **Conflicts with #775** on `full_restore.py` and `vault.py` — the guard-lifetime and swap-ordering fixes. Must be rebased and fully re-run after #775 merges.
3. **Owner-folder policy.** The resolver containment polices where `Picture.file_path` may point. Legitimate desktop situations get refused: a reference folder removed from the registry (orphan rows), a moved mount, path-mapper forms. The provider's refresh-on-miss covers added folders only.

Plan: after develop is green, split this — the internal `os.path.join(vault_root, …)` fixes (app-generated paths, no policy tension) proceed as a small PR; the resolver containment is shelved pending a green-Windows validation run, an orphaned-folder story, and explicit owner sign-off on the policy line.

lindkvis:
Closed in favour of #781, which lands the half of this that has no policy tension.

The split, and why: containment on a **delete or a write** is pure upside (nothing legitimate is ever outside the roots, and over-blocking strands a recoverable file), while containment on the **read** path polices where a user's pictures may live. The threat is a tampered `Picture.file_path`, but whoever can write the vault DB can read those files directly, so refusing to serve them buys little. Against that sits a real false-refusal surface: orphaned reference-folder rows, moved mounts, path-mapper forms, and the Windows path-spelling class. A resolver refusal returns `None`, which presents to a desktop user as their pictures failing to load.

That resolves all three parked objections rather than answering them. The Windows validation stops mattering because the new path predicates go away; the #775 interaction shrinks to mechanical `os.path.join` → `resolve_path_within` swaps; and the owner-folder policy question disappears because nothing polices owner folders any more.

One thing this review turned up that the original framing missed: the **scrapheap purge delete** resolves through `resolve_picture_path`, so it was the one destructive sink that genuinely needed the reference-folder root set. #781 keeps that, but supplies the roots from the caller (`Vault.reference_folder_roots()`) instead of the global provider registry and cache, which existed only because the resolver is a staticmethod with no vault access.

Branch kept in case the threat model changes.

PR #785, opened 2026-08-08
Give the sidebar one left edge and one indent step (#760)

### LINKED ISSUE #760: Indentation system: six left edges, three depth steps, and a rail compensation that does not compensate
Found during the v1.10.0 model shelf design review, in response to the observation that rows "jump a bit back and forth".

## Measured

**Six distinct left edges** in `SideBar.css` for rows that are all "a row in a list". Effective offset is `padding-left + border-left`, because most rows carry a 3px selection rail:

| Row | file:line | effective left |
|---|---|---|
| `.sidebar-brand` | 803 | 2px |
| `.sidebar-folder-section-header` | 1856 | 5px |
| `.sidebar-folder-row` | 1923 | 5px |
| `.sidebar-section-header` | 1081 | 8px |
| `.sidebar-project-tree-row` | 247 | 11px |
| `.sidebar-list-item` | 1112 | 11px |
| `.sidebar-project-tree-child` | 412 (`!important`) | 19px |
| `.sidebar-section-scroll .sidebar-list-item` | 1742 | 19px |

Scrolling from Folders to People moves the left edge 5 to 19.

**Three different depth steps:** folder tree 7px (off-grid: `--space-2` 4 + `margin-left` 2 + a 1px border, L1966-1968), People/Sets 8px, and the model shelf mock 16px.

**The selection rail is handled three incompatible ways in one file.** Most rows correctly reserve it with `border-left: 3px solid transparent`. But `.sidebar-collapsed-project-menu .sidebar-project-menu-item.active` (L452-459) *adds* the border on select and tries to compensate with `padding-left: var(--space-3)` under the comment `/* compensate for the 3px border so text stays aligned */`. The base already carries that padding (L638), so the label sits at 8px unselected and 11px selected. **Selecting a project moves its label 3px right, under a comment asserting that it does not.**

**Optional leading glyphs are in the flow, and this has already been patched locally twice, in two different mechanisms:**

- `SideBar.vue:5033-5048`, a 15-line inline `:style="{ visibility: ... }"` ternary purely to reserve the chevron's space.
- `SideBar.css:393-398`, an attribute-selector hack `.sidebar-project-tree-sub-chevron[style*="display: none"] { display: inline-flex !important; visibility: hidden }` with the comment `/* v-show sets display:none, override to visibility:hidden so the space is preserved */`.

Two local fixes for one systemic problem. Wherever nobody remembered, the row jumps.

## The pattern already exists in this repo

`frontend/src/components/settings/` is 25 files with exactly **one** `padding-left`, because rows go through `SettingsRow.vue`: a flex row with `gap` and symmetric padding and no per-caller offsets. The sidebar predates it.

## Proposed system

A four-column row grid, where **columns 1 and 2 are reserved and never conditional**:

```
[disclosure gutter] [identity mark] [label 1fr] [actions auto]
```

A thumbnail fills the identity column, an icon centres in it, nothing leaves it empty. Same width in all three cases, which is what makes a character with a face photo share a left edge with one without.

Depth becomes a unitless `--depth` custom property multiplied by one step, not per-level classes or `!important` overrides.

Three tokens, all derived from values already in the codebase rather than invented:

- `--gutter-glyph: 16px`, already the de-facto glyph box (`SideBar.css:2024-2028` forces chevrons to 16px with `!important`, `:2019-2020` forces the folder icon to 16px, section chevrons are `size="16"`).
- `--indent-step: 24px`, being `--gutter-glyph` + `--space-3`, so a child's chevron lands exactly under its parent's label. Equals `--space-6`, so on-grid by construction.
- `--entity-thumb: 24px`, a promotion of the existing sidebar-scoped `--sidebar-thumb-size: 24px` (`SideBar.css:696`) to a global token.

The rail is always present and always transparent on the base; only its colour changes on `.active`. Never compensated with padding.

## Deletions this enables

- `SideBar.vue:5033-5048`, the inline visibility ternary.
- `SideBar.css:393-398`, the attribute-selector hack and its confession comment.
- `SideBar.css:458-460`, the compensation that does not compensate.
- `SideBar.css:412` `!important` and `:1742`, the two hardcoded child indents.
- `SideBar.css:1966-1968`, the 7px composite step.

## Documentation

Belongs in `docs/design/visual-language.md` as a subsection of §5 Spacing and layout, not a new numbered section: §5 already states "columns of controls share a left edge", and this is the missing enforcement of a claim it already makes.

## Sequencing

This moves pixels in the most-looked-at surface in the app, so it should land as its own UI/UX-gated change rather than riding in on a feature. The v1.10.0 model shelf should be built to the system from day one so it does not become a fourth dialect to migrate later.

lindkvis:
Six left edges (2, 5, 8, 11, 19px) and three depth steps (7, 8, 16px) become one edge and one 16px step. Depth is a `--depth` integer times `--indent-step`, never a hand-summed padding + margin + border composite.

Lands before any model-shelf UI so the shelf is not a fourth dialect.

## Beyond what #760 records

**`FolderTreeNode.vue` carried a divergent duplicate of the folder-row styles and the issue never mentions the file.** `SideBar.css` is `<style scoped>`, so it cannot reach a child component; the copy had drifted to an 8px inset with no base selection rail, so **selecting a nested folder shifted its label 3px right**. That is a second instance of the uncompensated-rail bug #760 caught once at `SideBar.css:458`. The row system now lives in the unscoped `SideBar.global.css` and the copy is deleted.

## Two corrections to the issue

**Not the four-column grid it proposes.** The project row puts its caret after the label and row templates carry three to six children with varying meaning, so a positional grid would assign the wrong element to each column and force a visual redesign. A flex row with a fixed-width, never-collapsing glyph slot gives the same guarantee.

**Its token arithmetic pushed names off the rail.** `--indent-step` was specified as `--gutter-glyph + --space-3` = 24px so one chevron would advance exactly one level. Over three real levels that put character names at 91px in a 240px sidebar. The step is now 16px and set by legibility, not glyph arithmetic; rank is carried by type instead. Names start at 75px in the deepest path and 59px in the shallow one.

## Also

- Empty People/Sets captions go inert and dim, with the chevron slot still reserved and the `+` fully operable, since it is the only way out of the empty state.
- The project caret moves to the leading glyph column so every collapsible row discloses from the same place.
- Nothing in the project tree is dimmed. A header must not be quieter than the rows it heads.

## Verification
- 2614 frontend tests pass, `eslint .` clean, `npm run build` ok
- `SideBarRowSystem.test.js` asserts the stylesheet rather than a mount, because these are cascade bugs a component test cannot see. **Every assertion was mutation-tested**: each invariant was deliberately broken and confirmed to fail. One early version passed against a deliberately broken file and was fixed.
- Reviewed on screen by the owner across several rounds.

## Known gap, not introduced here
The section captions and their `+` are `<div>`/`<v-icon>` with click handlers and no `tabindex` or `role`, so **collapsing a section or adding a person is not possible from the keyboard.** Pre-existing; belongs with #755.

Closes #760

### DIFF
--- a/docs/design/visual-language.md
+++ b/docs/design/visual-language.md
@@ -472,19 +472,41 @@ Never add the border on select, and never "compensate" for it with padding. Addi
 3px of border and 8px of padding on `.active` moves the label 3px right; that is a
 bug that has shipped here twice, once under a comment asserting it does not happen.
 
-**Why these three widths.** `--gutter-glyph` is 16px because the sidebar already
-forced every chevron and row icon to 16px with `!important`. `--indent-step` is
-`--gutter-glyph + --space-3` = 24px, and equals `--space-6`, so it is on-grid by
-construction. `--entity-thumb` is 24px, promoted from the sidebar's own
-`--sidebar-thumb-size`. Nothing here was invented.
+**Rank is carried by type; depth is carried by indent.** Two rows at different
+ranks must differ in *type* even when they also differ in indent, so neither
+signal is load-bearing on its own. This is not decoration. When a project row and
+the section captions nested under it were set identically (both `--text-2xs`,
+semibold, uppercase), indentation had to signal both containment *and* rank, and
+the only way to make rank legible was a bigger step. Three levels of that pushed
+entity names to 91px in a 240px rail. One ramp step of contrast between project
+and caption is what makes a 16px indent enough.
 
-**One glyph advances the row by exactly one step.** This is the part that is easy
-to get subtly wrong: a 16px glyph in a row that gaps at 4px advances the label by
-20px, not 24, so glyph columns drift 4px per level and the tree reads as
-hand-placed. The glyph slot therefore carries its own `margin-right: --space-2` on
-top of the row gap, making the advance 16 + 4 + 4 = 24. With that, a child's
-chevron lands under its parent's icon and the child's icon under its parent's
-label. If you change a row's `gap`, re-check this sum.
+**The indent step is set by legibility, not by glyph arithmetic.** `--indent-step`
+is 16px because that is the smallest on-grid step that still reads as nesting at
+this density. It is deliberately *not* derived from the glyph column. An earlier
+version set it to `--gutter-glyph + --space-3` = 24px so that one chevron would
+advance exactly one level, and tuned a margin onto the glyph slot to make the sum
+come out. That bought a child's chevron landing precisely under its parent's
+label, which nobody notices, and charged 24px per level for it. Do not tune the
+step and the glyph advance to each other; they answer different questions.
+
+**Why the other two widths.** `--gutter-glyph` is 16px because the sidebar already
+forced every chevron and row icon to 16px with `!important`. `--entity-thumb` is
+24px, promoted from the sidebar's own `--sidebar-thumb-size`.
+
+**Children indent past their parent's glyph, not past its row.** A disclosure
+group's children step in once from the group's own glyph column, so a child's mark
+lands under its parent's label. That is what every file tree does. Do not flatten
+it to save width; take the width out of the step size instead.
+
+**A caption with nothing under it is not a disclosure.** Rendering a disclosure
+affordance for an empty group is wrong whether the chevron is visible or hidden,
+and reserved-but-blank is the worst of the three because the empty box reads as a
+missing glyph. Make the row inert instead: `aria-disabled`, no hover response,
+label at `--opacity-disabled`. The dimmed label is what explains the blank slot.
+Keep the slot reserved so the caption does not shift when the first child
+arrives, and **keep the group's own add action at full strength and operable**,
+because it is the only way out of the empty state.
 
 **Scoped CSS does not cross a component boundary.** A row rendered by a child
 component gets none of the parent's scoped rules, so a recursive tree component must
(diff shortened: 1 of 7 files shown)


lindkvis:
Heads-up on merge ordering: this and #764 both rewrite \`SideBar.vue\`, \`SideBar.css\` and \`FolderTreeNode.vue\`. Both report MERGEABLE against \`develop\`, but that is each against the base, not against each other. A local test-merge shows **3 conflicting regions across those 3 files**, so whichever lands second needs a rebase.

No action needed now. Whichever you merge first, I will rebase the other and re-verify rather than resolving blind.

lindkvis:
Rebased onto `develop` after #764 merged.

The conflict was a single file, `FolderTreeNode.vue`, and it was the exact duplication this PR removes. #764 added `.not-droppable` to that component's private style block under the comment *"repeated here because this component's styles are scoped."* That was true when it was written.

It stops being true once the folder-row family lives in `SideBar.global.css`. So rather than keep both copies, the resolution **deletes the private one and moves the shared `.not-droppable` rules into the unscoped sheet**, where they reach both the rows `SideBar` renders and the nested ones `FolderTreeNode` renders. One definition, no drift. A new test pins the placement so the copy cannot come back.

#757's behaviour is unchanged: the drop handlers and the refused-drop visual both survive, and `SideBarDropPayloadTypes.test.js` passes.

Re-verified after the rebase rather than trusting the pre-rebase run: **2669 tests across 153 files** (which now includes #764's and #790's suites), `eslint .` clean, build ok. Mutation checks re-run: copying `.not-droppable` back into `FolderTreeNode`, and re-adding the rail on `.active`, each fail.

PR #786, opened 2026-08-08
Let a screen reader follow a long export (#758)

### LINKED ISSUE #758: ProgressOverlay: no ARIA, colour-only failure state, no reduced-motion guard
Found during the v1.10.0 model shelf design review.

`frontend/src/components/widgets/ProgressOverlay.vue` is the shared progress surface for export and plugin progress, and v1.10.0 adds a third caller (the model shelf's cross-drive move, which can be multiple GB and greys the panel while it runs).

## Defects

1. **No ARIA at all.** No `role="progressbar"`, no `aria-valuenow` / `aria-valuemin` / `aria-valuemax`, no `aria-live`, no `role="status"`, no `aria-busy`. A screen reader user gets no start, no progress, no completion and no failure.
2. **Failure is encoded in colour alone.** The `--error` variant (L94-96) recolours the card and changes no glyph and no text. This contradicts the rule `DedupWhyPills.vue` states at L12-15, that glyphs rather than colour must carry the for/against split.
3. **No reduced-motion guard.** The indeterminate animation (L119-135) runs regardless of `prefers-reduced-motion`, unlike `DedupPictureStrip.vue` L532-542 which guards correctly.

## The correct donor is already in the repo

`DedupScanBanner.vue` L129-137 does this properly: `role="progressbar"`, `aria-label`, `aria-valuenow` deliberately null when indeterminate, and a wrapper carrying `role` / `aria-live` / `aria-atomic` at L98-100. Port that pattern rather than inventing one.

## Verification

With a screen reader, a long-running export announces start, periodic progress, completion and failure. The failure state carries a glyph and a word, not only a colour. The indeterminate animation is static under `prefers-reduced-motion`.

lindkvis:
ProgressOverlay had no ARIA at all, encoded failure as a red card and nothing else, and ran its indeterminate slide regardless of `prefers-reduced-motion`. It takes a third caller in v1.10.0 (the model shelf's cross-drive move), which can run for minutes on multiple gigabytes.

Ported from the donor the issue names, `DedupScanBanner.vue`, rather than inventing a pattern.

## The part that needed a judgment call

Getting ARIA *present* is easy. The failure mode that actually makes a screen reader unusable is a correct-looking `aria-live` region that re-reads on every tick.

The region is `aria-atomic`, so any text change re-reads it in full. Announcing each percent would leave the reader talking continuously for the whole length of a multi-gigabyte copy, drowning out everything else on the page. So announcements are **coarse on purpose**:

- quartiles while running, which is the "periodic progress" the issue asks for
- terminal states announced by name
- the exact figure available on demand via `aria-valuenow`
- the tick-by-tick `count / total` line is `aria-hidden`, because those numbers already reach a reader through the bar and would otherwise be a second thing changing inside the atomic region

`aria-valuenow` is omitted entirely when indeterminate. A bar reporting 0 forever is a worse lie than one reporting nothing.

## Also

- Failure carries a glyph and a word. Cancelled and completed are named too, so no state is left to colour or to absence.
- Under reduced motion the slide stops and the track holds filled, rather than parking the 38% stub at the left edge where it would read as stalled.

## Verification
- 2617 frontend tests pass, `eslint .` clean, `npm run build` ok
- 16 new tests. **Every assertion was mutation-tested**: announcing per-percent, claiming 0 when indeterminate, reverting failure to colour only, removing the reduced-motion guard, and re-exposing the counter were each reintroduced and confirmed to fail.

## Not verified
Not exercised with a real screen reader. The tests assert the ARIA contract, not what NVDA or VoiceOver actually says.

Closes #758

### DIFF
--- a/frontend/src/components/widgets/ProgressOverlay.test.js
+++ b/frontend/src/components/widgets/ProgressOverlay.test.js
@@ -1,0 +1,167 @@
+/**
+ * ProgressOverlay accessibility (issue #758).
+ *
+ * The shared progress surface for export, plugin progress and, in v1.10.0, the
+ * model shelf's cross-drive move. It shipped with no ARIA at all, a failure
+ * state encoded only as a red card, and an indeterminate animation that ran
+ * regardless of `prefers-reduced-motion`.
+ *
+ * The announcement tests are the ones with teeth. Getting ARIA *present* is
+ * easy; the failure mode that actually makes a screen reader unusable is a
+ * correct-looking `aria-live` region that re-reads on every tick of a
+ * multi-gigabyte copy.
+ */
+
+import { describe, it, expect } from "vitest";
+import { mount } from "@vue/test-utils";
+import { readFileSync } from "node:fs";
+import { resolve } from "node:path";
+import ProgressOverlay from "./ProgressOverlay.vue";
+
+const stubs = {
+  "v-icon": { template: "<i class='v-icon'><slot /></i>" },
+};
+
+const mountOverlay = (props = {}) =>
+  mount(ProgressOverlay, {
+    props: { visible: true, message: "Exporting", ...props },
+    global: { stubs },
+  });
+
+const live = (w) => w.find("[aria-live]");
+const bar = (w) => w.find('[role="progressbar"]');
+
+describe("the progress bar exposes its value", () => {
+  it("reports a determinate value with the full range", () => {
+    const w = mountOverlay({ percent: 42, status: "running" });
+    expect(bar(w).attributes("aria-valuenow")).toBe("42");
+    expect(bar(w).attributes("aria-valuemin")).toBe("0");
+    expect(bar(w).attributes("aria-valuemax")).toBe("100");
+  });
+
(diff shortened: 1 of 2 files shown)


lindkvis:
Closing as duplicate. #778 fixed #758 and merged at 18:32Z, before I started this.

My error: I checked open issues and open PRs, but not merged ones. Issue #758 was still open because #778 did not reference it in a closing keyword, and I took that as meaning the work was unclaimed. Checking `gh pr list --state merged` would have caught it in one command.

For the record, the merged version is better than this one in two ways I got wrong:

1. **The live region belongs outside the `v-if`.** #778 places it there with exactly the right reasoning: a live region inserted at the same moment as its first text is not reliably announced, so the *opening* line is the one that goes missing. I put mine inside the wrapper, following the `DedupScanBanner` donor, and would have shipped that defect.
2. **It clamps the percent** before feeding `aria-valuenow`. Mine passed the caller value straight through, so an out-of-range percent would have been reported verbatim.

Nothing here is worth salvaging. Issue #758 should probably be closed too, since #778 addressed it.

PR #787, opened 2026-08-08
Describe an ai-toolkit run without touching it

lindkvis:
B6 of the model shelf. The shelf has to show what a training run produced *before* the user commits to importing any of it: which steps were saved, what each looked like, and what it was trained against.

**It only reads.** No hashing, no copying, no writing. That is the load-bearing property, because the shelf runs this against a folder the user has merely pointed at. A test snapshots size and mtime of every file in the tree and asserts nothing moved.

**A run missing its config is still a run.** Steps and previews come from filenames, so an absent or malformed `config.yaml` costs the base model, trigger words and rank, and nothing else. The reason is recorded on the run rather than raised.

## Decisions worth a second opinion

**PyYAML is a local, optional import.** It is present transitively but is *not* declared in `pyproject.toml`. Declaring it would make this a dependency change, which per CLAUDE.md needs security review before push, and that seemed a poor trade for a field the reader works fine without. If you would rather declare it, say so and I will split that out for review.

**Config keys are found by searching the parsed tree, not by a fixed path.** ai-toolkit has moved that nesting between versions; searching by key means a layout change costs nothing here.

## Two traps the tests pin

- The bare final checkpoint carries no step. Treating that as step 0 sorts the finished adapter to the *front* of the run, and step 0 is itself a real save ai-toolkit emits, so the two cannot be conflated.
- `linear: true` is a different setting sharing a name with the rank. Since `bool` subclasses `int`, a naive `isinstance(rank, int)` invents a rank of 1.

Scoped to ai-toolkit deliberately. Other trainers lay output out differently, and a reader trying to cover all of them recognises none of them reliably.

## Verification
- 156 tests pass including the CI-shard gate; `ruff format` and `ruff check` clean
- `tests/test_aitoolkit_run.py` added to `ci.yml` in alphabetical position
- Mutation-tested. **One test was found to be decorative this way**: the "run name ending in digits" case did not actually exercise the step regex anchor, because `sdxl_v2` has no 4+ digit group. Added `flux_2024`, which does, and confirmed it fails against an unanchored pattern.

## Not verified
Not run against a real ai-toolkit output folder. Fixtures are built from the layout in the integration plan, so if real output differs, this is where it shows.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -294,6 +294,7 @@ jobs:
           CI_SHARD: ${{ matrix.shard }}
         run: >-
           python -m pytest $PYTEST_FLAGS --ci-shard "$CI_SHARD/4"
+          tests/test_aitoolkit_run.py
           tests/test_anomaly_region_scope.py
           tests/test_architecture_guardrails.py
           tests/test_async_import_staging.py
--- a/pixlstash/utils/aitoolkit_run.py
+++ b/pixlstash/utils/aitoolkit_run.py
@@ -1,0 +1,330 @@
+"""Read an ai-toolkit training run off disk and describe it, without touching it.
+
+The model shelf needs to show a user what a training run produced before they
+commit to importing any of it: which steps were saved, what each step looked
+like, and what the run was trained against. All of that is already on disk in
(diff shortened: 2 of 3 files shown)


lindkvis:
@copilot resolve the merge conflicts in this pull request

Issue #761, opened 2026-08-08
SET_COLORS: 48 raw Material hexes, and only 2 to 5 are distinguishable as thin marks


lindkvis:
Found during the v1.10.0 model shelf design review, which makes the shelf a second consumer of this palette.

## Measured

`frontend/src/utils/setAppearance.js` `SET_COLORS` is 48 raw Material hex values with zero design-token references. `#e53935`, `#ff5252`, `#f50057` and `#b71c1c` are four separate entries and four reds.

Gated as WCAG 1.4.11 requires for a graphical object (3:1 non-text, both themes), used as a **2px stroke on a 26px chip**:

- **14 of 48** entries are usable. Among those 14, 25 of 91 pairs collide under CVD.
- Maximum mutually distinguishable set: **3 colours** at a CVD floor of 8, **5** at a relaxed floor of 6.
- Against `--panel`, the surface a row thumbnail actually sits on: **4 of 48** usable, maximum set of **2**.

Re-stepping the palette does not fix this. The constraint is the mark, not the hues: a 2px ring on a small chip is the most contrast-hostile form available, and these colours were drawn as fills.

## The honest either/or

If the mark is decorative (identity carried by the thumbnail and hover label), 1.4.11 does not bite and all 48 are legal, but the stated benefit that "rows sharing a group read together at a glance" is false, since 26 of 48 measure under 3:1 against the panel. If the mark is a real grouping signal, 1.4.11 bites and the palette yields 2 to 5 colours. The current design claims the benefit of an encoding while accepting the constraints of decoration.

## Direction

Make the colour an **opaque fill** rather than a stroke: the chip's own background is the entity colour with a luminance-switched foreground glyph, plus a 1px hairline so a light fill still has an edge on white. That moves the contrast question from "hue against an unknown backdrop" to "opaque chip against panel" and "glyph against chip", both of which pass by construction, and it makes **all 48 usable without touching a single stored `set_color`**.

## Do not renumber the list

`set_color` stores the hex, and both `SideBar.vue:1144` and `pixlstash/routes/picture_sets.py:819` seed by list **position**, so any reorder or deletion orphans stored values. Keep all 48 as the manual picker. If an auto-assignment head is wanted later, add it as a named ordered subset those two call sites walk first; that is zero migration.

## Related: the palette is duplicated

[...]

PR #789, opened 2026-08-08
Attach shelf adapters to a library's characters and sets

lindkvis:
B3 of the model shelf. `adapter_attachment` is the **only** model-shelf table that lives in a vault. The adapters, their folders and the checkpoints beside them are hub tables (#788), because those describe the machine. Which character a LoRA belongs to describes *this library*, so it travels with the library.

## The link is a hash, and that is load-bearing

`adapter_sha256 TEXT`, never an integer adapter id. Two reasons, both of which have bitten this codebase before:

- **No foreign key can span the hub and a vault.** They are separate SQLite files, so an integer id would be an unenforceable reference that nothing checks.
- **SQLite hands a deleted row id to the next insert.** An integer link would silently re-point at a *different* adapter after a delete plus insert, and the row would look perfectly valid while being wrong. Same recycled-identifier hazard `library.uuid` and the hub AUTOINCREMENT ids exist to eliminate.

A mutation test confirms this: swapping the column for `adapter_id` fails 9 of 11 tests.

## Characters and sets are peers

Addressed through `entity_type`, not two nullable scalar columns. A scalar `character_id` plus `set_id` makes "attached to both" and "attached to neither" representable states that mean nothing, and every reader would then have to decide what to do about them.

## Rider worth calling out

`character_color` ships in the same migration, per the plan. It is a one-column addition mirroring `PictureSet.set_color`. **Only the column is here.** The first-unused assignment logic and the 48-hex list are #761 and are not touched, so nothing yet writes this column.

## Verification
- 163 tests pass including `test_migrations.py`, plus 42 more across `test_architecture_guardrails.py`, `test_openapi_response_schemas.py` and `test_identity_storage_guardrail.py`, since this adds a SQLModel field
- `ruff format` and `ruff check` clean
- `tests/test_adapter_attachment.py` added to `ci.yml` in alphabetical position
- Migration is conditional in both directions, because the baseline `create_all()` already builds the table on a fresh database from the model added here

## Not verified
No route reads or writes this yet, so there is no AuthzGate surface in this PR. That arrives with the attachment endpoints and will need `ROUTE_POLICIES` entries and coverage-matrix rows.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -294,6 +294,7 @@ jobs:
           CI_SHARD: ${{ matrix.shard }}
         run: >-
           python -m pytest $PYTEST_FLAGS --ci-shard "$CI_SHARD/4"
+          tests/test_adapter_attachment.py
           tests/test_adapter_header.py
           tests/test_aitoolkit_run.py
           tests/test_anomaly_region_scope.py
--- a/pixlstash/db_models/__init__.py
+++ b/pixlstash/db_models/__init__.py
@@ -1,3 +1,8 @@
+from .adapter_attachment import (  # noqa: F401
+    ENTITY_CHARACTER,
+    ENTITY_SET,
+    AdapterAttachment,
+)
 from .character import Character  # noqa: F401
 from .dedup import (  # noqa: F401
     DedupGroup,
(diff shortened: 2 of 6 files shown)


lindkvis:
Rebased onto `develop` after #787 merged. The conflict was the one predicted: #787 and this PR both insert into the same `ci.yml` test list, adjacent alphabetically. Resolved by keeping both, in order:

```
tests/test_adapter_attachment.py
tests/test_adapter_header.py
tests/test_aitoolkit_run.py
```

Re-verified after the rebase rather than trusting the earlier run: 198 tests pass including `test_migrations.py` (the Alembic chain moved under this branch) and `test_architecture_guardrails.py`; `ruff check` clean. `0103` is still the head and still sits on `0102`.

Issue #782, opened 2026-08-08
develop v10.0 nits: review follow-ups from #778, #766, #779, #792, #793, #794, #803


lindkvis:
Joint follow-up for the review nits left on #778, #766 and #779, so they can be cleared in one pass rather than one PR at a time (per the request on [#778 discussion](https://github.com/Pikselkroken/pixlstash/pull/778#discussion_r3741200792)).

All five were checked against the current source on the relevant branch before filing; none of them are stale.

---

## 1. `ProgressOverlay` fill width uses raw `percent` (from #778, merged)

`frontend/src/components/widgets/ProgressOverlay.vue` — the component computes `clampedPercent` to keep progress in 0–100 and to survive a `NaN`, and uses it for `aria-valuenow` and the live-region announcement. The visual fill still binds the raw prop:

```vue
:style="{ width: `${percent}%` }"
```

So a caller handing over `NaN` renders `width: NaN%`, and a caller over 100 overflows the bar — in both cases the bar disagrees with what `aria-valuenow` and the announcement say. Bind the fill to `clampedPercent` too.

Ref: https://github.com/Pikselkroken/pixlstash/pull/778#discussion_r3741095804

## 2. Tag suggestions commit on `mousedown` (from #766)

`frontend/src/components/panels/TbTagPanel.vue`, suggestion option:

```vue
@mousedown.prevent="selectTagSuggestion(item)"
```

`mousedown` does not fire on touch devices, so tapping a suggestion may not work at all; it also commits on press-down, so pressing and dragging away still writes the tag. The usual split is: `mousedown.prevent` only to keep the input from blurring, selection on `click` (or `pointerdown`).

Ref: https://github.com/Pikselkroken/pixlstash/pull/766#discussion_r3741116306

## 3. Tag panel `aria-expanded` can be true with no listbox in the DOM (from #766)

Same file. The input's `aria-expanded` / `aria-activedescendant` are driven by `suggestionsOpen`, but the teleported popup renders on `suggestionsOpen && tagInputRect`, and `tagInputRect` is only filled in on the next tick. There is a window where AT is told the combobox is expanded and pointed at an `aria-activedescendant` id that is not in the DOM. Tie the ARIA attributes to the same condition that renders the popup.

Ref: https://github.com/Pikselkroken/pixlstash/pull/766#discussion_r3741116314

## 4. `aria-checked` on `role="option"` in the add-to-entity list (from #779)

[...]

PR #816, opened 2026-08-09
Close the fixture-generator nits from #782

lindkvis:
Addresses issue #782 items 6-11:

## Items 6-7: generate_model_shelf_fixtures.py
- Remove non-existent `--adapters` workaround from sparse file check errors
- Clean up stale `.safetensors` files before generating new adapters to ensure idempotency

## Item 8: generate_model_shelf_library.py division by zero
- Validate that `pictures` exceeds the largest set span (6) early in `generate_library()`
- Raises `ValueError` with clear message when validation fails

## Item 11: generate_model_shelf_library.py sys.path shadowing
- Make `sys.path.insert()` conditional to avoid prepending repo root when package is already installed

## Item 9: test_model_shelf_library_fixture.py dead assertion
- Replace dead `assert stamped is not None` with actual non-empty check `assert len(stamped) > 0`
- Add regression test for item 8 validation

## Item 10: Deliberately skipped
- `_load_generator()` spec.loader dereference is defendable; a wrong path yields a legible FileNotFoundError

Test results: 26 passed


lindkvis:
Added `588ce7f2` — a leash on the cleanup this PR introduced.

```python
for old in root.glob("*.safetensors"):
    old.unlink()
```

The cleanup is needed: names come from a seeded list, so a shorter second run would leave the first run's tail behind and the folder would hold two generations at once. But `root` reaches it as a positional CLI argument with no default, so pointed at a directory that already has an `adapters/` subfolder it deletes real models silently, with no dry-run and no confirmation.

The generator now drops a `.pixlstash-fixture` marker the first time it fills a folder and cleans **only** a folder carrying it. An unmarked folder holding `.safetensors` stops the run with a message naming the count, untouched. An empty unmarked folder is adopted, so generating into a fresh directory is unchanged.

Three tests: regenerate-and-clean, refuse-untouched, adopt-empty. Mutation-tested — reverting to the bare glob fails two of the three, including the refusal. Cost is inside noise: 16 tests in 0.25-0.27 s, 19 in 0.27-0.39 s, no `Server` and no engine, per the new "reuse the environment" rule.

PR #822, opened 2026-08-09
Share one Server across the multi-project authz tests

lindkvis:
Part of #796 (CI speed). `tests/test_multi_project_membership_authz.py` built a whole `Server` per test — boot, two-picture import through the pipeline, teardown — 36 times. That is now paid once per module, with per-test isolation coming from a reset rather than a rebuild.

**204.72 s -> 61.69 s** (3.3x, ~143 s saved), measured as one uninterrupted old-then-new A/B on the same machine with nothing else running, exit code checked directly. **36 tests before, 36 after.** Nothing under `pixlstash/` is touched.

Module scope, not class scope: the gate shards individual tests over 8 shards, so a class fixture is rebuilt per shard per class and barely amortises.

## What the autouse `env` fixture resets

Every test starts from an identical world. In order:

1. **Domain state**, in one transaction: seeded faces (index >= 900), likeness rows and queue, all project/character/set membership rows, sets, characters, projects, stacks; the two fixture picture rows are then restored column-for-column from a snapshot taken right after import.
2. **Credentials**: every `UserToken` row deleted, token cache flushed with its revocation epoch bumped, owner re-logged-in, all three project tokens re-minted.
3. **The shape** (`_assert_fixture_shape`), asserted on identity — never counts.
4. **A positive control**: each freshly minted token must read its own project under enforcement *before* the test runs. Without this a dead credential produces exactly the 403 the negative assertions are hunting for.

The pictures are deliberately never deleted: they keep their ids, real faces and real embeddings, so no finder is left holding a claim on an id SQLite would hand to a different row.

## Keeping the existence-oracle test honest

`test_project_path_routes_are_not_an_existence_oracle` proves a scoped token cannot tell "this project holds your entity" from "this project exists and does not" from "this project does not exist". That is only a proof if the three are genuinely different to begin with — and `test_deleting_a_project_leaves_the_other_membership_intact` deletes P2 outright.

So `_assert_fixture_shape` re-establishes the difference **as the owner**, who is never scope-restricted, before any token is asked to be blind to it:

- `GET /projects` returns exactly `{P1, P2, P3}` by name **and** id;
- `SharedChar` and `SharedSet` are in `[P1, P2]`, `P1Only*` in `[P1]`;
- `/projects/P1/picture_sets/SharedSet` -> 200 with the right id;
- `/projects/P3/picture_sets/SharedSet` -> 404 **with body** `"Picture set not found"` (P3 exists and does not hold it);
- `/projects/{99999999,NoSuchProjectHere}` -> 404 with body `"Project not found"`.

It lives in the autouse fixture, not a trailing canary: the sharder deals tests individually, so a canary only guards its own shard.

The independent reviewer attacked this rather than taking it on trust. With the shape check neutered it collapsed the probe space two ways — P3 *also* holding SharedSet, and SharedSet detached from P1 — and **both made the oracle tests pass while proving the opposite of their docstrings.** With the check active, both mutations ERROR in fixture setup. A third (P3 deleted) is caught even earlier, by the token positive control.

## Mutation check

`return` inserted at the top of the **body** of `enforce_project_path_scope` (`pixlstash/authz/membership.py`), uniquely matched, in the worktree, uncommitted. Not the `gate.py` docstring trap: the replacement asserted a single occurrence and the inserted line is verifiably below the closing `"""`.

RED:
```
FAILED ...::test_project_picture_sets_listing_both_directions
FAILED ...::test_project_path_routes_are_not_an_existence_oracle
FAILED ...::test_project_token_is_not_told_which_other_projects_exist
3 failed, 33 passed in 59.34s   (exit 1)

E  AssertionError: picture_set token on /api/v1/projects/1/picture_sets/SharedSet got 404;
   a token with no project visibility must be refused: {"detail":"Project not found"}
E  AssertionError: assert 200 == 403   # P3 token read P2's set listing
```

GREEN after restoring the guard (`git status` shows only the test file modified):
```
36 passed in 62.17s   (exit 0)
```

Single-layer removal was enough here, so no defence-in-depth is masking the assertion.

## Independent security sign-off (author did not certify)

A `chief-security-officer` review enumerated all 36 tests hunting for assertions that now pass for the wrong reason. Verdict: 34 HONEST, 2 SUSPECT-latent, 0 BROKEN, **one blocking finding**. All are fixed:

- **BLOCKING — the credential reset was a no-op.** `usertoken` is a **hub** table; the vault carries an empty, never-read copy from the baseline migration. Deleting the vault's copy revoked nothing, and every previous test's token stayed live. Proven both ways: a probe replaying the previous test's token got **200** before the fix and **401** after. Now deleted through `server.auth._db`, with a post-reset assertion that zero rows survive. The docstring claim that ids "are themselves new" was also wrong — SQLite recycles them — so the file was resting its isolation on rowid reuse; that is now stated accurately.
- **Anti-vacuity guard bypassed.** Making `env` autouse moved its own ~25 HTTP calls, including the bare `status_code == 200` positive control, *ahead* of `no_spa_fallback`. Pulled back under it via `request.getfixturevalue`.
- **Planner could stay dead.** `stop()` only warns on join timeout and `start()` returns early on a live thread, which would 503 the import tests for the module's life. Now asserted running.
- **`pending_character_id` survived** on leftover imported pictures and could re-target the next test's `SharedChar` (character ids recycle). Now nulled.
- Inert scaffolding in `_seed_set_sort_inputs` documented as no longer load-bearing.

## Coordinator findings from #821, applied

- **PRAGMA.** Confirmed unreliable exactly as described: issued before any DML it reads back **0**. Issued after the first DELETE it reads back 1, and the file *asserts* that in-transaction rather than assuming it. It was load-bearing as first written, so the deletes were reordered children-before-parents; **removing the pragma entirely from the new order now leaves all 36 green**, so ordering carries correctness and the pragma is a demonstrated-live safety net for unenumerated FK edges.
- **Route existence on negative assertions.** Scope checks sit ahead of routing, so a renamed path 403s identically to a scope refusal. Added `assert_real_route` to the four negatives that hit a path by string without proving it: `test_project_token_is_not_told_which_other_projects_exist`, `test_project_token_keeps_filtering_by_its_own_project`, `test_picture_scope_follows_the_shared_set`, and `_assert_picture_reaches_both_projects`.

## Shard independence

The gate may run any subset alone. Verified: shards 1,2,4,5,7,8 of 8 all green; six individual tests standalone in their own process (including the oracle test, the likeness test, both imports, and the P2-deleting test); and the reviewer ran the file in reverse order, 36/36.

## Where the pattern did not transfer cleanly

- `seed_likeness_stable` could not be reused: it calls `wait_likeness_settled`, which requires *every* picture to carry an embedding. The staging-import tests leave pictures behind whose embedding finder is detached, so it would have hung for the full timeout — the 378 s trap. The wait is paid once in the module fixture instead, and the test seeds directly.
- `_detach_volatile_finders` reaches into `WorkPlanner` privates and must stop the scheduler first: mutating `_task_finders` under the live thread threw `IndexError` in `_run_finders_once` and killed the planner outright. Observed, not theoretical.
- `_imported_picture_id` assumed a library with exactly the two fixture pictures. Replaced with a before/after diff inside `_staged_import`, and the generated PNG is now content-distinct per filename so the second import is not deduplicated away.

`ruff format` and `ruff check pixlstash` clean. `tests/ci_test_durations.json`, `.github/workflows/ci.yml`, `test_read_token_security.py` and `test_picture_mutation_scope.py` untouched.


lindkvis:
## Why all 8 Linux shards failed, and the structural fix

**Cause.** `tests/test_generic_field_reader_allowlist.py` borrowed this suite's fixtures by copying them out of the module namespace:

```python
from tests import test_multi_project_membership_authz as _multi_project
env = _multi_project.env
```

pytest resolves a fixture's parameters against the module it was *collected* in, so a copied fixture keeps working right up until it grows a dependency the copy never mentioned. This PR gave `env` a `_module_env` parameter; only `env` was re-exported, so every test in the borrowing module errored at setup:

```
E       fixture '_module_env' not found
20 errors in 0.08s
```

The gate deals tests across all eight shards, so all eight went red. Windows passed because the OS-sensitive list names only `test_multi_project_membership_authz.py`, which was fine — the exporting module never noticed.

**Fix (approach B).** Both suites now live in `tests/multi_project_authz/` with a package `conftest.py` holding `_module_env` and `env`, plus a `shared_env.py` for the plain helpers each module imports normally. Fixture dependencies resolve themselves, nothing is re-exported by hand, and `autouse` stays scoped to exactly these two files. The one-line `_module_env = _multi_project._module_env` was rejected: it re-breaks the next time a shared fixture grows a parameter, and there are more module-fixture conversions coming.

A/C were considered and dropped: `pytest_plugins` in a non-root conftest is an error on pytest 9.1.1 (the pinned version), and putting `env` in the global `tests/conftest.py` would either make an autouse reset apply to the whole suite or need a distinct name and a rename of the `env` parameter in all 56 tests.

**Guardrail.** `test_every_test_file_is_classified` globbed `tests/test_*.py` non-recursively, so files in a sub-package would have dropped out of CI without going red. It now uses `rglob`. `ci.yml` (both the Linux gate and the Windows OS-sensitive list) and the keys in `ci_test_durations.json` follow the new paths.

### Verification

| run | before (this branch) | after |
|---|---|---|
| `test_generic_field_reader_allowlist.py` alone | 20 errors in 0.08s | **20 passed** in 46.8s |
| `test_multi_project_membership_authz.py` alone | 36 passed in 60.5s | 36 passed in 72.4s |
| both together | 36 passed + 20 errors | **56 passed** in 105.8s |
| both, reverse order | — | 56 passed in 110.0s |
| `--ci-shard 2/8` and `5/8` over the package | — | 7 passed each |
| `tests/test_ci_shards.py` | — | 133 passed |

Against `develop` (per-test `Server`, pre-conversion) the same 56 tests take **279.5s**, so the shared `Server` still buys ~2.6x and no per-test `Server` came back.

**The autouse safety property still holds.** A scratch test was added to *each* module: one that deletes P2 through `env`, followed by one that takes no `env` parameter at all. Both non-`env` tests saw `{P1, P2, P3}` — the reset ran anyway. Flipping the fixture to `autouse=False` made both fail with `{'P1', 'P3'}`, so the probe is not vacuous. Scratch tests removed afterwards.

**The anti-vacuity guard still covers the fixture's own HTTP calls.** With `request.getfixturevalue("no_spa_fallback")` in place, `TestClient.request` is the guard's `_guarded` wrapper at the point `env` makes its ~25 calls; deleting that one line leaves it as the bare `request`. The explicit call is still load-bearing and the ordering note in the docstring is still accurate.

### Review comments

All three were real and are fixed in their own commits.

1. **Planner restart race** (most serious, confirmed against `work_planner.py`). `stop()` joins with `timeout=5` and only *logs* if the thread survives; `start()` then returns early on `_thread.is_alive()` **before** reaching `self._stop.clear()`. The survivor exits a moment later and the planner is dead for the rest of the module — and the existing `assert planner.is_running()` passes, because the *old* thread is what is alive. Symptom: an intermittent 503 from a staging-import test several tests later. `_detach_volatile_finders` now waits for the thread to actually exit before touching `_task_finders`, and asserts `is_running() and not _stop.is_set()` after the restart. Fixed in the test helper, not in `WorkPlanner`: `stop()`'s best-effort contract has other callers, and changing it is not this PR's business.
2. **Brittle count assertion.** `session.scalar(...)` now returns the int directly. Worth stating why it mattered: `in (0, (0,))` passes for a `Row` holding *any* count, so the credential-reset guard could have gone green with tokens still live.
3. **Order-insensitive PNG seed.** `sum(filename.encode())` collides on anagrams (`abc.png` and `cba.png` both sum to 665); now `zlib.crc32` over the whole name, feeding all three channels. Same bug class as the model-shelf fixture generator in #815.

`ruff format` and `ruff check` clean across `tests/` and `pixlstash/`.

PR #849, opened 2026-08-09
Warn on a stale durations map, and stop opening PRs to fix PRs

lindkvis:
`tests/test_ci_shards.py::test_recorded_durations_actually_balance_the_gate` currently hard-fails when any gated test file is missing from `tests/ci_test_durations.json`. That is red on `develop` right now, and it has been failing PRs that have nothing to do with it.

## Why the hard fail is the wrong shape

The failure is **non-local**, and structurally cannot block the change that causes it:

| | |
|---|---|
| #832's CI run (adds `test_security_supported_versions.py`) | 12:01 |
| #833 merges, introducing the guardrail | 13:23 |
| #832 merges on its pre-guardrail green check | 13:54 |

By the time the guardrail evaluates the new file, that file is already on `develop`. So it can only ever fail **bystanders** — it is not a forcing function, it is a lottery, and #839 (a websocket test fix) drew the ticket.

The documented contract already said so, in `.github/copilot-instructions.md`:

> a test missing from the map keeps its seeded position and is charged the median cost … Refreshing the map after adding tests is an optimisation chore, **never a correctness obligation**

The sharder honours that (`tests/conftest.py`). The guardrail did not. A missing entry costs shard skew and never coverage, so blocking a merge over a pure optimisation input inverts the cost.

`UNRECORDED_IN_DURATIONS_MAP` made it worse rather than better: it is self-cleaning in both directions, so each new file costs two edits — add, then remove after the next regeneration — and the removal is the same non-local failure in reverse.

## What this does *not* do

It does not simply flip to a warning. The defect #833 fixed was real: the balance was modelled over the map's own keys, so 28 of 119 gated files were structurally invisible and it reported a perfect `1.000` over 76% of the gate. A pure warn re-opens that by drift, silently.

So: **warn per file, fail on a coverage floor.**

- `UNRECORDED_IN_DURATIONS_MAP` deleted (it was empty, and existed only to escape the hard fail).
- Missing files are named in a warning every run.
- `MINIMUM_GATE_COVERAGE = 0.9` — below that, the 1.05 balance figure is describing a shrinking subset rather than the gate, and it fails.

Verified both directions: today's map warns and passes at 119/120 (99.2%); a simulated 76/100 — the historical defect — fails with the right message.

## The failing branch is exercised

An unexercised fail branch is a guardrail on paper, which is the specific way this repo has been bitten. `_warn_or_fail_on_map_coverage` is split out and tested directly at 100/90/89/76/0 percent, plus the warn path and the no-warning path.

**Mutation-tested:**
- `MINIMUM_GATE_COVERAGE = 0.0` → 3 tests red
- warn reverted to a hard `assert` → 6 tests red

142 passed, ruff clean.

## Note

`CLAUDE.md` is a symlink to `.github/copilot-instructions.md`, so the documentation edit lands in both from one change.


lindkvis:
Added `1885de4d`: a new `## Fixing a CI failure: update the existing PR, do not open another` section in `.github/copilot-instructions.md` (`CLAUDE.md` is a symlink to it, so both get it).

It belongs here rather than in its own PR — this branch already edits that file, and opening a fourteenth PR to add a rule against opening unnecessary PRs would have been its own counterexample.

The five rules, each with the evidence from 2026-08-09:

1. **A CI failure on a PR is fixed on that PR's branch**, even when the cause is elsewhere. Pushing to the red PR spends the run it was already going to spend; a second PR spends a second one and leaves the first red.
2. **`gh pr list --state open` before opening anything.** Two sessions independently fixed the same red guardrail (#848 and #851) — ~400 runner-minutes for one change, and the check costs a second.
3. **Fold the unblock into the fix.** #849 and #851 were one problem in one file; splitting them doubled the CI for no review benefit.
4. **Close a superseded PR immediately.** #841 and #845 stayed open after #846 absorbed them, drawing runs from every push to their bases — and merging #845 would have reintroduced code that later fixes corrected.
5. **One PR per work step**, stacking only when the pieces can genuinely merge independently. B7's six-PR stack all had to land together, so it re-ran the gate on growing supersets and bought nothing.

Plus the corollary that actually caused this: per-agent instructions bound each agent's diff and nothing bounds the aggregate. Counting PRs across in-flight lanes is the orchestrator's job.

Numbers for context: one full gate run is ~200 runner-minutes; 13 PRs were open at once and 47 were opened in a day.

PR #869, opened 2026-08-10
Give the model shelf its four missing verbs

lindkvis:
**F3a**, the backend half of F3. Of the five verbs the plan names, only Assign
had a route (`PUT /adapters/{sha256}/attachments`); Rename, Set base model, Set
kind and Forget had none at all. 414 lines of source.

The five verbs were never written down anywhere, so they were derived from what
the plans pin — the two named confirmations, the shipped assignment route, and
the filter's `Needs a name` queue — and ruled by the owner on 2026-08-10, now
recorded in the shelf plan.

## Two routes, not four

`PATCH /api/v1/models` carries Rename, Set base model and Set kind, because all
three write one curated hub column and differ in nothing else. Three routes
running the same UPDATE with a different column name would be three sets of
guards to keep in step.

- **Only the fields the body carries are written**, via `model_fields_set`
  rather than a null check. So an explicit `null` is a *clear* (a wrong base
  model back to unset, which returns the row to the filter's "not set" bucket)
  while an absent field is untouched. That is what lets one route hold three
  verbs without Set base model blanking the names in the selection.
- **`display_name` is refused for more than one id.** A name is a fact about one
  file; in bulk it would give every selected row the same one, and there is no
  undo.
- **`file_kind` cannot be cleared, only corrected.** Every file is something and
  `unknown` is how the shelf says so; a null leaves a row neither list matches.
- **A correction the hub's CHECK would reject is refused by name.** Promoting an
  unhashed checkpoint to `adapter` violates `CHECK (file_kind <> 'adapter' OR
  sha256 IS NOT NULL)`; left to SQLite that is a 500 naming a constraint, which
  says nothing about the file the owner picked.

`POST /api/v1/models/forget` is the fifth, and the only shelf operation that
destroys curation.

- **Gated on the row's state, never on the size of the selection.** A model is
  forgettable only when no copy is `present` *or* `unreachable`. The second is
  the one that matters: `unreachable` is the we-could-not-look state an
  unplugged NAS produces, and acting on it would let one call wipe the curation
  for a whole drive.
- **Ids that fail the gate come back under `refused` with a reason**, rather
  than failing the call: a selection is made against a list that may be seconds
  old, and failing everything because one file came back is the wrong answer to
  good news. That response is the receipt the shelf will show.
- **Vault attachments survive.** `adapter_attachment` is keyed by content hash
  in each library's vault, so libraries that are not open are unreachable from
  here and deleting only the active half would be an arbitrary subset. Left
  alone they are invisible and re-link if the file returns.

## Addressing

Every verb takes hub `model.id`, never a hash. A 24 GB checkpoint is listable
the moment it is registered and stays `sha256 NULL` until the hash finder reads
it, so a hash-addressed verb layer would leave the largest files on the shelf as
the only ones that cannot be corrected.

## Authz

Both routes are `OWNER_ONLY` with their registry entries and coverage-matrix
rows in this PR. Deliberately **not** the §16.3 locality tier the folder
mutators carry: both take a list of hub ids, name no host path, and neither
touches the filesystem — Forget deletes database rows and never unlinks a file,
because the files being gone is the precondition for the call. The handler's own
guards are data checks, not scope checks.

**This wants the independent CSO pass before merge**; I wrote it and should not
certify it.

## Verification

105 tests in `test_model_shelf_api.py`, 17 of them new and covering both authz
directions with a live-token positive control. Mutation-tested the state gate:
dropping `unreachable` from `_KEEPS_A_MODEL_ALIVE` turns
`test_forget_refuses_a_model_we_could_not_look_for` red, and restoring it turns
it green. Architecture guardrails and the §16.3 tier split pass;
`render_backend_architecture.py --check` is clean. `ruff format`/`check` clean.

F3b (selection bar, the verb controls, receipts, the two confirmations) follows
once this lands.


pixlstash/routes/model_shelf.py
```diff
@@ -560,4 +700,140 @@ def put_adapter_attachments(
...
+        """Refuse a correction the hub's own CHECK constraints would reject.
+
+        ``model`` carries ``CHECK (file_kind <> 'adapter' OR sha256 IS NOT
+        NULL)`` and the same for ``kind``. Promoting an unhashed 24 GB
+        checkpoint to ``adapter`` therefore violates a constraint, and left to
+        SQLite it surfaces as a 500 naming ``CHECK constraint failed``, which
+        tells the owner nothing about the file they picked. Checked here so the
+        answer names the file instead.
+        """
+        if changes.get("file_kind") != FILE_ADAPTER:
```
lindkvis: This early return is finding #1, case A. It branches on the `file_kind` **in the body**, but the CHECK fires on the row's *resulting* state. `PATCH {"ids":[id], "kind": null}` against a row that is already `file_kind='adapter'` names no `file_kind`, returns here, and then violates `CHECK (file_kind <> 'adapter' OR kind IS NOT NULL)` -> 500.

Reproduced against the real `_V2_MODEL` DDL.

Computing the effective post-write state (`changes["kind"] if "kind" in changes else row["kind"]`, same for `file_kind`) removes the need for this early return and covers both missed cases at once.

pixlstash/routes/model_shelf.py
```diff
@@ -560,4 +700,140 @@ def put_adapter_attachments(
...
+        placeholders = ", ".join("?" for _ in ids)
+        blocked = [
+            dict(row)
+            for row in server.hub.fetchall(
+                f"SELECT id, filename, sha256, kind FROM model "
+                f"WHERE id IN ({placeholders})",
+                tuple(ids),
+            )
+            if row["sha256"] is None
+            or (row["kind"] is None and changes.get("kind") is None)
```
lindkvis: Finding #1, case B. `row["kind"]` is the **stored** algorithm, but the call may be about to null it out. `PATCH {"ids":[id], "kind": null, "file_kind": "adapter"}` on a row with `kind='lora'` leaves `blocked` empty — `row["kind"] is None` is False — so the guard passes and the UPDATE writes `kind=NULL, file_kind='adapter'` straight into the CHECK. 500.

The test should be on the value being written, not the value being replaced.

pixlstash/services/model_shelf_service.py
```diff
@@ -396,3 +401,126 @@ def write(session: Session):
...
+        return [], []
+
+    placeholders = ", ".join("?" for _ in ids)
+    known = {
+        int(row["id"])
+        for row in hub.fetchall(
+            f"SELECT id FROM model WHERE id IN ({placeholders})", tuple(ids)
+        )
+    }
+    alive = {
```
lindkvis: Finding #2. This `alive` read is its own lock acquisition — `HubDatabase.fetchall` takes and releases `self._lock` per call — and the DELETEs happen in a separate `hub.transaction()` further down. `ModelFolderScanner` writes `model_file.state` from background tasks (`model_folder_scanner.py:715/722/748/770`), so a rescan can flip a row `missing` -> `present` (or `unreachable` -> `present`) in that window and the model gets forgotten regardless.

That is the exact outcome the gate exists to prevent, on the one shelf operation with no undo. Moving this and the `known` read inside the existing transaction block, as `conn.execute`, holds the lock from gate to delete and closes it.

lindkvis (commented):
## CSO review — F3a verb layer

**Authz: no findings.** `OWNER_ONLY` is the right tier for both routes and the reasoning for *not* using the §16.3 locality tier holds — I traced both handlers and neither takes a host path nor touches the filesystem; `forget_models` is `DELETE FROM model_file` / `DELETE FROM model` and nothing else. Registry entries and coverage-matrix rows are present and match. The negative test asserts the route resolves first (`assert_real_route`) and carries a live-token positive control, so the 403s are genuine refusals rather than a dead path.

Also cleared: no SQL injection (the interpolated column names come from the closed `CURATABLE_FIELDS` tuple and `update_models` raises on anything else; every id is a `?` bound from a Pydantic `list[int]`); `MAX_MODELS_PER_EDIT = 5000` is comfortably under `SQLITE_LIMIT_VARIABLE_NUMBER` (32766 since SQLite 3.32); no sensitive values in the log lines; `extra="forbid"` on both request bodies.

**Two findings below, neither an authz hole.** One is the exact failure mode the PR says it eliminated; the other is in the destructive path.

### 1. `_refuse_impossible_adapters` misses two reachable inputs — both still 500

The PR's stated contract is "a correction the hub's CHECK would reject is refused by name rather than left to surface as a 500". Two bodies get past the guard and hit `CHECK (file_kind <> 'adapter' OR kind IS NOT NULL)`. Reproduced against the real `_V2_MODEL` DDL from `pixlstash/hub/schema.py`, on a row that is already `file_kind='adapter', kind='lora'`:

```
PATCH /api/v1/models {"ids":[1], "kind": null}
  -> IntegrityError: CHECK constraint failed: file_kind <> 'adapter' OR kind IS NOT NULL

PATCH /api/v1/models {"ids":[1], "kind": null, "file_kind": "adapter"}
  -> IntegrityError: CHECK constraint failed: file_kind <> 'adapter' OR kind IS NOT NULL
```

Case A never reaches the guard at all: it returns at `if changes.get("file_kind") != FILE_ADAPTER` because the body names no `file_kind`, but the *stored* `file_kind` is already `adapter`. Case B does reach it, and passes, because the guard tests the **stored** `kind` (`'lora'`, not None) instead of the `kind` the call is about to write (None).

No global `IntegrityError` handler exists, so both are a 500. `hub.transaction()` rolls back and logs, so there's no data damage — the cost is the unhelpful error the guard was written to avoid, on a route the owner reaches by clearing a field the API description explicitly offers.

Smallest fix: decide the effective post-write state instead of branching on the body. `effective_kind = changes["kind"] if "kind" in changes else row["kind"]`, same for `file_kind`, then refuse when `effective_file_kind == FILE_ADAPTER and (effective_kind is None or row["sha256"] is None)`. That collapses both cases and the currently-covered one into a single condition, and drops the early return.

Worth adding both bodies to `test_model_shelf_api.py` — `test_an_unhashed_checkpoint_cannot_be_called_an_adapter` covers only the `sha256` limb of one CHECK.

### 2. Forget's state gate is read outside the transaction that deletes

`forget_models` takes three separate hub lock acquisitions: `fetchall` for `known`, `fetchall` for `alive`, then `hub.transaction()` for the two DELETEs. `HubDatabase.fetchall` takes `self._lock` per call and releases it, and `ModelFolderScanner` writes `model_file.state` from background tasks in its own transactions (`model_folder_scanner.py:715/722/748/770`).

So a rescan can flip a row `missing` -> `present` in the window between the `alive` read and the DELETE, and the model is forgotten anyway. That is precisely what the gate exists to prevent, on the one operation the PR describes as destroying curation with no undo and no operation-log half. The `unreachable` case is the same shape and is the one the PR argues hardest about: a NAS coming back mid-call.

The window is small, but the consequence is unrecoverable, which is the wrong side of that trade. Fix is to make it one critical section — move both `SELECT`s inside the existing `with hub.transaction() as conn:` block and use `conn.execute` for them, so the lock is held from gate to delete. Nothing else needs to change; the refusal-building loop moves in with them.

(Minor, same area: `update_models` re-selects `existing` inside its transaction while `_refuse_impossible_adapters` selects outside it. Harmless today because the outcome there is a 400 or a 500 rather than a delete, but it is the same seam.)

---

**Release blockers:** none. Neither finding is an authz or data-exposure issue.

**Fix before merge:** #2 — it is five lines and it protects a no-undo delete.

**Can follow:** #1, though it is cheap enough to fold into the same push, and the PR body currently claims the behaviour it does not yet have.

**Accepted, noted not flagged:** `display_name` / `base_model` / `kind` are unbounded free text with no global body-size limit, consistent with the rest of the codebase and owner-only; forgetting a hub row destroys curation other libraries' `adapter_attachment` rows point at, which the PR documents as deliberate and which the single-owner threat model makes fine.

lindkvis:
Both findings fixed in e86a4bfc.

**1. `_refuse_impossible_adapters` missed two reachable inputs.** Confirmed, and the diagnosis was exact: case A never reached the guard because the body named no `file_kind`, and case B passed it because the guard read the stored `kind` rather than the one about to be written. Rewritten to decide on the post-write state, `changes.get(col, row[col])` for both columns, which collapses all three cases into one condition and drops the early return. The cheap short-circuit that remains is `if "file_kind" not in changes and "kind" not in changes: return`, since nothing else a caller may write can reach either constraint.

Three tests added, both reported bodies plus a positive control that clearing `kind` on a *checkpoint* still succeeds — the constraint only binds adapters and over-blocking would be its own regression. Mutation-checked: restoring the stored-`kind` read turns exactly the two new refusals red and leaves the checkpoint one green.

**2. Forget's gate read outside the transaction.** Confirmed. Both `SELECT`s now run on the transaction's own connection inside the existing `with hub.transaction()`, with the refusal-building loop moved in with them, so the hub lock is held from gate to delete. `tests/test_model_shelf_api.py::test_forget_reads_its_gate_inside_the_write_transaction` counts the reads that escape the transaction rather than trying to schedule the race, and goes red when either `SELECT` is moved back onto `hub.fetchall`.

On the minor note about `update_models`: left as is for now. It is the same seam, but the outcome there is a 400 or a 500 rather than a delete, and folding the guard's read into that transaction would mean the route's validation ran with the hub write lock held — a worse trade on a path that mutates nothing when it refuses. Worth revisiting if the guard ever grows a second query.

Both fixes are documented in `docs/backend_architecture.md` §"The shelf's five verbs", and the PR body's claim about the CHECK guard is now true rather than aspirational. 109 tests in the file, guardrails and the architecture render check green.

PR #877, opened 2026-08-11
Recover two merged-but-missing shelf changes, and adopt the folded base model

lindkvis:
**Read this part first: two changes that show as merged are not on `develop`.** I
found it while starting the folding adoption and checking my baseline.

| PR | What happened |
|---|---|
| **#872** | Merged at head `1b43d113`. My later commit `0f2c033d` — the `base_model_folded` field — was pushed after that and **was never merged**. |
| **#873** | Merged into **`feature/shelf-file-manager-selection`**, not `develop`. Its base was never retargeted when #871 landed, so the merge commit `e54b8dd6` sits on that branch and none of it reached `develop`. |

Neither was lost — #872's commit was still on its branch and #873's on its base
branch — and both are recovered here. Verifiable on `develop` before this PR:
`grep -c withEmptyFolders frontend/src/utils/modelShelf.js` → **0**, and
`grep -c base_model_folded pixlstash/routes/model_shelf.py` → **0**.

Worth a look at the merge settings: a PR whose base branch is another PR's branch
will merge *there*, and GitHub only auto-retargets when the base is deleted.

## What this PR contains

1. **#872's missing commit**, cherry-picked: `base_model_folded` computed in
   `_to_response`, with its four tests.
2. **#873's missing work**, recovered by merging its base branch: registered
   folders that hold no models get a group, `withEmptyFolders`, and the move of
   `compareGroups` / `UNSET_GROUP_KEY` into `utils/modelShelf.js`.
3. **The adoption itself** — the actual new work.

## The adoption

`baseModelKey()` prefers the folded label and falls back to the raw string. Four
call sites moved together, and they had to:

- **group** — one header per base instead of one per spelling;
- **facet** — one checkbox per base;
- **filter** — matched on the same key the facet was built from;
- **sort** — a run ordered by base rather than by how the trainer typed it.

Splitting them would have been the silent failure: a facet list built from folded
values with a filter matching raw ones offers a box that hides most of the rows it
promises. A test pins exactly that.

An unrecognised base model keeps its raw string as its key, so a private or
brand-new base stays selectable rather than being swept into "not set". The row
still **displays** `base_model`, because the raw spelling is what the file says.

## Verification

2,835 frontend tests, 7 new. Mutation-checked: dropping the folded preference from
`baseModelKey` turns four of them red across both suites — which is also the proof
that the four call sites are genuinely wired to one helper.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -172,6 +172,42 @@ Decompose by domain first, then fan out. **Independent** sub-tasks should run co
 - **Build frontend:** `npm run build` (in `frontend/`)
 - **Dev frontend:** `npm run dev` (in `frontend/`)
 
+## Never open a PR onto another PR
+
+**A PR's base must be a long-lived branch — `develop`, `main`, a release branch.
+Never another PR's branch.** Stacking reads as tidy and loses work.
+
+It lost work here on 2026-08-11. Three shelf PRs merged inside forty seconds:
+
+* **#873** had **#871**'s branch as its base. #871 merged to `develop`; #873 then
+  merged **into #871's branch**, not into `develop`. GitHub only auto-retargets a
+  PR when its base branch is *deleted*, and this one was not. The badge said
+  merged, the content was not in the product, and it was found by chance two
+  hours later — `grep -c withEmptyFolders frontend/src/utils/modelShelf.js`
+  returned **0** on `develop`.
+* **#872** merged at the head GitHub had recorded, which was one commit behind
+  what had just been pushed to it. The newer commit was simply left on the
+  branch. Same silent shape, different cause.
+
+Neither was recoverable by looking at the PR list, because both said MERGED.
+
+**If you need to add to an open PR, there are exactly two ways:**
+
+1. **Push the commits onto that PR's own branch.** This is almost always right.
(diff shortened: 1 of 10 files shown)


lindkvis:
Added the policy this PR's incident argues for (`43a4a69d`), folded in here rather than opened as a seventh PR — the guardrail and the recovery are one change, and the PR body already carries the evidence.

**Never open a PR onto another PR.** A PR's base must be a long-lived branch; never another PR's branch. Two permitted ways to add to an open PR:

1. push the commits onto that PR's own branch, or
2. open a replacement PR carrying the old PR's full history, targeting the same base the old one did, then close the old one.

Both keep a single merge into a long-lived branch, and neither leaves a base that can be merged, deleted or retargeted out from under you.

It also records the corollary, which is the part that would have caught this in minutes rather than hours: **verify the merge, not the badge.** `MERGED` is a statement about a pull request, not about the branch you are about to build on — `git merge-base --is-ancestor <head-at-merge> origin/develop` is.

Worth noting the two failures had *different* causes and the same symptom, which is why the rule is about the base rather than about timing: #873 merged into its base PR's branch, while #872 merged a head that was one commit behind what had just been pushed. Both showed MERGED.

Issue #897, opened 2026-08-12
Model shelf: the four naming states are not distinguished


lindkvis:
Naming is the most common fix on the shelf, so the design gives it four distinct
states. The shelf currently renders a name and nothing else.

| State | Treatment in the design |
|---|---|
| named | plain, with a dashed underline and a pencil appearing on row hover, so the field advertises that it is editable |
| **needs a name** | italic placeholder with a dashed accent underline and a persistent pencil — an empty *field*, never disabled-looking text |
| **derived** | a real readable name we generated, marked as ours-not-the-file's so it is not mistaken for something the owner typed |
| **from the file** | a small accent chip stating the name came from the filename |
| editing | bordered input with a focus ring, committed on Enter |

The distinction that matters: an unnamed model must look like *an empty field
inviting a name*, not like greyed-out text. The current rendering makes an
unnamed row look inert, which is why it never gets named.

### Done when

The four states are visually distinct, the affordance appears on hover and on
keyboard focus (not hover alone), and the derived / from-file distinction is
legible without opening anything.

Design source: the resolved design artifact, "Naming".

lindkvis:
Implemented and merged in #914 (`f389f2d7`), which landed on `develop` — the model shelf does not exist on `main`, so the `Closes #897` in that PR never fired.

All four naming states are drawn distinctly on `develop` today: `named` (semibold, dashed rule on hover **and** `:focus-within`), `derived` (UI face + outline DERIVED tag), `from-file` (mono + accent FROM FILENAME tag), `needs-a-name` (italic *Name this model* prompt, permanent accent rule, pencil that never hides), plus the bordered editing field with a focus ring, committed on Enter. The derived/from-file distinction is carried by the word on the tag, so it reads in greyscale.

Covered by `ModelShelf.test.js` / `modelShelf.test.js` (179 tests green at `034aca2e`). Closing as done.

PR #918, opened 2026-08-13
Give the drive band three segments, a legend and a low-free state

### LINKED ISSUE #893: Model shelf capacity meter shows two fills, not three segments and a legend
The drive band ships (`GET /model-folders/devices`, #868, rendered as
`.shelf-band-meter` in `ModelShelf.vue`), but the meter is **two fills** where
the design specifies **three segments and a legend**:

| Segment | Meaning |
|---|---|
| ours | what this shelf's folders occupy on that drive |
| other | everything else on the drive |
| free | remaining space |

Today there is `shelf-band-fill` and `shelf-band-fill--shelf` and no key, so a
reader cannot tell "PixlStash is using this much" from "the drive is this full",
which is the entire question the band exists to answer.

Also missing:

- **A legend** naming the three segments, once per view rather than per band.
- **A low-free warning** on the band, in the warning colour, when the drive is
  near full.
- **Square segments.** The track rounds its outer ends only; rounding each
  segment puts a curve mid-stack where the data has no end, and a rounded edge
  reads as "the bar stops here" when it does not.
- The colour pair must stay distinguishable for deuteranopia — the design
  records a measured separation for the neutral against the accent, in both
  light and dark, and dark is stepped independently rather than flipped.

### Done when

Three segments, a legend, a low-free state, square inner edges, and the contrast
check recorded for both themes.

Design source: the resolved design artifact, drive band / capacity meter.

lindkvis:
Closes #893.

The drive band's meter was two overlaid fills with no key, so a reader could see a boundary without being able to tell which of the meter's two questions — "how full is this drive" and "how much of that is ours" — it marked. This makes it three segments and names them.

## What changed

**Three segments instead of two fills.** `bandUsage` now returns `shelfPct` / `otherPct` / `freePct`, where `other` is the *rest* of the used space rather than all of it. They sum to exactly 100% by construction, which is what lets them be a plain flex row with no rounding sliver opening at the right-hand end. Rounding lives on the track and nowhere else — `overflow: hidden` clips the outer ends and every inner boundary stays square, so no curve appears mid-stack where the data has no end.

**A legend, once per view.** Rendered only when a *measured* band is actually on screen: an unmeasured drive draws no meter, so a shelf of only offline drives would otherwise print a key to a picture nobody can see.

**A low-free state.** Triggered on an absolute floor (`LOW_FREE_BYTES`, 50 GiB), never a percentage — the question is "does the next checkpoint fit", and a percentage is wrong in both directions on this hardware: 10% of a 4 TB model drive is 400 GB and cries wolf, while 10% of a 256 GB SSD is 25 GB — but so is 60 GB free on that same disk. It is carried by the word "Only" leading the label, an alert glyph and semibold figures, with the warning hue additive on top, so it survives greyscale. No live region: it is a standing fact about a disk, can be true of several bands at once, and would otherwise fire a burst on every device refresh.

## The contrast work, which was the substance of this

The issue asks for a measured separation in both themes, with dark stepped independently. That turned out to be forced by arithmetic rather than a matter of taste, and what ships today fails:

| pair | light (today → new) | dark (today → new) |
|---|---|---|
| `ours` : `other` | 2.90 → **3.29** | **1.34** → **3.36** |
| `other` : `free` | 1.53 → **14.03** | 1.84 → **11.28** |

The dark **1.34:1** is the deuteranopia failure the issue describes: the same alpha flipped from light gives a neutral almost exactly as luminous as the olive, so in greyscale the two segments merge. The issue also only names the `ours`/`other` pair, but `other`/`free` — the boundary that answers "how full is this drive" — measured 1.53 and 1.84 and was arguably the worse of the two.

Three findings worth a reviewer's attention:

- **The neutral is what steps per theme, not the accent.** `primary` stays "ours" in both themes; a legend swatch that is orange in light and olive in dark is not one system.
- **Dark cannot be a flipped alpha.** On a dark canvas an ink alpha can only *lighten*, so the neutral cannot pass below the olive. It goes light instead and the *track* takes the dark end via a `scrim` tint.
- **In light the neutral has to go past the ink itself.** `on-background` tops out at L .0153 and measures 2.95 — just under the floor — so the fill is `shadow` `#1c160c`.

Separating the segments with a hairline instead was considered and rejected: the requirement is a *luminance* separation and a drawn line is not one, so with a quiet mid-grey neutral the pair still measures 1.73:1 and greyscale still merges them.

The track itself is only 1.22 / 1.11 against the canvas, which would make a nearly-empty drive a ghost, so it carries a 1px `outline` at 4.24 / 5.67. A meter's own frame is a fair substitute for a fill boundary; a line between two fills is not.

Measurements are recorded in `docs/design/visual-language.md` §4 as a reusable rule — *in a multi-segment meter the neutral goes to the extreme away from the canvas* — since this is the first meter the design system has had.

## One judgement call to confirm

`shadow` is documented as an elevation colour and using it as a **fill** is a role stretch. It is the light theme's warm near-black extreme and the only on-token value that clears 3:1 (`scrim` clears at 3.85 but is cold black on a warm palette). If that is rejected, the swap is a named `meter-neutral` theme key — identical pixels, cleaner semantics — and it is localised to one line per theme in `style.css`.

## Notes

- **This targets `develop`, not `main`.** The model shelf does not exist on `main`; #868, which shipped the band, merged to `develop`.
- The meters lost their `role="img"` and `aria-label`. `.shelf-band-figures` already renders the identical string as visible text in the same heading, so every band was announcing its figures twice. `role="meter"` would be wrong for a separate reason — one `aria-valuenow`, three numbers.
- `ModelShelf.test.js > a run's disclosure > keeps the count a drawn figure and not a second control` fails on this branch. **It is not from this change** — it fails on `develop` too. #897's rename pencil carries `role="button" tabindex="-1"`, which the test's `[tabindex], [role="button"]` selector matches even though `tabindex="-1"` is not a tab stop. Left alone as out of scope; worth a follow-up on whether the selector or the pencil is wrong.

### DIFF
--- a/docs/design/visual-language.md
+++ b/docs/design/visual-language.md
@@ -378,6 +378,37 @@ Findings to honour:
   4.90 / 5.30. All clear the 3:1 UI floor. Light `success` and `info` were Material
   500s until 2026-07 and measured 2.64 / 2.97 — **below the floor** — which is the
   reason they were deepened to `#2e7d32` and `#1a6ec4`.
+- **A multi-segment meter steps its neutral to the extreme *away* from the canvas**,
+  never to a mid-grey between the accent and the track. The range between the accent
+  and the canvas is not wide enough for two 3:1 steps; the range *past* the accent is.
+  Proven on the model shelf's drive band (#893), whose three segments are `ours` /
+  `other` / `free` and therefore have two adjacencies to keep readable. Composited over
+  `background`, light on `#faf9f7` / dark on `#1b1f24`:
+
+  | pair | light | dark |
+  |---|---|---|
+  | `ours` `#567309` : `other` (`#1c160c` / `#d4c9c1`) | **3.29** | **3.36** |
+  | `other` : `free` (`#e5e3e1` / `#121518`) | **14.03** | **11.28** |
+  | `free` : canvas | 1.22 | 1.11 — *drawn*, not inferred (see below) |
+
+  Three things this pins down, each of which was got wrong first:
+  **(a)** The neutral is what steps per theme, not the accent — `primary` stays the
+  "ours" colour in both, because a legend swatch that is orange in one theme and olive
+  in the other is not one system. **(b)** Dark is stepped *independently*, never a
+  flipped alpha: on a dark canvas an ink alpha can only lighten, so the neutral cannot
+  pass below the olive, and the shipped `rgba(ink, .28)` measured **1.34:1** — the
+  deuteranopia failure that opened the issue. The neutral goes light and the *track*
+  takes the dark end via a `scrim` tint. **(c)** In light the neutral must go past the
+  ink itself: `on-background` tops out at L .0153 and measures 2.95, just under the
+  floor, so the fill is `shadow` `#1c160c` (L .0085).
+  Separating the segments with a hairline instead was rejected: the issue's requirement
+  is a *luminance* separation, and a drawn line is not one — with a quiet mid-grey
+  neutral the pair measures 1.73:1, so in greyscale you would see two segments and be
+  unable to tell which was which.
+  The track is only 1.22 / 1.11 against the canvas, which would make a nearly-empty
+  drive a ghost, so it carries a 1px `outline` at 4.24 / 5.67 rather than relying on a
+  colour step. A meter's own frame is a legitimate substitute for a fill boundary; a
+  line *between two fills* is not.
 - **The eight `on-<status>` values, on their solid fill:** light `on-error` #ffffff
   4.86 · `on-warning` #23211d 4.95 · `on-success` #ffffff 5.13 · `on-info` #ffffff
   5.16; dark, all four `#1b1b1b`: `on-error` 4.68 · `on-warning` 5.53 · `on-success`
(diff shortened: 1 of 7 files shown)


lindkvis:
Good catch — fixed in e831d77c.

This was a regression I introduced: the previous two-fill version clamped with `Math.min(100, …)` on each fill, and I removed those because the three segments sum to 100 by construction. That identity only holds while `free <= total`, which I did not guard.

It is reachable rather than theoretical. `free_bytes` and `total_bytes` are two separate reads of the device (`shutil.disk_usage` → `statvfs` in `system_utils.py`), and thin-provisioned or transparently-compressed filesystems (ZFS, btrfs) report more free than total by design; a network mount's `statvfs` can simply be wrong. Unclamped, a 1000-byte drive reporting 1500 free gave `freePct` 150 and a segment running past the track.

`free` is now clamped to `[0, total]` before the identity is relied on, alongside the existing `shelf`-to-`used` clamp, and the docstring now states that both inputs are clamped and why. Added a unit test for it — I verified it fails (150 vs 100) with the clamp removed, so it is not a vacuous assertion.

PR #919, opened 2026-08-13
Restore foreign keys after clean_db wipes (#712)

### LINKED ISSUE #712: clean_db fixtures silently leave foreign keys OFF on a pooled connection
`clean_db` fixtures silently leave foreign keys OFF on a pooled connection

Found while adding the missing engine-configuration regression test for #709: a test asserting `PRAGMA foreign_keys` on a live pooled connection read back `0` where it should have read `1`.

## Cause

The `clean_db` fixtures wrap their wipe like this:

```python
def _wipe(session):
    session.exec(text("PRAGMA foreign_keys = OFF"))
    session.exec(delete(Snapshot))
    ...
    session.exec(delete(Picture))
    session.exec(text("PRAGMA foreign_keys = ON"))   # <- never takes effect
    session.commit()
```

`PRAGMA foreign_keys` is **a no-op inside a transaction** ("foreign key constraint enforcement may only be enabled or disabled when there is no pending BEGIN or SAVEPOINT"). The first `DELETE` opens one, so the restoring `ON` is silently ignored — no error, no warning:

```
start           : 1
after OFF       : 0
after ON in txn : 0   <- no-op inside a transaction
after COMMIT    : 0   <- still off, connection keeps it
```

SQLAlchemy's pool does not reset PRAGMAs when a connection is returned, so that connection carries `foreign_keys = 0` for the rest of its life and hands it to whichever test picks it up next.

## Impact

Test-only — no production code turns FK off on a pooled connection (verified by grep; the only occurrences are under `tests/`). But any test in a file with one of these fixtures that expects FK enforcement can silently get FK-off, depending on which pooled connection it draws. Affected: `tests/test_snapshots.py:61`, `tests/test_metadata_hash.py:48`, `tests/test_metadata_hash_batching.py:154`, and `tests/test_restore.py`.

That is a *weakened* test rather than a failing one, which is the bad direction: a test written to prove FK enforcement works can pass because enforcement was off.

## Suggested fix

Any of:
- issue the pragmas outside a transaction (commit first, then restore), which is the minimal change;
- drop the ordered-delete-with-FK-off approach and delete children before parents, so no pragma is needed;
- or restore it at the connection level rather than the session level.

Whichever is chosen, a `PRAGMA foreign_keys` that silently no-ops is worth an assertion — the failure mode here is that the restore looks like it happened.

Related: the engine-configuration work in #709, whose new `test_full_restore_rebuilds_the_live_engine_with_the_startup_configuration` deliberately references a fresh `create_configured_engine` rather than the pre-restore live engine because of this.

lindkvis:
Fixes #712.

## What was wrong

`PRAGMA foreign_keys` is a no-op while a transaction is pending. Every `clean_db`
fixture issued the restoring `ON` after its `DELETE`s, so it was silently ignored
and the connection went back to the pool with FK enforcement off — for the rest of
its life, and for whichever test drew it next.

Six fixtures had the bug, two more than the issue lists: `test_snapshots.py`,
`test_restore.py`, `test_metadata_hash.py`, `test_metadata_hash_batching.py`,
`test_server_simple.py` and `test_scrapheap_retention.py`.

## What changed

One shared `wipe_tables(session, models)` helper in `tests/utils.py`, passed
straight to `db.run_task`, replaces all six copies of the pragma dance. It commits
before restoring the pragma, restores in a `finally` so a delete that raises
mid-wipe cannot leak enforcement off either, and raises `RuntimeError` if the
restore did not take — the failure mode here was that it looked like it had, so the
check must not be droppable by `python -O`.

In `test_scrapheap_retention.py` only the vault wipe goes through the helper; the
hub database's identity wipe issues no pragma and is left as it is.

The other `PRAGMA foreign_keys = OFF` sites (`test_migrations.py`,
`test_picture_is_video.py`) are on short-lived raw `sqlite3` connections that are
closed immediately, not on the pool, so they are left alone.

## Fallout: 7 tests were relying on the leak

With FK enforcement genuinely on, seven `test_restore.py` tests failed. Each has a
helper that simulates "the user deleted this character/project" by deleting the row
directly, which violates `face.character_id` / `picture.project_id` /
`pictureset.project_id` / `character.project_id` — plain FKs with no `ON DELETE`
action. The production routes null those pointers first
(`routes/characters.py::clear_character_and_nullify_faces`,
`routes/projects.py::delete_project`); the tests only ever passed because
enforcement was off. One comment even asserted the wrong mechanism
("Face.character_id cascades to NULL"), now corrected.

Added `delete_characters` / `delete_projects` to `tests/utils.py` covering the
FK-relevant part of what those routes do, and pointed the six call sites at them.
Their docstrings name what they deliberately leave out (the reference picture set;
the #125 primary-project re-derivation, which differs only when the entity belongs
to another surviving project). No production code changed — this is the
weakened-test class the issue predicted, surfacing on contact.

## Verification

All six affected files plus `test_ci_shards.py`: 371 passed. `ruff check` clean.
`test_server_simple.py::test_scrapheap_delete_forever_upgrades_stale_kept_flag`
fails, but fails identically on a clean `develop` checkout — it is pre-existing and
untouched by this branch.

No new test file, so no CI shard-list change is needed — the guard inside
`wipe_tables` runs in every one of those fixtures, on every test.

### DIFF
--- a/tests/test_metadata_hash.py
+++ b/tests/test_metadata_hash.py
@@ -15,8 +15,7 @@ import json
 import tempfile
 
 import pytest
-from sqlalchemy import text
-from sqlmodel import delete, select
+from sqlmodel import select
 
 from pixlstash.db_models import (
     Character,
@@ -30,6 +29,7 @@ from pixlstash.db_models import (
 from pixlstash.db_models.tag import Tag
 from pixlstash.db_models.snapshot import Snapshot
 from pixlstash.server import Server
+from tests.utils import wipe_tables
 
 
 @pytest.fixture(scope="module")
@@ -44,21 +44,20 @@ def server():
 
 @pytest.fixture(autouse=True)
 def clean_db(server):
-    def _wipe(session):
-        session.exec(text("PRAGMA foreign_keys = OFF"))
-        session.exec(delete(Snapshot))
-        session.exec(delete(Tag))
-        session.exec(delete(Face))
-        session.exec(delete(PictureSetMember))
-        session.exec(delete(PictureProjectMember))
-        session.exec(delete(PictureSet))
-        session.exec(delete(Project))
-        session.exec(delete(Character))
-        session.exec(delete(Picture))
-        session.exec(text("PRAGMA foreign_keys = ON"))
-        session.commit()
-
-    server.vault.db.run_task(_wipe)
+    server.vault.db.run_task(
+        wipe_tables,
+        [
+            Snapshot,
+            Tag,
+            Face,
+            PictureSetMember,
+            PictureProjectMember,
+            PictureSet,
+            Project,
+            Character,
+            Picture,
+        ],
+    )
     yield
 
 
--- a/tests/test_metadata_hash_batching.py
+++ b/tests/test_metadata_hash_batching.py
@@ -35,8 +35,7 @@ from datetime import datetime
 
 import numpy as np
 import pytest
-from sqlalchemy import event, inspect as sa_inspect, select as sa_select, text
-from sqlmodel import delete
+from sqlalchemy import event, inspect as sa_inspect, select as sa_select
 
 from pixlstash.database import (
     _HASH_SKIP_COLS,
@@ -55,6 +54,7 @@ from pixlstash.db_models import (
 from pixlstash.db_models.snapshot import Snapshot
 from pixlstash.db_models.tag import Tag
 from pixlstash.server import Server
+from tests.utils import wipe_tables
 
 
 # ---------------------------------------------------------------------------
@@ -150,21 +150,20 @@ def server():
 
 @pytest.fixture(autouse=True)
 def clean_db(server):
-    def _wipe(session):
-        session.exec(text("PRAGMA foreign_keys = OFF"))
-        session.exec(delete(Snapshot))
-        session.exec(delete(Tag))
-        session.exec(delete(Face))
-        session.exec(delete(PictureSetMember))
-        session.exec(delete(PictureProjectMember))
-        session.exec(delete(PictureSet))
-        session.exec(delete(Project))
-        session.exec(delete(Character))
-        session.exec(delete(Picture))
-        session.exec(text("PRAGMA foreign_keys = ON"))
-        session.commit()
-
-    server.vault.db.run_task(_wipe)
+    server.vault.db.run_task(
+        wipe_tables,
+        [
+            Snapshot,
+            Tag,
+            Face,
+            PictureSetMember,
+            PictureProjectMember,
+            PictureSet,
+            Project,
+            Character,
+            Picture,
+        ],
+    )
     yield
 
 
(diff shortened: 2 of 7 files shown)


lindkvis:
Merged `main` in (no conflicts once merged locally — main's only overlap was stylistic `assert client.post(...)` reformatting in `test_server_simple.py`, far from the fixture) and addressed both review comments in 13525740.

**1. Leak on exception / `assert` under `python -O`** — both taken. The wipe is now in a `try`, with the restore in a `finally` that first `rollback()`s (a no-op after the successful commit, and what ends a half-done wipe when a delete raises) so the pragma has no pending transaction either way. The check now raises `RuntimeError` rather than asserting — the point of the guard is that a no-opped pragma is silent, so it shouldn't be droppable by `-O`. Verified both paths restore FK=1 and that the original exception still propagates from the failure path.

**2. Docstrings promise more than the helpers do** — correct, fixed by scoping the docstrings rather than growing the helpers. Each now says it implements the FK-relevant part only and names what it leaves out: `delete_characters` does not delete the character's reference picture set, and `delete_projects` nulls `project_id` unconditionally instead of re-deriving it from surviving memberships (#125) — which is equivalent only when every project the entity belongs to is deleted, and its single caller deletes the entity's sole project. Implementing the full route semantics would be unused code in these tests.

225 tests across the six affected files pass; `ruff check` clean.

lindkvis:
Rebuilt on `develop`. The branch had picked up a merge of `main`, which dragged 15 main-only commits onto it and caused the conflicts; that merge is gone. Force-pushed, so the branch is now the same two commits it always should have been, on top of `develop` — nobody else's commits were on it.

Resolving against current `develop` needed three real adjustments, not just a replay:
- identity moved to the hub database, so `test_scrapheap_retention.py` keeps its separate `hub_engine` wipe and only the vault wipe goes through `wipe_tables`;
- `test_snapshots.py` needs its `text` import back — develop's newer identity test uses it, and the original commit had removed it as unused;
- `test_scrapheap_retention.py` likewise needs `delete` back for the hub wipe.

371 tests pass across the affected files. `test_server_simple.py::test_scrapheap_delete_forever_upgrades_stale_kept_flag` fails, but fails identically on a clean `develop` checkout, so it is pre-existing and not from this branch.

lindkvis:
Fixed in 344501ed.

The failure was mine, and it was a process error rather than a code one: `test_snapshots.py` needed its `text` import back and `test_scrapheap_retention.py` needed `delete` back (develop's identity tests use them; removing the fixture's pragma made both look unused). I made both edits when rebuilding on develop, verified them, and then pushed the branch tip **without committing them** — so my local run passed against a working tree that the pushed commits did not match.

Reproduced from a clean checkout of the pushed commit before changing anything: `NameError: name 'text' is not defined` and `NameError: name 'delete' is not defined`, matching Windows shard 1. Re-verified from a clean checkout of the fix rather than from my working tree.

One unrelated failure remains locally, `test_server_simple.py::test_scrapheap_delete_forever_upgrades_stale_kept_flag`, which fails identically on a clean `develop` checkout here and is not touched by this branch. It passed on Windows shard 2/3 in CI, so it looks local to this machine.

PR #920, opened 2026-08-13
Model shelf: make the capacity meter the drop target, showing projected usage

### LINKED ISSUE #894: Model shelf: the capacity meter should be the drop target, showing projected usage
The design makes the capacity meter itself the **drop target** for a move, and
shows the consequence *before* the drop commits:

- Dragging models over a drive band highlights the band as a target.
- A **ghost segment** projects how much the drive would hold after the drop,
  hatched so it reads as provisional rather than measured.
- If the drop would not fit, the projection renders in the error treatment and
  the band shows a reject state instead.
- A label under the band states the outcome in words.

None of this exists. Drops currently target the folder group header
(`shelf-group-btn--drop`), and the meter is decorative.

This is the affordance that makes "which disk has room" actionable instead of
informational, and it is also the honest place to refuse a move: the refusal is
visible while the pointer is still down, rather than as a message after the fact.

### Done when

- A drag over a band shows the projected usage and an explicit fits / does not
  fit state.
- A rejecting band cannot be dropped on.
- Projection is visually distinct from measured usage (hatch, not a solid).

Depends on the three-segment meter issue. Design source: the resolved design
artifact, drop states.

lindkvis:
Closes #894. Built on #893's three-segment meter, so this targets `develop`
where that landed — not `main`.

## What changed

The drive band's capacity meter was decorative; a move could only be dropped on
a folder group header, and whether the disk had room for it was something you
found out afterwards. Now the band **is** a drop target, and it draws the
consequence while the pointer is still down.

- **A fourth, hatched segment** projects what the drop would add, carved *out
  of* the free segment rather than laid over it. `bandProjection` returns a
  replacement for `bandUsage`'s object with `freePct` already reduced, so the
  four segments still sum to exactly 100% and the existing flex row and
  no-per-segment-clamp guarantee are untouched.
- **Hatched, not solid.** Three segments are things that were measured and this
  one is a thing that has not happened; a fourth flat colour would have read as
  "already on the disk". It reuses the 45° texture the sidebar's
  `.not-droppable` and the grid's ghosted tiles already use, at 2px/4px because
  the track is only 6px tall.
- **A drop that will not fit is refused while the pointer is down**, next to a
  projection saying by how much, rather than as a message after release. The
  band takes the error treatment and `preventDefault()` is simply not called, so
  the browser's own "no drop here" cursor lands on it.
- **The refusal belongs to the disk, not to one header.** `dropFits` is checked
  in the folder-header handler too, so a folder on a full drive stops
  highlighting as well — the band above it is where the reason is drawn.
- **The outcome is stated in words** under the band: "100.0 GB fits · 300.0 GB
  free after", "100.0 GB will not fit · 60.0 GB short", "Already on this drive ·
  nothing to copy". The hatch and the hue are readable neither aloud nor in
  greyscale; this is the half that is.

## Two things worth a reviewer's attention

**A same-drive move adds nothing.** Those are renames — the server reports
`bytes_to_copy` of zero for them — so a band nets out the copies already sitting
on it before projecting. Without that, a drive would refuse a move onto itself
that costs no space at all. `movableCopies` now returns `bytesByFolderId`
alongside `items`; it is deliberately kept **off** `items`, which is posted to
`/model-moves` verbatim. `bandKeyFor` is factored out of `bandGroups` so
"is this copy already on that drive?" is answered by the one rule that builds
the bands, not by a second copy of it.

**A band is a disk, but a move needs a folder.** A band drop resolves to the
first folder on that drive a move may go to. That is safe because a drop still
does not move on release — the dialog states the destination and its select
corrects it — and it is kinder than refusing a drive that happens to hold two
eligible folders, which would be a refusal the reject treatment does not mean.

An **unmeasurable** drive does not refuse: `bandProjection` returns `null` and
`dropFits` answers true, because "we cannot say" must not be drawn as "does not
fit". The band still highlights as a target — it just has no ghost and no
outcome to state — and the server checks before it copies.

## Tests

13 new: `bandProjection`'s arithmetic (the four segments summing to one track,
the bar clamping while the verdict does not, the exact-fill boundary, the
unmeasured null), `movableCopies`' per-folder byte split, and seven component
tests covering the projection, the in-flight refusal, a refused drop not opening
the dialog, the destination a band drop resolves to, the same-drive case, the
folder header refusing on a full drive, and the unmeasured drive still
accepting.

`npx vitest run` — 3016 passed, 167 files. `npm run build` clean, eslint clean
(the one `KIND_ICON` warning is pre-existing), prettier formatted.
`docs/frontend_architecture.md` §9.1a updated for the meter-as-drop-target rules
and the move section's "two ways in" corrected to three.

### DIFF
--- a/docs/frontend_architecture.md
+++ b/docs/frontend_architecture.md
@@ -1619,6 +1619,49 @@ models is thirty-six headers. Under `Folder` the second level is spent on the
   `mdi-alert-outline` glyph and semibold figures, with the warning hue additive
   on top; so it survives greyscale, and it gets no live region, because it is
   true of several bands at once and would fire a burst on every device refresh.
+- **The meter is the drop target, and it projects the consequence before the
+  drop commits (#894).** A drag over a band draws a **fourth, hatched segment**
+  carved *out of* the free one — `bandProjection` returns a replacement for
+  `bandUsage`'s object with `freePct` already reduced, so the four still sum to
+  100 and the flex row and the no-clamp guarantee both survive untouched. The
+  hatch is the point: three segments are things that were *measured* and this
+  one is a thing that *has not happened*, and a fourth flat colour would have
+  said it was already on the disk. It is the same 45° texture the sidebar's
+  `.not-droppable` and the grid's ghosted tiles use, at 2px/4px because the
+  track is 6px tall.
+  - **The projection nets out copies already on that drive.** A move inside one
+    drive is a rename — the server reports `bytes_to_copy` of zero for it — so
+    `movableCopies` returns `bytesByFolderId` alongside `items`, and the band
+    sums only the copies whose `bandKeyFor` differs from its own. Without that
+    a drive refuses a move onto itself that costs nothing. `bytesByFolderId` is
+    deliberately **not** on `items`, which is posted to `/model-moves` verbatim.
+  - **A band that has no room refuses, and so does every folder header on it.**
+    The refusal belongs to the *disk*: `dropFits` is checked in both handlers
+    and drawn in one place, the band, so the reader is told why while the
+    pointer is still down rather than by a message after the release. Refusing
+    is simply not calling `preventDefault()`, so the browser's own "no drop
+    here" cursor lands on a band already in the error treatment.
+  - **An unmeasurable drive does not refuse.** `bandProjection` returns `null`
+    and `dropFits` answers *true*: "we cannot say" must not be drawn as "does
+    not fit". The band still highlights as a target (`bandDropState` keys on the
+    pointer and the fit, never on a projection existing) — it simply has no
+    ghost and no outcome to state. The server checks before it copies.
+  - **A band drop resolves to the first folder on that drive a move may go to.**
+    A band is a disk and a move needs a folder, so one has to be chosen.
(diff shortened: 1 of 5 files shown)


lindkvis:
Addressed the review comment on `bandProjection`.

The object is documented as a replacement for `bandUsage`'s, but it was spreading `usedPct` and `lowFree` through unprojected — measured before the drop, sitting on the same object as segments drawn from after it. Nothing reads them off the projection today (the band's low treatment deliberately goes through `bandUsage`), so this was latent rather than visible, which is precisely why it would have been found late.

Both are now recomputed:

- `usedPct` from the **drawn** figure, so it stays exactly `shelfPct + otherPct + addedPct` and the bar cannot report itself as more than full.
- `lowFree` from the **unclamped** `freeAfter`, so a drop that overruns the drive reports low rather than reporting the zero free space the segment was clamped to — the opposite reading, and on the worst case.

Two tests pin it: a drive that is not low before a drop and is after it (the projected and measured answers must differ), and the over-full case where `freePct` clamps to 0 but `usedPct` reaches 100 and `lowFree` stays true. Both fail against the previous code.

No rendering changed. Full suite 3018 passed.

PR #947, opened 2026-08-15
Show the shelf when Engines is the only block ticked

lindkvis:
Ticking **Engines** alone in the model shelf's `Show` panel fetched the engine rows, counted them in the toolbar — and then drew *"Nothing is selected in Show"* over the top of them.

## The cause

`nothingSelected` named three of the four blocks:

```js
() => !filters.adapters && !filters.checkpoints && !filters.unclassified
```

`engines` was added to `BLOCKS`, to `fetchRows`, to `blockOf` and to `defaultFilters`, but not here. So a selection asking only for engines read as a selection asking for nothing.

## The change

- `nothingSelected` and `activeCount` both derive from `BLOCKS` instead of naming blocks one at a time, so the two computeds that decide "is anything selected" can no longer disagree with the fetch that decides what is loaded. The four lists that say something *different* per block (`defaultFilters`, `storedFilters`, `blockOf`, `fetchRows`) stay explicit.
- `activeCount` now counts turning **Engines** off, the way it already counts `unclassified` and `checkpoints`. That also un-greys the `Show` panel's Reset button when engines-off is the only deviation — a second, deliberate consequence, called out here because it is not the reported symptom.
- The empty states are reordered: *no models match these filters* now runs ahead of the terminal *no models found* whenever `activeCount` is non-zero. A narrowed selection only fetches the blocks it asks for, so a shelf reopened with one empty block ticked arrives with `rows` empty on a machine full of models — and the old order dead-ended that reader on "Add a model folder" with no Reset, which is the exact conflation the component's own comment forbids.
- `docs/frontend_architecture.md` gains the two rules above.

## Tests

Four new ones, each verified to fail against `develop`: the store's `nothingSelected` and badge behaviour, and — at the layer the bug was actually seen — the shelf rendering an engine row under an engines-only selection, plus the narrowed-and-empty shelf offering Reset. Full frontend suite green (3116 tests).

## From the pre-push adversarial review

Two of its findings are answered above (the doc gap, the missing component-level test); the reordered empty state and the generalized `activeCount` came from it too. Its remaining point — that `defaultFilters`, `storedFilters`, `blockOf` and `fetchRows` still enumerate the blocks by hand — is left as is on purpose: each of those states something block-specific (a default, a validator, a bucket, a request), and folding them into one loop would trade a real distinction for a false symmetry.

### DIFF
--- a/docs/frontend_architecture.md
+++ b/docs/frontend_architecture.md
@@ -1562,6 +1562,23 @@ undo by accident:
   would have defeated the change on exactly the machines that had used the shelf
   longest, so `pixlstash:modelShelfFilters` gained `FILTERS_SCHEMA_VERSION` and
   a blob from another `v` is discarded whole — the same trade `storedView` makes.
+- **The four blocks are one list, and the empty states read it.** `nothingSelected`
+  and `activeCount` both derive from `BLOCKS`, never from a hand-written run of
+  `filters.adapters && filters.checkpoints && …`: naming three of the four is how
+  a shelf with **only Engines ticked** fetched its engines, counted them in the
+  toolbar, and then drew "Nothing is selected in Show" over the top of them.
+  `activeCount` therefore counts turning `engines` off exactly as it counts
+  `unclassified`, which also un-greys the panel's Reset button. The remaining
+  block-by-block lists (`defaultFilters`, `storedFilters`, `blockOf`,
+  `fetchRows`) each say something different per block and stay explicit.
+- **A narrowed selection that comes back empty is not "there is nothing here".**
+  Only the ticked blocks are fetched, so a shelf reopened with one empty block
+  selected has `rows` empty on a machine holding 1,800 adapters. The three empty
+  states are therefore ordered `nothingSelected` → *no models match these
+  filters* (whenever `activeCount` is non-zero) → the terminal *no models found*,
+  so the reader is only sent to "add a model folder" when nothing is narrowing
+  the shelf. Reset refetches every block and the terminal state then tells the
+  truth.
 
 **The assignment ring** (`assignmentRing` in `utils/modelShelf.js`, drawn by
 `ModelMark.vue`) is what the shelf says about who a model belongs to. It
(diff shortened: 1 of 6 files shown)


lindkvis:
## The red shard, and why the fix is in a Python test file

`backend shard 5` failed on `tests/test_read_token_security.py::TestResourceScopedReadTokenIsolation::test_list_all_tags_cannot_leak_out_of_scope_vocab`:

```
AssertionError: Set-A token tag count wrong (must reflect only Set A): {'close-up': 1, 'photo': 1}
```

The test posts `tag-a-only`, reads `/tags` back, and finds its tag gone and the *tagger's* vocabulary in its place. That is a `TagTask` — it deletes a picture's tags before writing its own — landing inside the test body.

**Not caused by the frontend change in this PR** (which touches no Python), and `develop` is green. It is a pre-existing race that a slow runner exposes.

**Root cause, measured rather than guessed.** The module fixture stops the planner and documents that as "no legitimate work left here". `WorkPlanner.stop()` stops the loop that *finds* work; the `TaskRunner` owns its own queues and threads and is untouched, so anything the planner submitted seconds earlier is still in flight. Probing the fixture at exactly that moment:

```
PENDING cpu: 0    PENDING gpu: 1    ACTIVE: ['FaceExtractionTask']
```

On this machine the pipeline finishes during setup, so shard 5 passes locally (425 tests) and the file passes on its own (68 tests) — which is why the failure only shows up on a runner slow enough to keep the pipeline alive into the first test body.

**Fix:** `_stop_planner_and_settle()` stops the planner, drains what is queued and waits for what is running, looped so a completion callback's follow-up is caught too. Both shared environments in the module use it. Same shape as the settle already in `tests/test_picture_mutation_scope.py`. After the fix the same probe reads `PENDING 0 / 0, ACTIVE []`, and the file still runs in 69s — the settle costs about two seconds of module setup and no assertion was touched.

The sibling modules that call `_work_planner.stop()` (`test_smart_score_invalidation.py`, `test_likeness_worker.py`) carry the same latent race; they are green and are left alone here rather than widening a bugfix into them.

Issue #326, opened 2026-05-25
Enable the loading of custom captioning engines


lindkvis:
The plugin system is there but I haven't yet extended it to load user-provided plugins.

HuntingSuccubus:
defiantly need it. qwen 3/3.5/gemma has impressive vision capabilities. there are NSFW ablirated/uncensored version available for nsfw caption.
Moondream2
Florence2 large promptgen by MiashouAI

gguf can be used by llama.cpp or llama python.

<img width="1098" height="1231" alt="Image" src="https://github.com/user-attachments/assets/744f02ac-784c-4f3c-827e-5651fae863a4" />
I am interested in red marked ones

lindkvis:
**Plan approved in Piecework.**

- plan: `pwl_c81838b33643426c`
- approved: 2026-08-14T15:57:48.365Z
- issue: 326

---

# Issue #326 — Enable the loading of custom captioning engines

## Context

`TaggerPlugin` (`pixlstash/tagger_plugins/base.py`) is the plugin ABC that both taggers
and captioners implement. It was **designed for third parties**: the file carries an
explicit MIT header in an otherwise GPL-3.0 repo, and its docstring says so — *"This file
is MIT-licensed so third-party plugin authors can depend on this API without pulling in
the GPL"* (`docs/licensing.md:22-24`).

Everything downstream of the registry is already generic:

- `DescriptionWorkflow._generate_batch_plugin` (`pixlstash/inference/workflows/description.py:80`)
  dispatches captioning to **any** registered plugin with `supports_descriptions`, merging
  `default_params()` with the user's configured params, calling the optional `setup(device)`
  hook then `init()` then `generate_descriptions()`, with a Florence-2 fallback on failure.
- `GET /taggers` (`pixlstash/routes/taggers.py:83`) already returns per-plugin
  `parameter_schema` and already renders failed plugins as rows with `load_error` set.
- `PluginsTable.vue` / `PluginParametersUI.vue` already build the parameter UI from that
  JSON schema, for any plugin.
- Settings validation (`pixlstash/utils/service/user_settings_utils.py:480-538`) is
  registry-driven — `known_plugins = set(manager.plugin_names())`. Its comments already
  read *"third-party plugin names must be registered and support tags"*.

The one thing missing is **discovery**. `TaggerPluginManager.reload()`
(`pixlstash/tagger_plugins/registry.py:50`) walks a hardcoded four-entry list:

```python
[...]

PR #951, opened 2026-08-15
Serve an adapter's bytes over the shelf API

lindkvis:
## Why

`GET /adapters` and `GET /adapters/{sha256}` serve `locations[].folder_path` and `relpath` — but those are **this** host's paths and name nothing on the machine that asked. A generator elsewhere on the network can therefore see every LoRA on the shelf and load none of them.

This adds the one route that closes that: `GET /api/v1/adapters/{sha256}/file`, which streams the adapter itself. It is the server half of the ComfyUI adapter loader in `Pikselkroken/ComfyUI-PixlStash` (issue #12, its PR #13) — the client side is written and waiting on this route, and until it lands that node only works when both processes share a filesystem.

## Authorization — the one decision worth reviewing

**`LOCAL_OWNER_ONLY`, not the `owner_only` tier every other shelf read sits on.** The sentence that kept the others there is *"they surface host paths but take none"*. This route takes none either — a sha256 the scanner already registered is the whole input, and the join of `model_folder.path` with `model_file.relpath` never touches caller data — but it does not *surface* a path. It returns the **raw bytes of a file inside a registered model folder**, which is `GET /model-folders/{folder_id}/runs/{run_name}/samples/{filename}`'s authority class exactly: reads inside a registered host root, writes nothing.

The correction recorded against that route on 2026-08-11 applies here too, so I am explicitly **not** making the argument it refutes: bytes are a new capability, not a narrower view of the metadata route beside them. "It is a subset of what `GET /adapters` already serves" is not available and is not being used.

The tier costs the feature nothing. Loopback, RFC1918 LAN and Tailscale all pass — that is every deployment the route exists for, including PixlStash in a container reached over the docker bridge. A genuinely remote generator needs `allow_remote_host_ops`, which is the safe direction to fail in for a route whose output is a multi-gigabyte file.

Locality total goes `31 = 26 + 5` → **`32 = 27 + 5`**; the tally test is renamed accordingly, §16.3 carries a dated update, and the coverage matrix has its row.

## Four narrowings inside the handler, none of them the tier

- **`present` copies only.** `missing` means the scan looked and found nothing, `unreachable` means the drive is unplugged, and a forgotten folder leaves its rows **tombstoned rather than deleted** — so any other state would serve bytes out of a folder the owner un-registered.
- **The join is contained**, with `path_is_within`, even though neither half is caller-supplied. On this route a `..` from a faulty scan or a restored hub would be an arbitrary-file reader over HTTP rather than a wrong row — the same argument B7 makes for containing its writes. Lexical first, deliberately: a model symlinked into a models directory is ordinary practice and realpath-only containment refuses every one of them.
- **409, not 404, for a known hash with no readable copy.** "No such adapter" tells a client to stop asking; "the file is not here right now" tells it to retry. Only one of those is right per case.
- **`Content-Disposition: attachment`.** Nothing in this app sets `X-Content-Type-Options`, so a served file is sniffable; the disposition is what keeps these bytes from being rendered as a document on our own origin.

The digest is deliberately **not** re-verified on the way out — that reads every byte twice per request, and the caller addressed the file by a hash it can check itself, which the ComfyUI client does.

## Verification

15 new tests in `tests/test_model_shelf_api.py`, in the module's existing warm environment (no new `Server`). Each guard was mutation-checked — the mutation named turns exactly the test named red, and nothing else:

| mutation | goes red |
|---|---|
| drop the containment check | `test_a_relpath_that_escapes_its_folder_is_refused_rather_than_served` |
| drop the `present` filter | `test_a_copy_the_scan_marked_missing_is_never_served` |
| realpath containment instead of lexical | `test_a_model_symlinked_into_the_folder_is_still_served` |
| drop the disposition | `test_the_bytes_are_served_as_an_attachment_and_never_as_a_document` |
| declare it `owner_only` | the declaration test, the remote-owner test, and the tally |
| delete `_enforce_unscoped_owner` from the gate's `LOCAL_OWNER_ONLY` branch | `test_no_share_token_can_download_a_model_file` |

That last row is the one I would look at first. `test_share_tokens_never_reach_a_folder_mutator` warns in as many words that **a GET added to this tier without a `READ_BLOCKED_GET_PATHS` entry is a real hole and nothing goes red** — and this route is a GET whose templated path that frozenset could not match in any case. So the gate's owner check is the only thing refusing a share token here, and it is asserted rather than reasoned about, both token shapes, each behind a live-credential control, against a file that genuinely serves 200 for the owner in the same test. The mutation above confirms the assertion is load-bearing; I have added a dated note to that docstring, because its "nothing goes red" claim is no longer true of that file.

The happy path runs end to end against a **real scan** — register a folder, rescan, download, and assert `sha256(body)` equals the hash asked for. Asserting the digest rather than the length is the point: a route serving *a* file of the right size would pass a length check and hand a generator the wrong weights.

`ruff format` / `ruff check` clean. Ran green locally: `test_model_shelf_api` (180), `test_authz_host_capability_16_3`, `test_architecture_guardrails`, `test_ci_shards`, `test_authz_gate_step4`, `test_model_shelf_scan_invariants` — 419 passed.

## What is not done

**The independent adversarial security sign-off has not been run.** CLAUDE.md requires one for any change to the registry's scope declarations, and requires that the author not be the one certifying it — so it is not something this PR can contain by definition. The self-audit above and the mutation table are evidence, not a sign-off. Please run it against this diff before merging.

Two residuals I did look at and am accepting, both stated so a reviewer can disagree:

- **TOCTOU on the served path.** A symlink planted inside a registered folder between the scan and the request would be followed. That is the case `path_is_within` documents as accepted — planting one needs filesystem write access to the models folder, which is a larger problem than this check is for — and the attacker would additionally have to get the file scanned and hashed to have a sha256 to address it by.
- **`Content-Disposition` filename.** `model.filename` reaches a response header. Starlette percent-encodes anything `quote()` changes into the `filename*=utf-8''` form, so a quote or a newline cannot break out; I checked rather than assumed.

No migration, no frontend change, no new dependency.


lindkvis:
Merged `develop` (through #937). Two conflicts, both of the same kind — and one of them was a real decision, not a text clash.

**The tally.** #326 put `GET /taggers/plugin-diagnostics` on the locality tier and claimed `32 = 27 + 5`; this PR claimed the same number for `GET /adapters/{sha256}/file`. Both notes are kept and the total is now **`33 = 28 + 5`**, with the download renumbered as the later of the two. Same resolution in `docs/backend_architecture.md` §16.3 and in the tally test, which is renamed to `test_host_capability_tier_split_is_28_local_5_loopback`.

**The decision.** #326 also added `test_every_untemplated_locality_get_is_on_the_read_blocked_belt`, which derives the belt from `ROUTE_POLICIES` and pins the templated locality GETs — the ones an exact-match frozenset cannot express — as a known set of two, so that adding a third fails the build rather than waiting for a review to notice. This route is that third, and it is **the sharpest member of the set**: the other two serve a run listing and a preview image, and this one streams model weights, so under the documented `AUTHZ_GATE_ENFORCING = False` rollback a share token would not read a directory — it would download every adapter on the shelf.

I added it to the pinned set rather than closing the gap. Closing it means prefix matching in a belt every request passes through, which is its own change with its own review, and a bespoke `startswith` for one route is the special case that rots. The reasoning is recorded in the guardrail's docstring and in §16.3 rather than left in this comment. **To be explicit about what that does and does not claim:** the gate refuses a share token here today, and `test_no_share_token_can_download_a_model_file` proves it by mutation. The note is about the rollback, not about today.

Re-ran locally after the merge: `test_authz_host_capability_16_3`, `test_architecture_guardrails`, `test_model_shelf_api`, `test_authz_gate_step4`, `test_ci_shards` — 401 passed. `ruff check` clean, `render_backend_architecture.py --check` clean.

The independent adversarial security sign-off is still outstanding, and this merge is a reason to want it rather than a step toward it: two changes landed on the same tier in the same day, each unaware of the other.

PR #964, opened 2026-08-15
Complete the base model, and edit it on the row

lindkvis:
`base_model` is free text, 37 % of real adapters record none, and the only way
to set one was a dialog with an empty box. This completes the field and lets the
row edit itself.

## What changed

**`GET /api/v1/models/base-models`** serves `known_base_models.completions()`:
the canonical labels the module ships — so the field is useful on a fresh
install that records nothing yet — plus every distinct `base_model` already
stored here that folds to none of them. `completions()` was written for this and
had no caller. The whole list, once, no prefix parameter: it is a few dozen
strings and the field filters them as you type, which beats a request per
keystroke. `OWNER_ONLY` on the shelf's default library pin, declared in the
registry, tabled in the coverage matrix, both directions asserted.

**`BaseModelInput.vue`** is the field, used in two places. Matching folds case,
spacing and punctuation away, so `sdxl` finds `SDXL 1.0`, and prefix matches
sort ahead of substring ones. ArrowDown opens the menu, Tab fills without
committing, Enter commits taking the highlight, the first Escape closes the menu
and only the second reaches the host. It offers and never constrains: the column
stays free text, so a base model released after this build is typed and stored
verbatim.

**The Base cell is now editable in place** — double-click, or **Shift+F2** for
the keyboard — including the `not set` chip, which is a value like any other and
is the row that most needs correcting. Committed on Enter or blur, abandoned on
Escape, exactly like the inline rename beside it. The bulk dialog keeps its
counted overwrite warning.

The client skips a row whose `base_model_folded` is set when it unions the rows
on screen into the completion list: that is the server's own rule (`completions`
drops an extra that folds to something it ships) applied to the field the server
already computed, so the client cannot hand back the aliases the server
deliberately dropped.

## Answers to the review this got before it was pushed

Two adversarial passes ran over the diff. Most findings are fixed in the second
and third commits; these four I disagree with, and would rather say so here than
obey in silence.

- **"The inline editor writes every member of a stack with no overwrite
  confirmation, unlike the dialog."** Deliberate, and the same rule the inline
  rename already ships: a cover *is* its run, and one training run was trained
  against one base model. The dialog's warning counts values it will destroy
  across a selection the reader cannot see at a glance; a single row is a
  gesture on one visible thing. It is the one part of this change that can lose
  curation, so it is the part worth arguing about if you disagree.
- **"`BaseModelInput` duplicates `OverlayTagsPanel`'s dropdown."** The menu's
  ~40 lines of CSS are copied on purpose — two completion menus that looked
  different would be two things to learn — but the logic is not shared, because
  the tag panel's list is wired into predictions, rejected-tag confidences and
  the image's own tags. Prising it out is a refactor of a component this change
  does not otherwise touch. Collapse them if a third completion field appears.
- **"A `<datalist>` would have been free."** It cannot match on the folded
  spelling, which is the whole feature, and it renders in browser chrome rather
  than the app's theme. The reasoning is now in the component's docstring.
- **"The matrix cell words the access tier out instead of naming the enum."**
  The automated secret scan false-positived on the backticked identifier in that
  prose cell and held the push; the tier is named in words there and still named
  in `registry.py`. Reverting the wording would re-block the publish.

A failed completions fetch was retrying once per keystroke, and the list
survived a session reset — both found by the second pass, both fixed, both now
asserted.

## Testing

Frontend 3210 across 174 files; backend shelf, guardrail and known-base-model
suites green; ruff clean; the generated route table re-rendered and `--check`
clean. Every fix above has an assertion that fails when the fix is removed —
checked by mutation, not by reading: the fold skip, the retry throttle, the
session reset, `aria-controls`/`aria-activedescendant`, Shift+Tab, the scroll
close, the focus handoff, commit-on-blur and the two-stage Escape.


lindkvis:
Note for review — a test gap I found after the branch was pushed, left unfixed to avoid another push.

`BaseModelInput.test.js`'*s "matches on the folded spelling, not the literal one" does not actually pin folding. It types `sdxl` and expects `SDXL 1.0`, which plain lowercasing already satisfies, so the assertion stays green if the `.replace(/[^a-z0-9]/g, "")` in `norm()` is deleted.

The inputs that separate the two rules — each finds its label with folding and nothing without it:

| typed | lowercase only | folded |
|---|---|---|
| `sdxl` | `SDXL 1.0` | `SDXL 1.0` |
| `flux1` | — | `FLUX.1 dev` |
| `zimage` | — | `Z-Image Base` |
| `sdxl1` | — | `SDXL 1.0` |

Swapping the test input to `flux1` (or adding it as a second case) closes it. The shipped behaviour is correct — typing `flux1` does find `FLUX.1 dev` today; what is missing is the tripwire that stops a later edit removing it quietly. Worth folding into a review round rather than a commit of its own.

Issue #1296, opened 2026-09-13
Design cleanup: fields and choice controls


lindkvis:
- One field height, `--control-h`, everywhere in the shell.
- No label inside the field.
- `--font-mono` on path, hash, port and id fields.
- `Segmented` for two to five choices, `OptionRows` for sort-by. Pick one per site.
- Option rows take no fill: a trailing check and an olive mark.
- Segmented track: trough, inset ring, concentric corners.
- `v-text-field` goes, as `v-btn` does.

Specs in `docs/design/buttons.md`.

Depends on: #1294

Issue #1294, opened 2026-09-13
Design cleanup: selection, focus and hover states


lindkvis:
Bring the app onto the approved colour and state rules in `docs/design/buttons.md` and `docs/design/visual-language.md`.

- Amber acts, olive selects. Selected tiles, rows and tabs are olive.
- No olive text on an olive wash: selected rows take `--text`. Olive stays on the check, underline or icon.
- Focus is a 2px `--text` ring with a 2px gap, on everything, from its own per-theme token.
- Hover is a `--text` wash at 16%; fills darken 20%.
- `--accent-on` becomes white.
- Retire `primary_green`; its sites use `primary`.

Done when no component carries its own selection, focus or hover colour.

PR #1327, opened 2026-09-13
Design cleanup: two button dialects, and shared selection/focus/hover states

### LINKED ISSUE #1295: Design cleanup: two button dialects
Replace the parallel button styles with the two approved dialects in `docs/design/buttons.md`.

- Raised, for dialogs, panels and forms: 28px, compact 24px, `--radius-sm`.
- Flat, for the toolbar and the selection bar: 32px.
- Everything else that is pressable becomes a menu row, a choice control or an in-content affordance, as that file classifies it.
- `AppButton` replaces `v-btn` everywhere.

Only the rows marked approved. The open ones stay open.

Depends on: #1294

lindkvis:
Closes #1295 and #1294 (done together: the button dialects need the new state tokens).

**Buttons (#1295)**
- Tokens `--control-h` 28, `--control-h-sm` 24, `--control-h-bar` 32. `AppButton` moves to 28/24 on `--radius-sm`, gains `block` and an `outline` variant.
- New `AppBarButton` for toolbar, selection bars and close/dismiss (paint stays in the global `.bar-btn*` family).
- Every `v-btn` → `AppButton` (the two inside `v-btn-toggle` stay; that's #1296). The parallel families (`.gbtn`/`.gcompare`/`.dq-btn`/`.qdecided`, `.btn`/`.dlg-btn`, the filled one-offs, `.selbar-btn`, bespoke closes) are gone.
- Left alone as still open: on-dark buttons (`.rs-*`, lightbox chrome), `.tbm-action` in popovers, outlined → filled neutral (kept as `outline`). TitleBar window controls stay native.

**States (#1294)**
- Focus: one global 2px ink outline with a 2px gap (`--focus-stroke`); component focus colours removed.
- Hover: `--hover-wash` ink 16%, `--hover-shade` on fills, `--hover-neutral` on grey fills. No more `brightness()` hovers.
- Selection is olive (`--active-*`, `--selected-ink`), words in ink. `on-accent` is white; `primary_green` retired.
- `App.vue` also puts the theme class on `<html>`, so popups teleported to `<body>` resolve these tokens.

**Pre-push review, answered here**
- White on amber is 3.41:1, under the 4:1 text floor. That's the approved exception in `docs/design/buttons.md`; hover takes it to 5.03:1.
- `docs/design/buttons.md` is cited but not in the repo yet. It's the approved spec and needs committing separately.
- Two assertions were dropped on purpose: DedupGroupRow's `.gbtn:disabled` opacity (the class is gone; `AppButton` owns disabled) and the notice dismiss's 40px `::before` hit area (the button is 32px now).
- Decided but not changed: amber stays on rating stars, progress fills, attention badges and callout tints, since none of them is selection, hover or focus.

Vitest is green (230 files). eslint and the build pass. e2e page objects are updated but e2e wasn't run locally.


frontend/src/components/widgets/AppBarButton.vue
```diff
@@ -0,0 +1,97 @@
...
+    ]"
+    :disabled="disabled || loading"
+    :aria-busy="loading ? 'true' : undefined"
+  >
+    <span v-if="loading || icon" class="bar-icon-badge-wrap">
+      <!-- 24px when the glyph is the whole control, 18px beside a label: MDI
+           is drawn on a 24-unit grid, and a labelled glyph tracks its text. -->
+      <v-icon
+        :size="iconSize || (iconOnly ? 24 : 18)"
+        :class="{ 'mdi-spin': loading }"
```
lindkvis, line 29: [P3] Preserve the loading status animation under reduced motion

The PR's new `visual-language.md` §11 explicitly says the shared loading spinner must keep spinning under `prefers-reduced-motion`, because it is a status readout. `design-tokens.css` otherwise forces every animation to 0.001 ms and one iteration. `AppButton` restores `mdi-spin::before`, but this new primitive has no equivalent rule. In headless Chromium with reduced motion enabled, this glyph computed to `animation-duration: 1e-06s`, `animation-iteration-count: 1`, and a settled `transform: none`, while AppButton remained `2s`/`infinite`. The changed SelectionBar exercises this through `:loading="clearingImpossible"`; its text still changes to “Clearing…”, so the impact is limited, but the promised spinner renders as a frozen ring for reduced-motion users. Please add the same narrowly scoped media override for `.bar-btn--loading .mdi-spin::before` (in this component or the global bar-button stylesheet).

lindkvis (commented):
Senior frontend review of `c5fa5941`. I found one low-severity accessibility/status defect and left it inline. I traced the AppButton/AppBarButton prop, attribute, listener, loading, disabled, focus-restoration and menu-activator contracts; checked the migrated selection, undo/redo, toggle, tooltip, ARIA and event-modifier paths; and audited the global state tokens plus scoped/global CSS boundaries.

Focused validation: 575 tests passed across 10 component suites covering both primitives, SelectionBar, ShelfSelectionBar, Toolbar, UndoControl, ActionReceipt, SearchResultBar, ModelShelf, DuplicateQueue and DedupGroupRow. The exact-head CI checks and E2E jobs are green; all reported backend jobs are green except backend shard 1, which GitHub still shows in progress. A targeted headless-Chromium check with `prefers-reduced-motion: reduce` reproduced the inline finding. I did not perform a broad manual visual sweep of every migrated screen or live backend operation.

frontend/src/components/widgets/AppBarButton.vue
lindkvis: Fixed in a97bac68. The reduced-motion override now lives beside the `.bar-btn` family in `App.css` as `.bar-btn .mdi-spin::before` (2s, infinite), so it covers the `loading` glyph and any spinner a call site slots into a bar button. The model shelf's Add spinner is one of those, and its own scoped copy of the rule is deleted. Checked in headless Chromium with reduced motion on: both cases compute to `2s infinite`. With the rule removed they compute to `1e-06s 1`, which matches what you measured.

PR #1329, opened 2026-09-13
PluginsTable: re-selection test that reaches setActive; v-model:settings

### LINKED ISSUE #1257: PluginsTable: two tests named for the removed toggle-off do not exercise it
Follow-up to #1253. Nothing here is broken at runtime — the problem is that two
tests are named for a behaviour they do not exercise, so they will read as
coverage for it indefinitely.

## The behaviour they claim to cover

#1253 removes the toggle-off from `PluginsTable.setActive`, because the new
explicit "None" row is now how you turn a capability off:

```diff
-  // Toggle off if already active
-  const next = activePlugin.value === pluginName ? null : pluginName;
```

`describe("PluginsTable re-activation is inert")` in
`frontend/src/components/widgets/PluginsTable.test.js` is the block that reads
as the guard for that removal.

## Why neither test reaches it

`setActive` is wired to `@change`, and neither test produces a `change` event
that reaches it:

- **`does nothing when the already-active plugin is clicked again`** clicks the
  radio that `mountTable("wd14")` already renders `:checked`. jsdom's radio
  activation behaviour fires `input`/`change` only when the checked state
  actually changes. It does not, so no `change` event exists and `setActive` is
  never called. `patchUserConfig` is uncalled for a reason unrelated to the
  toggle.
- **`does not clear the selection on a double click`** clicks an unchecked radio
  (one `change`, one call), then triggers twice more. But `setActive` sets
  `settingActive.value = true` before awaiting, the radio is
  `:disabled="settingActive"`, and vue-test-utils will not dispatch on a
  disabled element. `toHaveBeenCalledTimes(1)` is measuring that guard, not the
  toggle.

Both pass with the toggle-off patched back into `setActive` — 10/10 green
against the mutation.

## Why there is nothing there to test as written

The property the block is named for is a *browser* guarantee: a real click on an
already-checked radio emits no event, so no component code runs. There is no
component behaviour to assert.

What #1253 actually changed is what `setActive` writes when it *is* called for
the already-active plugin, which is testable by firing the `change` a browser
would not:

```js
it("re-selecting the active plugin does not clear it", async () => {
  const w = mountTable("wd14");
  await pluginRadios(w)[0].trigger("change");
  await flushPromises();
  expect(patchUserConfig).not.toHaveBeenCalledWith({
    tagger_settings: { active_tag_plugin: null },
  });
});
```

Verified both directions: passes on #1253's branch, reds under the
toggle-restored mutation.

## Fold in: `v-model:settings`

Same file pair, and it removes the failure mode #1253 fixed rather than only
the instance:

```diff
-            :settings="taggerSettings"
-            @update:settings="(s) => (taggerSettings = s)"
+            v-model:settings="taggerSettings"
```

Both call sites in
`frontend/src/components/settings/BehaviourSection.vue`. `PluginsTable` already
declares a `settings` prop and an `update:settings` emit, which is exactly the
contract `v-model` with a named argument compiles to:

```js
{
  settings: taggerSettings.value,
  "onUpdate:settings": $event => ((taggerSettings).value = $event)
}
```

The point is not the two saved lines — it is that the assignment which was
written wrong twice stops being hand-written.

## Scope

- Replace the two tests in `describe("PluginsTable re-activation is inert")`.
- Swap both handlers in `BehaviourSection.vue` to `v-model:settings`.
- Confirm the replacement test reds under the toggle-restored mutation before
  calling it done.

Land after #1253 merges; the tests it replaces arrive with it.

lindkvis:
Closes #1257.

- Replaces the two \`re-activation is inert\` tests, which never reached \`setActive\` (a click on a checked radio fires no \`change\`; the double click was stopped by \`:disabled\`). The new test fires \`change\` on the active plugin and asserts it is written and emitted as still active. It goes red with the toggle-off restored, including a toggle to \`undefined\`, and when \`setActive\` does nothing.
- \`BehaviourSection.vue\`: both tables use \`v-model:settings\` instead of hand-written handlers. Updated the comments in its test to match.

Review notes:
- The deleted double-click test did in effect cover the in-flight \`:disabled\` guard. The issue scopes it out, so nothing replaces it here. It can be a separate test if wanted.
- `docs/frontend_architecture.md` still names `TagPluginsTable`/`DescriptionPluginsTable`. That drift is older than this change and is left out of scope.


lindkvis (commented):
Careful re-review at `491b09e8`: no actionable findings introduced by this PR.

The two named `v-model:settings` bindings compile to the same setup-state assignments as the previous explicit handlers, and `PluginsTable` emits updated settings only after its PATCH succeeds. I ran the 14 focused tests plus three additional review probes (17/17): sequential sibling-table synchronization, failed-save/retry behavior, and native checked-radio behavior. A temporary toggle-off mutation made the replacement test fail as intended. Headless Chromium also confirmed that re-clicking an already checked radio emits no `change` event.

All 17 executed CI checks passed; one informational release-preparation job was skipped. I did not run the complete application manually in a browser. A same-tick sibling-save edge case reproduces unchanged on base `5b9a700e`, so I excluded it as pre-existing rather than treating it as a finding here.

Issue #1223, opened 2026-09-07
v1.11.1 review leftovers: pre-existing, none blocking


lindkvis:
Things the v1.11.1 reviews found and nobody fixed. **All pre-existing — none is
a regression from the PRs that surfaced them** (#1216, #1219, #1221), which is
why they are here rather than in those diffs.

Not urgent. Triage before starting.

---

**1. Containment is one-directional.** `refuse_path_inside_a_library`
(`pixlstash/utils/library_roots.py`) refuses a path *inside* a library root. It
does not refuse one that *contains* a root. Only `image_root` gets the reverse
check, via the surviving lexical test at `pixlstash/routes/import_folders.py:193`
(`image_root.startswith(folder + os.sep)`) — every other root has no reverse
coverage.

Left as-is deliberately: making it bidirectional would newly refuse exporting
to, say, your home directory. Wants a decision about which roots deserve the
reverse test, not a blanket change.

**2. `register_reference_folder` bypasses the shared refusal.**
`services/folder_structure_commit_service.register_reference_folder` calls
`validate_reference_folder_conflicts` directly rather than going through the
router, so the new blocklist does not apply. It only registers an existing
folder for in-place indexing and moves nothing, so the hazard is smaller — but
it is the one path that skips the check.

**3. `GET /server-config/watch-folders` has no callers.** Delete it or wire it
up. Reviewed and found not to be a security concern on its own: an owner-only
read of paths the owner already sees in the import-folders UI. The argument for
deleting is maintenance, not exposure. This is a product call, which is why it
was left alone twice.

**4. The `DELIBERATELY_OPEN` reasons are unverifiable by test.**
`electron/test/urlPolicy.test.ts` now requires every IPC channel to carry a gate
or a written reason, and 22 carry reasons. A test can check a reason *exists*;
it cannot check the reason is *true*. Worth one human read of that map.

**5. `playwright test --ui` bypasses the production route block.** UI mode
reuses one context via `_newContextForReuse()`. The `context` fixture override
added in #1216 covers it, but a reused context keeps `localStorage`, so the 24h
throttle applies after the first check — roughly one check-in per UI session.
Developer-local only. Noted, not obviously worth more.

**6. Two claims that only a manual dispatch can settle.** `docker-build.yml` is
`workflow_dispatch`-only, so nothing has yet proven the smoked container reports
`install_type == "docker"` under the assertion added in #1216. One dispatch
answers it.

---

[...]

PR #1330, opened 2026-09-13
Watch and reference folders may overlap each other, never a library (#1223)

lindkvis:
Refs #1223 (item 2, plus the owner's ruling on which folders may overlap).

**Rule:** watch and reference folders may overlap each other. Neither may equal, sit inside or contain a library's folder: the active library's or any registered one's, detached included. A library may not be added inside, or around, a watch or reference folder of any registered library: `LibraryRegistry.create`/`attach`/`relocate` enforce it, so `POST /libraries` and the CLI alike, reading each library's vault file. The check and the write hold one in-process lock, `FOLDER_OVERLAP_LOCK`, shared with the watch and reference folder writes.

- `utils/library_roots.py`: `refuse_folder_overlapping_a_library` is the new two-way check. `refuse_path_inside_a_library` (the full list) now serves only the folder export, which still may not write into watch or reference folders.
- The rule is applied by `POST /import-folders`, the reference-folder create, PATCH and relocate routes, and `register_reference_folder`. That last one is the folder-structure commit, which used to skip the route's check (item 2).
- `GET /libraries/inspect` / `POST /libraries`: a folder overlapping a watch or reference folder gets verdict `overlaps` with `library: null`. The picker only reads `headline`, `detail` and `can_add`.
- `validate_reference_folder_conflicts` is unchanged: reference folders still may not overlap each other or `image_root`.

Tests in `tests/test_library_destination_blocklist.py`. Each refusal was checked by mutation: removing it from import folders, reference create, PATCH, relocate, the commit or inspect, or making the check one-way, fails a test.

Known limits and review notes:
- A `delete_after_import` watch folder over a reference folder is now allowed. Files the reference scan has already indexed are skipped by hash. Files it has not reached yet are imported and deleted from the reference folder. This follows from the rule as asked; it deserves a UI warning if it is kept.
- A watch folder that contains any registered library is now refused, because the watcher recurses. The message does not name the library.
- A CLI in another process racing a running server is not serialised; the lock is in-process.
- A registered library whose vault file is not a database is ignored by the check (nothing can be reading it). A locked vault refuses.
- When a vault cannot be read, inspect answers `can_add: false` ("Could not check this folder") and `POST /libraries` answers 400.
- Paths stored in their Docker host form are compared as they were before; not changed here.

Item 7 landed in #1255. Items 1, 3, 4, 5 and 6 are not in this PR.


lindkvis (commented):
Careful backend review of `45bf1f20`. I found two P2 gaps in the new folder-overlap invariant: the documented CLI mutation paths bypass it, and the route-level checks are not coordinated with the writes they protect.

Validation: 76 focused changed/path tests passed; three isolated real-server probes reproduced the CLI bypass, concurrent registration race, and active-library-switch stale read; Ruff lint/format and `git diff --check` passed. Exact-head CI also passed `checks`, E2E, all 8 Linux backend shards, and all 4 Windows backend shards. This was backend-only, so no browser verification was applicable.

pixlstash/routes/libraries.py
```diff
@@ -479,6 +486,45 @@ def _inspect(
...
 
+        # A library inside, or around, one of this library's watch or reference
+        # folders would index or move the files that folder imports or scans
+        # (#1223). Only the active library's folders are known here; another
+        # library's live in its own vault. With no library open (a failed
+        # switch) there are no folders to read, and adding a library may be the
+        # way out of that state.
+        try:
+            folders = (
+                watch_and_reference_folders(server.vault)
```
lindkvis: **[P2] Enforce this overlap rule for CLI library mutations too**

This guard only runs through `_inspect`, so the documented CLI entry points bypass it: `libraries create` and `libraries attach` call `LibraryRegistry` directly, and `libraries relocate` does as well. In a real-server probe I registered watch root `W`; HTTP `POST /libraries` correctly returned 409 for `W/nested`, but `cli.main([... libraries, create, W/nested ...])` returned `EXIT_OK` and created `W/nested/vault.db`. That leaves the changelog's new invariant false for a supported script/cron interface and preserves the recursive `delete_after_import` hazard. Please route create/attach/relocate through shared overlap validation and cover the CLI path with a regression test.

pixlstash/routes/libraries.py
```diff
@@ -479,6 +486,45 @@ def _inspect(
...
+            )
+            return LibraryInspection(
+                verdict="overlaps",
+                path=folder,
+                can_add=False,
+                headline="Could not check this folder",
+                detail=ROOTS_UNAVAILABLE_DETAIL,
+                suggested_name=suggested,
+            )
+        if any(
```
lindkvis: **[P2] Keep this folder snapshot valid until registration commits**

The overlap read and the later registration are not coordinated across the hub and active vault. A deterministic two-client probe paused both checks before either write: concurrent `POST /import-folders` for `W` and `POST /libraries` for `W/nested` then both succeeded (200 and 201). A second probe paused this request after reading library A's folders, switched to B (whose watch root was `W`), then resumed; the add still returned 201 inside B's watch folder because `POST /libraries` remains `HUB_ONLY`, holds no active-vault lease, and may straddle a switch. Please hold the appropriate vault lease and serialize, or post-validate and roll back, competing library/watch/reference mutations so validation and registration form one coordinated operation.

pixlstash/routes/libraries.py
lindkvis: Fixed in 486a9d68. The rule now lives in `LibraryRegistry`: `create`, `attach` and `relocate` call `refuse_overlapping_watch_or_reference_folder` (raises `LibraryOverlapError`), so `POST /libraries` and the CLI verbs share one check. It reads every registered library's watch and reference folders straight from its vault file, detached included, so it no longer depends on which vault the server has open. `_inspect` calls the same method, so the picker still says so before Add.

Exempt with `check_folder_overlap=False`, because a refusal there would stop start-up or wedge a discard: start-up's own registration, the scratch library, and restore's relocate.

Test: `test_the_cli_library_verbs_apply_the_same_rule` (`libraries create` inside a watch folder is refused and writes no vault.db; `attach` is refused; a free create still works). Removing the registry check turns it red.

pixlstash/routes/libraries.py
lindkvis: Fixed in 486a9d68.

- **One lock:** `FOLDER_OVERLAP_LOCK` (`hub/registry.py`, an in-process RLock) is held from the check to the commit by the registry verbs and by the watch/reference folder writes: `POST /import-folders`, the reference-folder create, PATCH and relocate routes, and `register_reference_folder`, via `@holding_folder_overlap_lock`.
- **Switch straddle:** the library-side check now reads all registered libraries' vault files rather than the active vault, so a switch mid-request cannot change the answer.
- **Test:** `test_a_library_and_a_watch_folder_added_at_once_cannot_both_land` pauses a library add inside `create` right after its check and starts a watch-folder add around it. The watch add must still be blocked after 0.5 s, and must then get 409 while the library gets 201. Removing the lock from the guard or from the import-folder route turns it red.
- **Also:** `test_a_watch_folder_of_another_library_blocks_a_library_too` covers a watch folder recorded only in a non-active library's vault.
- **Not covered:** a CLI in a separate process racing a running server. The lock is in-process, and the watch-folder row commits in the vault while the library row commits in the hub, so there is no shared transaction to lean on. This is stated in docs/backend_architecture.md.

PR #1340, opened 2026-09-14
Design cleanup: one tooltip surface, icon sizes, z-index and off-token values

### LINKED ISSUE #1299: Design cleanup: tooltips, icons and z-index
- One `Tooltip` surface. `HelpTip` becomes a preset of it.
- A `tooltip` prop sets the accessible name and the tip from one string.
- Native `title` stays only to reveal clipped text.
- Icons have no scale: a component owns its icon slot. A free-standing icon follows its text, 16px by default. An icon-only control 32px or taller takes a 24px glyph.
- The z-index ladder stops at `--z-drawer`.
- No sub-pixel type. Progress tracks take `--radius-pill`.
- A raw value that equals a token is a bug: use the token.

Specs in `docs/design/buttons.md`.

Depends on: #1295

lindkvis:
Closes #1299.

**Tooltips**
- New `Tooltip` widget: the one surface (`surface` ground, `--radius-md`, `--elevation-3`, `--tooltip-max-w`). Hover waits `--tooltip-delay`, focus opens at once, Escape closes from anywhere, pressing the control closes it, and the tip is hoverable. `HelpTip` is now a preset of it.
- `AppButton` and `AppBarButton` take `tooltip`. On an icon-only button it is also the `aria-label`; an explicit `aria-label` still wins. `AppButton`'s `title` prop is gone.
- About 450 native `title`s and all 10 `v-tooltip`s moved to it. `title` stays only where it reveals clipped text.
- Vuetify writes the tip's id over a control's own `aria-describedby`, so a blocked button lost the link to its reason. `Tooltip` gives the value back and describes through `aria-description` instead.
- `activator="parent"` tips build their `v-tooltip` on first hover or focus. Built up front, each cost ~20x a `title` to mount, and a grid tile carries several. It is ~2x now.

**Icons, layers, off-token values**
- Icon sizes follow the three rules in `buttons.md`. Bar and lightbox icon-only buttons go to 24px, menu glyphs to 16px, and the dialog close button stays 20px.
- Raw z-index values equal to a rung use the token. Hand-written values above `--z-drawer` are gone except five teleported dropdowns and previews. Those have to become overlays, so they are listed as debt in the guard.
- No sub-pixel type. Raw sizes and radii equal to a token use the token, and progress tracks use `--radius-pill`.
- `styles/designDrift.test.js` guards all of the above.
- `visual-language.md` §8/§14 and `frontend_architecture.md` are updated. e2e selectors move from `[title=…]` to `[aria-label=…]`.

**Pre-push review, answered**
- *Tips on natively disabled buttons may never open.* True where the browser fires no mouse events on a disabled control. Before, the `title` showed for the mouse and keyboard users never got it. The fix is `aria-disabled` or a `HelpTip` per site, which is a UI/UX call. Not done here.
- *Stacking reordered in the review overlay (4210 to 4400 flattened) and the lightbox draw layers.* Deliberate. The overlay is its own stacking context, and the dialogs that now tie can't be open at the same time. Order inside one context now comes from the DOM, per the spec.
- *Off-token values were rounded to a rung, e.g. a 0.55rem hint became 11px.* That is the "no sub-pixel type" rule. Radii that weren't equal to a token moved only where the tier rule clearly applied.
- *The guard doesn't read JS-set styles* (the drag ghost's `zIndex`). Stated in its docblock and left out of scope.
- *`aria-description` is ARIA 1.3.* Chromium, which Electron ships, supports it. A tip is never the only home of information either way.

Vitest is green (232 files, 4082 tests); eslint has no new warnings and the build passes. e2e wasn't run locally.


frontend/src/components/panels/SelectionMenu.vue
```diff
@@ -14,39 +14,53 @@
...
-        :title="groupingLockReason || undefined"
-        :readonly="isReadOnly"
-        :locked-set-ids="lockedSetsStore.lockedSetIds"
-        @added="$emit('added-to-set', $event)"
-      />
+        :text="groupingLockReason || ''"
+      >
+        <template #activator="{ props: tipProps }">
+          <AddToEntityControl
+            v-bind="tipProps"
```
lindkvis: **HIGH — `v-bind="tipProps"` before `ref=` throws away Vuetify's activator ref, so the tip has no anchor.**

Vue's `mergeProps` has no special case for `ref` — last writer wins (`runtime-core.cjs.js:7735`). `VOverlay` puts its own `ref: activatorRef` in the slot props (`VOverlay.mjs:244`). Here `ref="ateProjectRef"` is merged *after* it, so `activatorEl` is never set, `target` stays undefined, and `updateLocation` bails on `if (!data.target.value …) return` (`locationStrategies.mjs:167`) — the tip opens unpositioned at the overlay container's origin instead of beside the control.

Repro: read-only session (so `groupingLockReason` is set), hover Project/Character/Set in the selection menu.

`SideBar.vue:1566` already ships `mergeTooltipRef()` for exactly this collision. Same defect at lines 37 and 52 of this file.

frontend/src/components/panels/SelectionMenu.vue
```diff
@@ -14,39 +14,53 @@
...
+            :disabled="selectedCount === 0 || !!groupingLockReason"
+            :readonly="isReadOnly"
+            @selected="$emit('set-project', $event)"
+          />
+        </template>
+      </Tooltip>
+      <Tooltip :text="groupingLockReason || ''">
+        <template #activator="{ props: tipProps }">
+          <AddToEntityControl
+            v-bind="tipProps"
```
lindkvis: **HIGH — same ref collision as line 23.** `v-bind="tipProps"` merged before `ref="ateCharacterRef"`, so Vuetify's `activatorRef` is discarded and the tip opens unanchored.

frontend/src/components/panels/SelectionMenu.vue
```diff
@@ -14,39 +14,53 @@
...
+            :readonly="isReadOnly"
+            @added="$emit('add-to-character', $event)"
+            @removed="$emit('remove-from-character', $event)"
+          />
+        </template>
+      </Tooltip>
+      <Tooltip :text="groupingLockReason || ''">
+        <template #activator="{ props: tipProps }">
+          <AddToEntityControl
+            v-bind="tipProps"
```
lindkvis: **HIGH — same ref collision as line 23.** `v-bind="tipProps"` merged before `ref="ateSetRef"`, so Vuetify's `activatorRef` is discarded and the tip opens unanchored.

frontend/src/components/views/ImageOverlay.vue
```diff
@@ -247,32 +252,44 @@
...
-            :locked-set-ids="lockedSetsStore.lockedSetIds"
-            :class="{ hidden: chromeHidden }"
-            @added="(payload) => emit('added-to-set', payload)"
-          />
-          <AddToEntityControl
+            :text="stackGroupingLockReason || ''"
+          >
+            <template #activator="{ props: setTipProps }">
+              <AddToEntityControl
+                v-bind="setTipProps"
```
lindkvis: **HIGH — same ref collision.** `v-bind="setTipProps"` is merged before `ref="addToSetControlRef"`, so the app's ref overwrites Vuetify's `activatorRef`. `activatorEl` is never set and `updateLocation` bails, so the tip opens at the overlay container's origin rather than beside the control. `mergeTooltipRef()` (`SideBar.vue:1566`) is the existing fix.

frontend/src/components/reviews/ReviewBinaryCard.vue
```diff
@@ -54,24 +57,35 @@
...
-          :alt="`picture ${item.picture_id}`"
-          title="Click to zoom"
-          @click="openZoom(item.picture_id, item.picture_ext)"
-          @load="onImgLoad"
-          @error="onImgError($event, item.picture_id)"
-        />
+        <Tooltip text="Click to zoom">
+          <template #activator="{ props: tipProps }">
+            <img
+              ref="imgRef"
```
lindkvis: **HIGH — same collision, opposite direction: `imgRef` is now permanently null.**

`ref="imgRef"` is merged *before* `v-bind="tipProps"` (line 67), so Vuetify's `activatorRef` wins and the app's ref loses. Consequences, all silent: `measure()` early-returns on `if (!el)`, `watch(imgRef, …)` never fires, the `ResizeObserver` is never attached, `boxW`/`boxH` stay 0, and `renderedRect` is `null` forever. The contain-rect readout in the review binary card is dead.

frontend/src/components/editors/PictureSetEditor.vue
```diff
@@ -24,35 +24,47 @@
...
-          label="Description"
-          placeholder="Optional description…"
-          :rows="2"
-          :disabled="isLockedSet"
-          :title="isLockedSet ? LOCK_REASON : undefined"
-        />
+        <Tooltip :text="isLockedSet ? LOCK_REASON : ''">
+          <template #activator="{ props: tipProps }">
+            <AppInput
+              ref="nameInputRef"
```
lindkvis: **MEDIUM — `nameInputRef` is null, so the picture-set dialog no longer focuses its name field.**

`ref="nameInputRef"` before `v-bind="tipProps"`, so Vuetify's `activatorRef` overwrites it. Line 324's `nameInputRef.value?.focus?.()` / `.select?.()` no-op; the optional chaining means no error, just a dialog that opens with nothing focused.

lindkvis (commented):
Review of the tooltip sweep. The sweep itself is careful — the findings below are one recurring mechanism plus four independent issues.

**The through-line:** `v-bind` and a template `ref` on the same element. Vue's `mergeProps` has no special case for `ref`, so last writer wins, and one of the two refs is silently discarded. Four sites, two directions:

- `v-bind` first -> Vuetify's `activatorRef` is lost -> tip opens unanchored (`SelectionMenu.vue` x3, `ImageOverlay.vue`)
- `ref` first -> the app's ref is lost -> `imgRef` / `nameInputRef` are permanently null (`ReviewBinaryCard.vue`, `PictureSetEditor.vue`)

`mergeTooltipRef()` already exists for this (`SideBar.vue:1566`) and is applied only inside `SideBar.vue`. Worth promoting to a shared helper — `v-bind` plus a template ref is now the house pattern in eight places.

**What held up**, for the record:

- All ~196 plain-`<button>` tooltip parents kept an `aria-label` or visible text; no accessible name was dropped.
- All five renamed e2e selectors resolve against the real markup.
- `tooltip=` only ever lands on `AppButton`/`AppBarButton`, both of which declare it.
- `designDrift.test.js` passes against the branch's own sources; the z-index debt list matches.
- The removed `chip.title` in `DedupPictureStrip.vue` is correct — `.gsmart` is `pointer-events: none`, so that `title` never rendered.

One thing to know rather than fix: `activator="parent"` is never used inside a `v-btn`, so `Tooltip.vue`'s hand-rolled `parentElement` walk dodges the `data-no-activator` trap by luck of `AppButton` being a plain `<button>`. The day someone puts one inside a real `v-btn`, Vuetify anchors to the `<button>` and this component anchors to `.v-btn__content`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

frontend/src/components/widgets/Tooltip.vue
```diff
@@ -0,0 +1,252 @@
...
+  el.addEventListener("mouseleave", () => clearTimeout(pendingOpen), {
+    once: true,
+  });
+}
+
+onMounted(() => {
+  if (!deferred) return;
+  parentEl.value = instance?.proxy?.$el?.parentElement ?? null;
+  parentEl.value?.addEventListener("mouseenter", arm);
+  parentEl.value?.addEventListener("focus", arm);
```
lindkvis: **MEDIUM — arming listens for `focus`, which does not bubble.**

For the ~89 `activator="parent"` sites whose DOM parent is a `div`/`span`/`label`/`li`/`p`/`dt` wrapper rather than the focused element itself, tabbing to the control inside never runs `arm()`, so the tip is never built and keyboard users get nothing. Vuetify uses `onFocusin` for this (`useActivator.mjs`) — `focusin` is the correct event. The ~196 plain-`<button>` parents are fine, which is part of why this isn't red; `Tooltip.test.js` covers no focus path at all.

Same swap needed on the `removeEventListener` at line 166.

frontend/src/components/widgets/Tooltip.vue
```diff
@@ -0,0 +1,252 @@
...
+});
+
+const attrs = useAttrs();
+const slots = useSlots();
+const instance = getCurrentInstance();
+const open = ref(false);
+
+const { openDelay, closeDelay } = tooltipDelays();
+
+const description = computed(() =>
```
lindkvis, line 123: **LOW/MEDIUM — `aria-description` is ARIA 1.3 and is not exposed outside Chromium.**

The rationale holds where the tip repeats information that has another home. It doesn't where the tip *is* the only home: `read-only-features.spec.js` now asserts `aria-description` carries "Duplicate review is only available in your own library" and "The model shelf is only available in your own library". Those were `title`, which Firefox and Safari do expose as the accessible description; `aria-description` they do not. A read-only guest on a non-Chromium browser loses the reason entirely.

frontend/src/components/panels/ProjectFiles.vue
```diff
@@ -60,14 +60,16 @@
           :key="file.id"
           class="pf-file-card"
           :class="{ 'pf-url-card': file.url }"
-          :title="file.url || file.original_filename"
+          :title="file.url ? undefined : file.original_filename"
           @click="openFile(file)"
         >
+          <Tooltip :text="file.url || ''" activator="parent" />
```
lindkvis, line 66: **MEDIUM — nested `activator="parent"` tips both open; the native `title` this replaces showed only the innermost.**

Hovering `.pf-file-delete` (line 72) shows "Remove" *and* this card's URL tip simultaneously. 22 such nestings exist on the branch — `SideBar.vue:6174/6187/6199/…`, `OverlayTagsPanel.vue:139/212`, `PictureSetEditor.vue:113/144/176`, `OverlayFilmstrip.vue:63/75`, `FolderTreeNode.vue:104`, `AddToEntityControl.vue:136`. The sidebar row plus its per-row action buttons is the worst of them.

frontend/src/components/widgets/StarRatingOverlay.vue
```diff
@@ -7,47 +7,64 @@
...
+      </Tooltip>
       <span
         class="star-number-label"
         :style="{ opacity: dScore > 0 ? 1 : 0 }"
         >{{ dScore > 0 ? dScore : 1 }}</span
       >
     </template>
     <template v-else>
-      <v-icon
+      <Tooltip
```
lindkvis: **MEDIUM — the deferral doesn't apply to the exact case the PR cites for it.**

The lazy-arming machinery is justified with "a grid tile carries several (the rating stars alone are five)", but these stars use the **slot** form, so `deferred = attrs.activator === "parent"` is false, `armed` starts `true`, and all five `VTooltip`s mount eagerly. `ImageGrid.vue:968` renders this per tile with `numberMode` defaulting to `false`, i.e. the 5-star loop. The claimed "~2x a `title`" doesn't hold here; it's still ~20x x 5 per tile.

frontend/src/components/panels/SelectionMenu.vue
lindkvis: Fixed in 4bb14f47. There's now a shared `utils/withRef.js` (`v-bind="withRef(props, (el) => (x = el))"`), which hands the element to Vuetify's activator ref and to the app's. It replaces `mergeTooltipRef` in `SideBar.vue` and is applied at all nine sites: the six tooltip ones you listed, plus three older `v-menu` activators with the same collision (the `SelectionBar` tag panel, the `ModelShelf` Add menu, the `LibrariesSection` row menu). `withRef.test.js` mounts the real `v-tooltip` and asserts both refs are filled; removing either half turns it red. `designDrift.test.js` now fails on any element that binds `*Props` beside a `ref`.

frontend/src/components/panels/SelectionMenu.vue
lindkvis: Fixed in 4bb14f47, same change as line 23: `withRef(tipProps, (el) => (ateCharacterRef = el))`.

frontend/src/components/panels/SelectionMenu.vue
lindkvis: Fixed in 4bb14f47, same change as line 23: `withRef(tipProps, (el) => (ateSetRef = el))`.

frontend/src/components/views/ImageOverlay.vue
lindkvis: Fixed in 4bb14f47: `withRef(setTipProps, (el) => (addToSetControlRef = el))`, so Vuetify anchors the tip and the app keeps its ref.

frontend/src/components/reviews/ReviewBinaryCard.vue
lindkvis: Fixed in 4bb14f47: `withRef(tipProps, (el) => (imgRef = el))`. `imgRef` is filled again, so `measure()` and the ResizeObserver run.

frontend/src/components/editors/PictureSetEditor.vue
lindkvis: Fixed in 4bb14f47: `withRef(tipProps, (el) => (nameInputRef = el))`, so the dialog focuses and selects its name field again.

frontend/src/components/widgets/Tooltip.vue
lindkvis: Fixed in b4700e9b. The parent form now listens for `focusin`/`focusout` on the parent. `focusin` arms the tip and opens it on a `:focus-visible` target inside, and `focusout` closes it when focus leaves the parent. A new test in `Tooltip.test.js` focuses a button inside a wrapper parent and asserts the tip opens and closes; it goes red with `focus` in place of `focusin`.

frontend/src/components/widgets/Tooltip.vue
lindkvis: Fixed in b4700e9b, centrally rather than for the two rows. `Tooltip` no longer uses `aria-description`. Each describing tip keeps a hidden text node and adds its id to the control's `aria-describedby`, next to the control's own ids, so every engine exposes it. A per-control MutationObserver re-adds the id when Vue or Vuetify rewrites the attribute. The unit tests resolve the ids to text (`testing/describedAs.js`), and `read-only-features.spec.js` uses `toHaveAccessibleDescription`, which goes red with the description turned off. Cost: a describing tip mounts in about 5x a `title` (jsdom, 1,000 buttons), 3.5x when the tip is the name.

frontend/src/components/panels/ProjectFiles.vue
lindkvis: Fixed in b4700e9b, centrally for all 22 nestings. Open tips are tracked, and of nested ones only the innermost stays open: an inner tip opening closes an open outer one, and an outer tip won't open while an inner one inside its control is open. That matches the native `title` behaviour this replaces. `Tooltip.test.js` covers a card tip with a delete button inside, and goes red with the outer-close removed.

frontend/src/components/widgets/StarRatingOverlay.vue
lindkvis: Right, fixed in b4700e9b. Each star is wrapped in an `inline-flex` span (not `display: contents`, which would give the tip no rect to anchor to), and its Tooltip uses `activator="parent"`, so the five per tile defer like the rest. The architecture doc now says a control repeated per tile takes the parent form, because the slot form can't defer.

PR #1346, opened 2026-09-14
Review leftovers from #1337: field and choice-control defects

lindkvis:
Five findings from a review of #1337, now merged. Each is a defect that PR
introduced, not a redesign of it.

- **An empty error message marked the field.** `AppInput`'s `error` prop listed
  `Boolean` before `String`, so Vue's boolean cast turned `""` into `true`: a
  caller that keeps its message in one ref (ShareDialog's expiry) rendered the
  red border and `aria-invalid` in the valid state, and clearing the message
  restored exactly the value that set it. Types unchanged, order swapped.
- **Two local `.tbm-select` overrides still pinned `min-height: 34px`**, which
  beats the base's new `height: var(--control-h)`. The filter panel's scope
  selects and the import panel's project select stood 6px above the inputs in
  the same rows, which is the defect #1337 set out to remove.
- **The zoom bar's Fit / Actual pixels did nothing inside the snap slack.**
  Both stops match within 1%, so the group already reported `fit` at 99% of
  fit and `Segmented` suppressed the emit for the current value. It now emits
  `pick` on every click, the same event and wording `OptionRows` already has;
  arrows keep going through the model.
- **The two always-dark surfaces take the dark theme's olive for the selected
  fill.** Both override blocks already restated ink, hover, trough, ring and
  focus, but left `--v-theme-primary` at the light theme's deep olive, which on
  a near-black trough separates on neither hue nor elevation. They now also map
  `primary`/`on-primary` to `dark-surface-primary` over the surface's own ink,
  the pairing main.js measures at 5.50:1 and 6.25:1.
- **A comment where the shared option mapper feeds both shapes**, since its
  `icon` renders in `OptionRows` and not in `Segmented`'s label variant.

Two new assertions, each shown red with the fix removed: the empty-string error
in `AppInput.test.js`, and `pick` on a click of the current value in
`Segmented.test.js`. Full vitest run green (4094 tests); eslint clean. No
changelog fragment: #1337 has not shipped, so none of this reached a release.


lindkvis (commented):
QA review completed for head 22ba4402a2e3bcf089bfa2818473fa25088f5cd7. No actionable regressions found in the 10-file patch. Inspected the AppInput empty-string prop casting fix, Segmented current-choice pick handling and zoom re-snap wiring, local control-height overrides, and dark-surface token overrides. git diff --check passes. Focused vitest and eslint could not run in the isolated checkout because frontend dependencies are absent (vitest/eslint not found).

PR #1347, opened 2026-09-14
Menu keyboard and ARIA leftovers from #1338

lindkvis:
Review leftovers from #1338. The CSS/markup consolidation was clean; what it
quietly changed was ARIA and keyboard behaviour.

### Menus answer the arrow keys again

Replacing `VList` with a plain `.ctx-menu` div dropped the roving focus `VList`
implements. What was *not* lost, and so is not re-implemented: a `v-menu`'s
activator already answers ArrowDown/ArrowUp by focusing the content's first
focusable child. The gap was movement **between** rows once focus is inside.

`utils/menuKeyboard.js` is one handler bound on the surface, so no row needs
wiring. Applied to every `.ctx-menu` that declares `role="menu"` and had no
keydown of its own - 12 surfaces across 10 components, not only the three the
review named. `ImageGridContextMenu` and the collapsed project menu keep their
existing handlers.

The sidebar's move menus also gained `role="menu"` / `menuitemcheckbox` on plain
`<div>`s with no tabindex and no key handling, which puts a screen reader into
menu mode on a widget that answers nothing - worse than the browse-mode divs it
replaced. Those rows are `<button>`s now (focusability and Enter/Space are the
browser's job), and because they are teleported divs rather than a `v-menu`,
nothing was moving focus into them at all: the opener now does.

### The rest

- A move menu's group heading lost its current-project emphasis; every heading
  read alike. Restored as `.ctx-label--current` on the shared surface, taking
  `--selected-ink` - the same olive the rows mark a current value with. The
  handoff doc claimed this had been "folded into `.ctx-label`"; it had not, and
  now says what actually happened.
- `scenes.js` asserted on a bare `getByRole('menu')`, which is strict-mode
  ambiguous now that a dozen more surfaces carry that role. Named menuitem.
- `GridPage.contextMenuItem` excluded flyout list rows but not each flyout's own
  trigger, which is `.ate-btn.ctx-item` and renders the assigned entity's name -
  the exact collision the comment said it prevented.
- `.ate--force-dark .ate-btn` still painted a filled pill unconditionally.
  Unreachable today, wrong the moment something pairs force-dark with
  `placement="right"`.

### Checks

`menuKeyboard.test.js` covers wrap, Home/End, both disabled spellings, submenu
isolation, which keys are claimed, and the empty-menu guard. Each assertion was
shown failing against a mutated helper and restored. Full vitest suite 4099/4099,
eslint clean, build clean. No changelog fragment: #1338 has not been released.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


frontend/src/components/panels/SideBar.vue
```diff
@@ -461,12 +462,22 @@ function _moveMenuPos(rect) {
   return { top: rect.bottom + 4 + "px", left };
 }
 
+// These menus are teleported divs, not a `v-menu`, so nothing moves focus into
+// them the way VMenu's activator does. Without this the menu declares
+// `role="menu"` and then answers no key at all - worse for a screen reader than
+// the unrolled divs it used to be, which at least read in browse mode.
+async function _focusMoveMenu() {
+  await nextTick();
+  document.querySelector(".sidebar-move-menu")?.focus();
```
lindkvis, line 474: **[P2] Focus the menu that was just opened**

Opening the People menu for a project and then activating its Sets `+` leaves both menus rendered. The document `mousedown` outside handler only closes these flags when the target is outside *any* `.sidebar-move-to-project-wrap` and `.sidebar-move-menu`; the Sets trigger is inside the wrapper, so it preserves the People flag before its `click` toggles the Sets flag. Both `v-if` conditions then pass for the selected project. After `nextTick`, `document.querySelector(".sidebar-move-menu")` returns the People menu because its teleport occurs first in DOM order, so focus remains there and Arrow/Home/End act on the wrong menu.

This is static control-flow evidence, not a browser runtime reproduction. Pass the intended menu element/type into `_focusMoveMenu` (or close the other menu before opening this one), rather than selecting the first matching surface.

lindkvis: Fixed in fca73b2b by closing the other menu before opening this one. Two move menus open together was the root problem, and with only one `.sidebar-move-menu` rendered, `_focusMoveMenu`'s query can't pick the wrong surface. Test: `SideBarProjectScoping.test.js` › "a project's move menus". It fails if either close line is removed.

Issue #1307, opened 2026-09-13
Workflows: run a workflow


lindkvis:
Run a workflow from PixlStash.

- From the selection pill for image-to-image, from the toolbar for text-to-image.
- The run panel sits in the inspector rail, so browsing is not blocked.
- Fill each picture input by its mode, submit, and collect output from the detected save node.
- State the run count before starting. Record the resolution lock as today.

Depends on: #1305, #1306

Issue #1306, opened 2026-09-13
Workflows: parameter form


lindkvis:
Turn a workflow's settings into a form.

- Controls typed from ComfyUI's `object_info`, with real ranges.
- Models are picked from the model shelf. Seed gets a randomize option.
- Connected and bound inputs are not controls.
- Show the few parameters that matter; the rest go behind "All N parameters". Pins are saved per workflow.
- When ComfyUI is unreachable, show the recorded values without ranges.

Depends on: #1302

Issue #1305, opened 2026-09-13
Workflows: set how each picture input is filled


lindkvis:
Per picture input, one of:

- **Selection**: the grid selection fills it. At most one per workflow; it sets the run count.
- **Picker**: asked when the workflow runs.
- **Fixed**: chosen here, read-only at run time.

One input defaults to Selection. Show the node id beside each input's name. A workflow with no Selection input is offered only from the toolbar, not the selection pill. Uses the existing picture picker.

Depends on: #1302

Issue #1303, opened 2026-09-13
Workflows: import a workflow as-is


lindkvis:
Import stores the file unchanged, in UI or API format.

- Drag and drop a workflow JSON onto the Workflows view.
- Remove placeholder rewriting (`applyWorkflowPlaceholders`, `_replace_placeholders`, `_find_placeholder_usage`).
- Migrate workflows already stored with placeholder tokens: record the node the token sat on as a binding and restore a neutral value. Flag a workflow whose original value can't be recovered.
- A dropped copy of a known workflow matches the existing one.

Depends on: #1302

PR #1350, opened 2026-09-14
Workflows: detect a workflow's inputs and outputs

### LINKED ISSUE #1302: Workflows: detect a workflow's inputs and outputs
Find what a workflow needs without any configuration:

- the save node,
- picture inputs,
- positive and negative prompts, by following the guider's own `positive` / `negative` inputs.

Report what it finds; never write it. Report ambiguity (two image loaders, two samplers) rather than guessing. Works through subgraphs.

This replaces placeholder detection as the source of `valid` and `workflow_type` in `GET /comfyui/workflows`: a save node makes a workflow valid, a picture input makes it `i2i`. The response shape does not change.

Starts after the design cleanup: #1294 #1295 #1296 #1297 #1298 #1299 #1300 #1301

lindkvis:
Closes #1302.

Adds `services/workflow_io.py`. It finds a workflow's save nodes, picture inputs and positive/negative prompts from the graph, with no configuration. Prompts come from following each guider's own `positive`/`negative` inputs upstream to the text encoder; a `ConditioningZeroOut` on the way means there is no prompt. It runs on `workflow_hash`'s reductions, so UI and API format both work and subgraphs are inlined. It only reports and never writes. Two save nodes, two picture inputs, or samplers reading different prompts go into `ambiguities`, and when samplers disagree the prompts are left empty rather than guessed.

`GET /comfyui/workflows` now takes `valid` (has a save node) and `workflow_type` (has a picture input → `i2i`) from detection. The response shape is unchanged, and `missing_placeholders` is still reported.

Because the list and the run routes can now disagree, `run_i2i` refuses any workflow without `{{image_path}}`. Otherwise a caption-only workflow with a fixed reference `LoadImage` would list as i2i, run, and have its output stacked onto a picture it never read.

Known limits, left for later cards:
- Picture inputs are matched by class name (`*LoadImage*`). A loader named otherwise isn't found until `object_info` typing (#1306).
- A mask loader counts as a picture input. #1305 decides how each input is filled.
- `DualCFGGuider` `cond1`/`cond2` aren't read, and prompts wired through a bypassed node aren't found. The UI reduction drops bypassed nodes rather than passing their edges through.
- SDXL base + refiner is reported as an ambiguity. The two samplers really do read different encoders, so that is the intended report, not a bug.
- A workflow whose only output is a `PreviewImage` named in `pixlstash_output_nodes` now lists as invalid, and UI-format files list as valid even though the run routes can't submit them yet. Both are resolved once import stores files as-is (#1303) and runs use the detected save node (#1307).


pixlstash/routes/comfyui.py
```diff
@@ -676,23 +677,29 @@ async def list_comfyui_workflows():
...
+                valid = False
+                missing = [PLACEHOLDER_IMAGE, PLACEHOLDER_CAPTION]
+                workflow_type = "t2i"
                 try:
                     payload = _load_workflow_json(path)
-                    valid, missing = _find_placeholder_usage(payload)
+                    missing = _find_placeholder_usage(payload)[1]
+                    detected = detect_workflow_io(payload)
+                    valid = detected.valid
+                    workflow_type = detected.workflow_type
```
lindkvis: **Regression: a caption-only workflow with a fixed `LoadImage` becomes unusable.** Before this PR it listed as a valid `t2i`, `TbComfyPanel` offered it (it filters on `workflow_type === "t2i"`), and `run_t2i` ran it. Now it lists as `i2i`, so the t2i panel hides it. The i2i menus (SelectionBar, ImageOverlay, RemixDialog) offer it instead, and `run_i2i` refuses it with a 400 because `{{image_path}}` is missing. So it can't be run from anywhere. The same happens to any workflow that has a `LoadImage` and no placeholder at all (dropped straight into the user folder, or imported without an image target). Those used to run through t2i. `test_run_i2i_refuses_a_workflow_without_the_image_placeholder` builds exactly this graph and only checks the refusal.

pixlstash/services/workflow_io.py
```diff
@@ -0,0 +1,184 @@
...
+        nodes = reduce_api_graph(document)
+
+    # The classes output collection imports from, so `valid` means runnable.
+    save_nodes = sorted(
+        key for key, node in nodes.items() if node.class_type in SAVE_NODE_CLASSES
+    )
+    # ponytail: name rule (LoadImage, LoadImageMask, LoadImageFromPath, ...);
+    # a loader named otherwise is not found until object_info types it.
+    picture_inputs = sorted(
+        key for key, node in nodes.items() if "LoadImage" in node.class_type
```
lindkvis: **The reverse case: `{{image_path}}` on a loader this name rule misses.** Examples: the pack's own `PixlStashPictureLoader`, `easy loadImageBase64` (the match is case-sensitive), WAS `Image Load`. Such a workflow now lists as `t2i`, so the i2i menus drop it. `run_t2i` then refuses it because the dump contains `{{image_path}}`. It used to list and run as i2i. Also, `SAVE_NODE_CLASSES` includes the PixlStash saver but this rule leaves out the PixlStash picture loader.

pixlstash/services/workflow_io.py
```diff
@@ -0,0 +1,184 @@
...
+    """
+    if isinstance(document, dict) and isinstance(document.get("nodes"), list):
+        nodes = reduce_ui_graph(document)
+    elif isinstance(document, dict) and isinstance(document.get("prompt"), dict):
+        # The import dialog stores an embedded prompt chunk as this envelope.
+        nodes = reduce_api_graph(document["prompt"])
+    else:
+        nodes = reduce_api_graph(document)
+
+    # The classes output collection imports from, so `valid` means runnable.
```
lindkvis: `valid` doesn't mean runnable, and this set isn't everything output collection imports from. `_extract_output_node_ids` takes the request's `output_node_ids` first, then the document's `pixlstash_output_nodes`, whatever their class. Detection ignores both. A workflow whose outputs were chosen explicitly (a `PreviewImage`, a custom saver) now lists as invalid, and RemixDialog and ImageOverlay filter on `valid`, so they stop offering it. The PR body mentions this. Still, it is a visible regression until #1303/#1307 land, and this comment claims otherwise.

pixlstash/services/workflow_io.py
```diff
@@ -0,0 +1,184 @@
...
+        if "TextEncode" in node.class_type:
+            found.add(key)
+            continue
+        if node.class_type == "ConditioningZeroOut":
+            continue
+        # A node emitting both sides (ControlNetApplyAdvanced) orders them
+        # positive, negative: follow the input matching the output we came from.
+        by_name = {name: (source, s) for name, source, s in node.inputs}
+        if "positive" in by_name and "negative" in by_name:
+            pending.append(by_name["positive" if slot == 0 else "negative"])
```
lindkvis, line 215: This assumes a node with `positive`+`negative` inputs outputs positive on slot 0 and negative on slot 1. It doesn't always hold. Example: chained `KSampler (Efficient)` nodes, where CONDITIONING+ is slot 1 and CONDITIONING- is slot 2. There, slot 1 maps to `negative`, and the result is `positive_prompts=['<negative encoder>']` with no ambiguity reported. I reproduced this. It is a wrong answer given with confidence, not a report.

pixlstash/services/workflow_io.py
```diff
@@ -0,0 +1,184 @@
...
+            continue
+        pending.extend(
+            (source, s) for name, source, s in node.inputs if "conditioning" in name
```
lindkvis: A passthrough node with only a `positive` input stops the walk. Example: core `HunyuanImageToVideo(positive=...)`. `by_name` lacks `negative`, and `"conditioning" in "positive"` is False, so nothing gets pushed. Reproduced: the sampler gets `positive_prompts=[]` and no ambiguity.

pixlstash/services/workflow_io.py
```diff
@@ -0,0 +1,184 @@
...
+        )
+    return tuple(sorted(found))
```
lindkvis, line 225: A dead end returns the same empty tuple as `ConditioningZeroOut`. A dead end is conditioning from a node that is neither a `*TextEncode*` nor has conditioning inputs, e.g. `Efficient Loader` or `ImpactWildcardEncode`. So "we couldn't find the prompt" reads as "this model has no prompt", with nothing in `ambiguities`. The module docstring says it reports rather than guesses, so a card that later fills prompts would trust an empty list that is really unknown.

pixlstash/routes/comfyui.py
```diff
@@ -786,8 +793,11 @@ async def run_comfyui_i2i(request: Request, payload: dict = Body(...)):
             raise HTTPException(status_code=404, detail="Workflow not found")
 
         workflow_payload = _load_workflow_json(workflow_path)
-        valid, missing = _find_placeholder_usage(workflow_payload)
-        if not valid and PLACEHOLDER_IMAGE in missing:
+        _valid, missing = _find_placeholder_usage(workflow_payload)
+        # Refused even when the caption placeholder makes the workflow a valid
+        # t2i: without the image placeholder the selected picture never reaches
+        # the graph, and its output would be stacked onto a picture it ignored.
+        if PLACEHOLDER_IMAGE in missing:
```
lindkvis, line 824: Altitude: the listing now uses one classifier (detection) and the run routes another (placeholders). This guard fixes one direction of the mismatch. Two stay open: the t2i side (above) and the frontend copy (see review body). Until #1303/#1307 make runs use detection, it may be simpler to keep `workflow_type` agreeing with what the run routes accept. For example, `i2i` = picture input *and* `{{image_path}}`. Detection can still be exposed as extra fields.

pixlstash/routes/comfyui.py
```diff
@@ -676,23 +677,29 @@ async def list_comfyui_workflows():
...
                 seen.add(entry)
                 path = os.path.join(folder, entry)
+                valid = False
+                missing = [PLACEHOLDER_IMAGE, PLACEHOLDER_CAPTION]
+                workflow_type = "t2i"
                 try:
                     payload = _load_workflow_json(path)
-                    valid, missing = _find_placeholder_usage(payload)
+                    missing = _find_placeholder_usage(payload)[1]
+                    detected = detect_workflow_io(payload)
```
lindkvis: Efficiency: every `GET /comfyui/workflows` now does a full graph reduction of every workflow file, including UI subgraph inlining (capped at 50k nodes). It runs synchronously inside an `async def`, so it blocks the event loop. The frontend calls this each time a ComfyUI menu or dialog opens. A cache keyed on `(path, mtime)`, or detecting once at import, would avoid redoing it.

pixlstash/services/workflow_io.py
```diff
@@ -0,0 +1,184 @@
...
+``ControlNetApplyAdvanced``, ...), to the text encoder that feeds them. A path
+through ``ConditioningZeroOut`` carries no prompt: that is how a model without
+a negative prompt is wired.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+
+from pixlstash.services.comfyui_service import SAVE_NODE_CLASSES
```
lindkvis, line 27: Reuse/consistency: there are already two rule sets here that this doesn't share. `utils/comfyui_utilities.py` walks positive prompts with `_follow_positive_api` / `_follow_positive_ui` and an exact `_CLIP_TEXT_ENCODE_CLASSES`. `comfyui_recipe_service` has `INPUT_IMAGE_FIELDS` and `SAVE_IMAGE_CLASSES`, and the latter *includes* `SaveImageWebsocket`. So recipe preflight says a websocket graph writes an image, while this listing (and `test_websocket_save_is_not_collected_so_not_valid`) calls it invalid. That makes three definitions of save node, picture input and prompt.

pixlstash/routes/comfyui.py
```diff
@@ -786,8 +793,11 @@ async def run_comfyui_i2i(request: Request, payload: dict = Body(...)):
             raise HTTPException(status_code=404, detail="Workflow not found")
 
         workflow_payload = _load_workflow_json(workflow_path)
-        valid, missing = _find_placeholder_usage(workflow_payload)
-        if not valid and PLACEHOLDER_IMAGE in missing:
+        _valid, missing = _find_placeholder_usage(workflow_payload)
```
lindkvis: `_find_placeholder_usage`'s `valid` result and its t2i/i2i branch are now unused by both callers (`[1]` here, `_valid` above). The helper could just return `missing`.

pixlstash/routes/comfyui.py
```diff
@@ -676,23 +677,29 @@ async def list_comfyui_workflows():
...
-                    valid, missing = _find_placeholder_usage(payload)
+                    missing = _find_placeholder_usage(payload)[1]
+                    detected = detect_workflow_io(payload)
+                    valid = detected.valid
+                    workflow_type = detected.workflow_type
                 except Exception as exc:
-                    logger.warning("Failed to read workflow %s: %s", entry, exc)
-                    valid = False
-                    missing = [PLACEHOLDER_IMAGE, PLACEHOLDER_CAPTION]
+                    logger.warning(
```
lindkvis, line 714: This `except Exception` now also catches every detection failure (malformed subgraph, missing definition), and it logs at warning on every list call. That is log spam for a file that stays broken. It also logs only `entry`, not `source`/`path`. CLAUDE.md asks to log exceptions "with as much context as possible (e.g., variable values, file paths, operation being performed)".

lindkvis (commented):
Review notes (reproduced by running `detect_workflow_io` on small graphs).

Two siblings outside the diff that the new classification affects:
- `frontend/src/components/views/ImageOverlay.vue` still says "No valid workflows found. Workflows need a {{image_path}} placeholder" and "N workflow(s) missing required placeholders". `invalidComfyWorkflows` (`!valid && i2i`) couldn't match anything before, because every i2i workflow was valid. It now matches workflows that *have* their placeholders but no `SaveImage`, so the note gives the wrong reason.
- `frontend/src/components/panels/TbComfyPanel.vue` filters on `workflow_type === "t2i"`. That is how the fixed-`LoadImage` caption workflows disappear from the t2i panel (see the inline comment on the list route).

The most important finding is the first inline comment: the list and run routes disagree in both directions, and a workflow that used to run can no longer be run from any menu.

tests/test_workflow_io.py
```diff
@@ -0,0 +1,241 @@
...
+        if getattr(route, "path", None) == "/comfyui/workflows"
+        and "GET" in route.methods
+    )
+    listed = {
+        item["name"]: (item["valid"], item["workflow_type"])
+        for item in asyncio.run(endpoint())["workflows"]
+    }
+    missing = {
+        item["name"]: item["missing_placeholders"]
+        for item in asyncio.run(endpoint())["workflows"]
```
lindkvis: Nit: this runs the endpoint a second time just to read `missing_placeholders`. One call can fill both dicts.

pixlstash/services/workflow_io.py
```diff
@@ -0,0 +1,184 @@
...
+from pixlstash.services.workflow_hash import (
+    ReducedNode,
+    reduce_api_graph,
+    reduce_ui_graph,
+)
+
+_PROMPT_SIDES = ("positive", "negative")
+
+
+@dataclass(frozen=True)
```
lindkvis, line 43: Nit: `frozen=True` with `list` fields gives shallow immutability, and `hash(WorkflowIO())` raises `TypeError`. Using tuples would make it actually frozen and hashable.

pixlstash/routes/comfyui.py
lindkvis: Fixed in afca4771. The list still classifies by detection, as #1302 asks. Until #1307, the four menus (TbComfyPanel, SelectionBar, ImageOverlay, RemixDialog) now choose by `missing_placeholders`, which is what the run routes accept, instead of `workflow_type`. The caption-only workflow with a fixed `LoadImage` is back in the t2i panel and out of the i2i menus.

pixlstash/services/workflow_io.py
lindkvis: Fixed in afca4771. The name match is now case-insensitive and ignores spaces, and a known-class list adds `INPUT_IMAGE_FIELDS`' loaders, `PixlStashPictureLoader` and WAS `Image Load`. The menus also no longer depend on `workflow_type` (see the thread on the list route), so a loader still missing from the rule can't hide a runnable workflow. Test: `test_loaders_the_name_rule_used_to_miss`.

pixlstash/services/workflow_io.py
lindkvis: Fixed in afca4771. Save nodes now follow `_extract_output_node_ids`: an explicit `pixlstash_output_nodes` / `pixlstash_output_node` choice wins whatever its class, otherwise `SAVE_NODE_CLASSES`. I corrected the comment. Test: `test_an_explicit_output_choice_is_the_save_node`. A per-request `output_node_ids` can't count, because the list has no request.

pixlstash/services/workflow_io.py
lindkvis: Fixed in afca4771. Slots 0/1 are only trusted for pure conditioning nodes. A two-sided node that also takes `model` (a pack sampler), or a slot above 1, now makes that side "not found" and records it in `ambiguities`, instead of guessing. Test: `test_a_pack_sampler_emitting_conditioning_is_not_read_by_slot`.

pixlstash/services/workflow_io.py
lindkvis: Fixed in afca4771. A passthrough now follows any input named `positive`/`negative` or containing `conditioning`. Test: `test_a_positive_only_passthrough_is_followed`.

pixlstash/services/workflow_io.py
lindkvis: Fixed in afca4771. `_prompts_for` returns `None` when a path ends at anything other than an encoder or `ConditioningZeroOut` (including a missing node). That side is left empty and `"<side> prompt not found"` is added to `ambiguities`. Test: `test_a_dead_end_is_not_found_rather_than_no_prompt`.

pixlstash/routes/comfyui.py
lindkvis: Addressed in afca4771, slightly differently. #1302 explicitly makes detection the source of `valid`/`workflow_type`, so the list stays on detection. The menus instead choose by `missing_placeholders` until #1307, which closes both directions in one place per menu. The overlay copy is updated too. Documented in backend_architecture.md §5 `comfyui.py`.

pixlstash/routes/comfyui.py
lindkvis: Fixed in afca4771. Each file is described by `_describe_workflow`, an `lru_cache` keyed on `(path, source, mtime_ns, size)`, so detection runs once per file version. The handler is now a plain `def`, so FastAPI runs it in the threadpool rather than on the event loop.

pixlstash/services/workflow_io.py
lindkvis: Partly addressed in afca4771: the picture-input rule now builds on `comfyui_recipe_service.INPUT_IMAGE_FIELDS`. The save rule already is output collection's own `SAVE_NODE_CLASSES`, which `valid` has to match. Preflight's `SaveImageWebsocket` answers a different question (does anything write an image at all) and history-based collection doesn't import websocket output, so I've left the two sets apart. `comfyui_utilities`' walk extracts prompt text for display and stops at the first match, which isn't reusable for the side-aware walk here. Leaving this open in case you want them unified anyway.

pixlstash/routes/comfyui.py
lindkvis: Fixed in afca4771. It's now `_missing_placeholders(payload) -> list[str]`, and both callers use it.

pixlstash/routes/comfyui.py
lindkvis: Fixed in afca4771. Reading and detection are caught separately inside the cached `_describe_workflow`. Each logs the source and full path, and only once per file version, since a failure is cached with the file's mtime. The test checks that a second listing doesn't log the broken file again, and that rewriting it relists it.

tests/test_workflow_io.py
lindkvis: Fixed in afca4771. One call now fills both dicts.

pixlstash/services/workflow_io.py
lindkvis: Fixed in afca4771. The fields are tuples, so `WorkflowIO` is genuinely frozen and hashable.

PR #1351, opened 2026-09-14
Workflow ghosts: Privacy purges, model-ghost forget, Ghosts filter

### LINKED ISSUE #1309: Workflows: what is kept when pictures are deleted
When pictures are purged, their workflows can outlive them as ghosts: a thumbnail with its prompt, or the filename of a model no longer on the shelf.

- Purge reaches the hub, inside the existing scrapheap purge. No second deletion path.
- A privacy setting with three positions: Off, Covered only (default), On. Covered only keeps a ghost while a surviving picture shares its instance.
- Purging the last covering picture removes the ghosts it covered.
- Forgetting a prompt also clears its text embedding.
- A forgotten model keeps its hash, so the workflow still groups: "3 models, names forgotten".
- Two counted purges in Settings › Privacy, one for picture ghosts, one for model ghosts.
- A ghosts filter in the Workflows view.

lindkvis:
Closes #1309.

Most of the backend was already on develop (#1089, #1326): purge reaches the hub, the Off / Covered only / On setting, the covered-ghost cascade, and the picture-ghost erase. This adds the parts that were still missing.

**Backend**
- `GET /server-config/ghost-retention` now also returns `picture_ghosts` (active library) and `model_ghosts` (hub-wide). Both are `null` when there is no hub.
- New `DELETE /server-config/ghost-retention/model-ghosts?expected=N`, owner-only, added to the registry and the coverage matrix (322 routes / 145 owner_only). It forgets model names without touching hashes or documents, so workflows still group together. It returns 409 and changes nothing if the count no longer matches the `N` that was confirmed.
- A model ghost is a `.safetensors` name that matches no shelf filename or copy, or a loader `*_sha256` value that matches no shelf digest. Tombstoned models still count as on the shelf.
- `/workflows` rows now carry `ghosts`, `model_ghosts` and `forgotten_models`, and variants carry `forgotten_models`. Forgotten names are counted from the stored documents: an asset reference with no asset row behind it is a forgotten name.

**Frontend**
- Settings › Privacy has a new "Deleted pictures" section: the three-position setting plus two purge buttons that show their counts. Each purge asks for confirmation in an `AppDialog`, and the counts are re-read afterwards.
- The Workflows view has a Ghosts toggle next to Group / Sort / Show. It is a toggle rather than another Show option because it combines with them. A row whose names were forgotten reads "3 models, names forgotten".

**Not changed:** "forgetting a prompt also clears its text embedding" was already true without new code. The prompt and its embedding are both columns on the picture row the purge deletes, and a ghost stores no embedding.

**Answering the pre-push review.** Fixed:
- The purge counted every `.ckpt`/`.pth`/`.gguf` as a ghost, because the shelf only scans `.safetensors`.
- Digests were left behind.
- The confirmed count did not limit the delete.
- A failed re-read hid the error.
- Row text mixed a union of names with a per-variant maximum.
- The variant expand scanned every document in the hub.

Not changed:
- **A model never added to the shelf counts as a ghost.** The hub keeps no record of what was on the shelf before, so "no longer on the shelf" cannot be told apart from "never on it". The copy says so plainly.
- **The purge DELETEs take no confirm token**, like the existing picture-ghost erase. `DELETE /pictures/scrapheap` requires one because of the localhost-CORS concern. That is worth a separate decision for both ghost routes rather than a one-off here.


pixlstash/hub/workflows.py
```diff
@@ -209,6 +213,138 @@ def forget_asset_names(hub: HubDatabase, normalized_filename: str) -> int:
...
+    )
+    shelf_digests = {
+        row[0].lower()
+        for row in fetchall("SELECT sha256 FROM model WHERE sha256 IS NOT NULL")
+    }
+    ghosts = set()
+    for widget, value in fetchall(
+        "SELECT DISTINCT widget_name, normalized_filename FROM workflow_recipe_asset"
+    ):
+        if SHA256_FIELD_RE.search(widget):
```
lindkvis, line 256: **P2 — Restrict digest ghosts to known model-loader inputs.** The documented deletion boundary is model-only, but `SHA256_FIELD_RE` accepts every input name ending in `_sha256`. Normal graph ingestion accepts arbitrary node classes and passes each non-secret string input through `structural_widget_value`, so a third-party node with an `image_sha256` input is persisted in `workflow_recipe_asset` just like the built-in loaders. If its content digest is absent from `model.sha256`, this loop counts it in the Model ghosts confirmation; confirming then deletes its readable asset row at lines 284–290. That loses the non-model asset's identity while presenting it as a model-name purge. I reproduced the classification and deletion with one `image_sha256` asset row and no matching model: `model_ghost_names` returned the digest and `forget_model_ghosts(expected=1)` removed the row. Please retain enough source/type information to classify only model-loader digests, or limit this purge to a known allowlist of model-loader widget inputs.

PR #1356, opened 2026-09-15
Workflows: watched workflows folder

### LINKED ISSUE #1304: Workflows: watched workflows folder
A `workflows/` folder in the app's data directory.

- A file dropped in is imported, the same as a drag and drop.
- Deleting a workflow writes it back to the folder, moves it to the system trash, then removes the row.
- The folder is an inbox: removing a file never deletes a workflow.
- Reconcile at start-up. Filenames are content-addressed.

Depends on: #1303

lindkvis:
Closes #1304.

A `workflows/` folder in the app data directory now works as an inbox for workflow files.

- **Import.** A `.json` put in the folder is stored the same way as a drop: through `_store_workflow`, which is the import route's body moved to module level, with `keep_both`. The file is then renamed `<stem>.<hash>.json`. The hash is the first 16 hex digits of the sha256 of the workflow as stored (tokens migrated, `pixlstash_*` keys ignored).
- **Start-up and watching.** Start-up reconciles the folder after the placeholder migration, then a watchdog observer reconciles again one second after changes. This runs only when `Server.DEFAULT_WATCH_WORKFLOW_INBOX` is set, and the suite turns it off.
- **Removing a file from the folder never deletes a workflow.**
- **Delete.** `DELETE /comfyui/workflows/{name}` writes the workflow back to the folder and moves every inbox file holding it to the system trash (`send2trash`), whether renamed yet or not. The stored file is removed only after that. If the trash refuses, the call returns a 500 and the workflow is kept. Restoring the file from the trash imports it again.
- **Locking.** Reconcile, import and delete all run under one lock.

Code is in `services/workflow_inbox.py`. The route change is mostly the move to module level.

**Pre-push review objections I did not change**
- *Every reconcile re-stores the whole inbox, and the inbox only grows.* That is true. Each file is compared against every stored workflow. One person's folder is small, and a stored copy is what makes a file restored from the trash import again. An index can come later if it ever gets slow.
- *Editing a file inside the inbox stores each saved version beside the last one.* That is intended for an inbox: the folder never changes or deletes a stored workflow. The docs say so.
- *Delete now fails where there is no usable trash (some headless or Docker homes).* The issue asks for the trash. Refusing keeps the workflow rather than losing it. A permanent-delete option like the model shelf has can follow if Docker users need it.
- *No test fails when the lock is removed.* The races it guards are timing windows. I did not add a sleep-based test for them.

The Workflows list does not refresh by itself after a watched import. It shows the workflow the next time the list is loaded.


pixlstash/services/workflow_inbox.py
```diff
@@ -0,0 +1,232 @@
...
+    workflow before releasing it.
+
+    Raises:
+        OSError, send2trash.TrashPermissionError: The file could not be written
+            or trashed. The stored workflow must then be kept.
+    """
+    digest = content_hash(workflow)
+    os.makedirs(folder, exist_ok=True)
+    written = f"{_split(name)[0]}.{digest}.json"
+    with open(os.path.join(folder, written), "w", encoding="utf-8") as handle:
```
lindkvis: [P2] Preserve an inbox edit that is still awaiting reconciliation

If `flow.<old-hash>.json` is edited in place and the original workflow is deleted before the one-second debounce finishes, this `open(..., "w")` targets the same path, overwrites the new contents with the old stored workflow, and then trashes it. I reproduced this against the current PR head. Please write the deletion copy to a collision-free path and determine matches from their actual contents so a pending edit remains available for the watcher to import.

lindkvis: Fixed in bf3d2727. The deletion copy is now opened with `"x"` and gets `<stem> (2).<hash>.json` when the name is taken, so it never overwrites an existing inbox file. Matches are decided by hashing each file's contents, never by the hash in its name, so an in-place edit to `flow.<old-hash>.json` stays in the inbox and the watcher imports it. New test `test_deleting_keeps_an_inbox_edit_the_watcher_has_not_imported` covers your scenario. It goes red if the write goes back to `"w"` and also if files are matched by name.

PR #1359, opened 2026-09-15
Build and run the desktop app from a worktree

lindkvis:
Adds `scripts/desktop-dev.sh build|run`, so a worktree with no venv, node_modules or build output can launch the Electron app in two commands (e.g. as a Piecework card's build/run).

- Finds a `.venv` in the worktree, the main checkout, or beside it (`PIXLSTASH_DEV_BACKEND` overrides).
- Puts the worktree on `PYTHONPATH`; otherwise an editable install of another checkout in a shared venv wins and the backend runs the wrong code.
- Runs Electron's `install-electron` when the binary is missing: 42.x no longer fetches it during `npm ci`.


scripts/desktop-dev.sh
```diff
@@ -0,0 +1,61 @@
...
+  if [[ ! -f "$ROOT/electron/node_modules/electron/path.txt" ]]; then
+    (cd "$ROOT/electron" && npx --no-install install-electron)
+  fi
+  (cd "$ROOT/frontend" && npm run build)
+  (cd "$ROOT/electron" && npm run check-electron && npm run build)
+}
+
+run() {
+  local python="${PIXLSTASH_DEV_BACKEND:-$(find_venv_python)}"
+  echo "desktop-dev: backend $python on $ROOT"
```
lindkvis, line 48: **Missing virtual environments do not stop execution.** The successful `local` builtin masks the non-zero status from `find_venv_python`, so `set -e` does not exit here. With no matching `.venv`, the script logs the error and continues with an empty `PIXLSTASH_DEV_BACKEND`, leading to a later Electron startup failure. Please split declaration from assignment so the failed substitution propagates:

```suggestion
  local python
  python="${PIXLSTASH_DEV_BACKEND:-$(find_venv_python)}"
```

lindkvis (commented):
One issue to address before merge.

scripts/desktop-dev.sh
lindkvis: Fixed in a04141f4 as suggested: `local python` on its own line, then the assignment. Checked both ways in a checkout with no `.venv`: before, it logged the error and carried on with an empty backend; now `run` exits 1 at the error. The normal run still resolves the shared venv.

PR #1362, opened 2026-09-15
Workflows: parameter form (backend)

lindkvis:
Part of #1306: the backend half. The form itself comes after #1307, so it can sit in the run panel. #1307 needs this part first.

**What it adds**
- `services/workflow_parameters.py` turns each settable value of a saved API-format workflow into a form control.
  - Types come from ComfyUI's `object_info`: numbers with their real min, max and step, combos with their options (numeric ones too), seeds (inputs flagged `control_after_generate`) and models.
  - Left out: connected inputs, inputs a run fills (every picture input and the bound prompt), credential names (`api_key`, `*_token`, `password`, `secret`), and values that are not scalars.
  - A `PrimitiveInt` counts as a seed only when it feeds a seed input. ComfyUI flags it re-rollable even when it drives a width.
  - `apply_values` checks values against that description and writes them into a copy of the graph, ready for the run in #1307.
- `GET /comfyui/workflows/{name}/parameters`
  - When ComfyUI is unreachable it still returns the file's values, with no ranges, plus `typed: false` and the error.
  - A UI-format file answers `readable: false`.
  - `pins` lists what shows before "All N parameters": the saved pins, or by default the models, seeds, steps, cfg, guidance, sampler, scheduler, denoise, width and height.
- `PUT /comfyui/workflows/{name}/pins` saves the pins in order, or restores the defaults with `null`. It doesn't ask ComfyUI.
  - Pins go in a new hub table `workflow_parameter_pins`, one row per file, never inside the file.
  - Deleting the workflow drops its pins.
- Both routes are owner-only: registry, coverage matrix (326) and the tests in both directions. The docs are updated.

**Pre-push review objections I did not change**
- *`object_info` is fetched uncached on every GET, and a slow ComfyUI can hold a worker for the 15 s timeout.* The recipe preflight already does the same fetch. The route is owner-only and the URL is the owner's setting. A cache can come if opening the form feels slow.
- *UI-format files get no parameters.* Their widget values are an unnamed list, and the run routes only submit API format.
- *A picture loader whose class name detection doesn't recognise shows its `image` field as a parameter.* Detection still works from class names. The comment in `workflow_io.py` now says so instead of pointing at this card.

No changelog fragment: nothing is visible until the form lands.


pixlstash/services/workflow_parameters.py
```diff
@@ -0,0 +1,382 @@
...
+        for target in targets:
+            path = target.get("path")
+            if isinstance(path, list) and len(path) >= 3 and path[-2] == "inputs":
+                bound.add((str(path[-3]), str(path[-1])))
+    # Every picture input, not only the one a run fills today: #1305 gives each
+    # one a mode, and none of them is a setting.
+    graph = api_graph(document) or {}
+    for node_id in detect_workflow_io(document).picture_inputs:
+        class_type = (graph.get(node_id) or {}).get("class_type")
+        for field in INPUT_IMAGE_FIELDS.get(class_type, ("image",)):
```
lindkvis: **Detected picture inputs leak into the parameter form.** `_bound_inputs` excludes only `INPUT_IMAGE_FIELDS.get(class_type, ("image",))`, so a loader whose picture field isn't `image` keeps it as an editable parameter.

Repro: `{"1": PixlStashPictureLoader {picture_ids: "12,13"}, "2": "Image Load" {image_path: "/x/a.png"}}`. `detect_workflow_io` reports both as picture inputs, but `describe_parameters` returns `("1","picture_ids")` and `("2","image_path")` as string parameters. The docs say every detected picture input is excluded, and with #1307 `apply_values` would write the field the input mode fills. Suggest one shared loader-class -> picture-field map (or carry the field on `WorkflowIO`).

pixlstash/services/workflow_parameters.py
```diff
@@ -0,0 +1,382 @@
...
+    graph = api_graph(document)
+    if graph is None:
+        return []
+    bound = _bound_inputs(document)
+    seed_feeders = {
+        str(value[0])
+        for node in graph.values()
+        if isinstance(node, dict) and isinstance(node.get("inputs"), dict)
+        for name, value in node["inputs"].items()
+        if _is_link(value) and SEED_FIELD_RE.search(name)
```
lindkvis: **Seed typing disagrees with the run.** A passthrough primitive is a seed only if the input it feeds matches `SEED_FIELD_RE`, even when object_info is available, while `detect_seed_targets` decides from the consumer's `control_after_generate`.

Repro: `PrimitiveInt {value: 5}` -> `CustomSampler.rng` declared `["INT", {control_after_generate: true}]`. `detect_seed_targets` returns `('1','value')` (re-rolled each run), but `describe_parameters` types it `int` and `default_pins` skips it. The owner sets a value and the run overwrites it.

pixlstash/services/workflow_parameters.py
```diff
@@ -0,0 +1,382 @@
...
+    find_input_spec,
+)
+from pixlstash.services.workflow_hash import MODEL_EXTENSIONS, SEED_FIELD_RE
+from pixlstash.services.workflow_inputs import node_title
+from pixlstash.services.workflow_io import detect_workflow_io
+
+# Anchored on the last word, unlike the library's SECRET_FIELD_RE: that one
+# scrubs stored documents and may over-match, but here an over-match hides a real
+# setting (``token_normalization``, ``max_tokens``, ``author``).
+_CREDENTIAL_RE = re.compile(r"(^|_)(api_?key|token|password|secret)$", re.I)
```
lindkvis: **Credential regex misses common names.** `(^|_)(api_?key|token|password|secret)$` doesn't match `secret_key`, `access_key`, `private_key`, `auth`, `authorization`, or camelCase `hfToken` / `authToken`, so GET /parameters returns those stored values in plaintext. The existing `SECRET_FIELD_RE` would catch most of them.

tests/test_workflows_api.py
```diff
@@ -1278,3 +1279,244 @@ def test_deleting_a_workflow_forgets_its_setup(workflow_env, edit_workflow):
...
+    assert owner.get(missing).status_code == 404
+
+
+def test_deleting_a_workflow_forgets_its_pins(workflow_env, sampler_workflow):
+    owner = workflow_env.owner
+    r = owner.put(
+        sampler_workflow.pins, json={"pins": [{"node_id": "1", "name": "steps"}]}
+    )
+    assert r.status_code == 200, r.text
+    r = owner.delete(f"{API}/comfyui/workflows/sampler.json")
```
lindkvis, line 1566: **This test touches the developer's real data dir and trash.** The DELETE route writes `sampler.<hash>.json` into the real, unredirected `workflow_inbox_dir()` (shared platformdirs dir) and then `send2trash`es it. Every run leaves another copy in `~/.local/share/Trash` (the sibling edit test already has left ten `edit.ca5c7c8564a46ff0*.json` there), and a PixlStash instance on the same data dir can import it first. The fixture patches `_workflow_dirs` and `workflow_user_dir` but not `workflow_inbox.workflow_inbox_dir`.

pixlstash/services/workflow_parameters.py
```diff
@@ -0,0 +1,382 @@
...
+                if class_type in SEED_PASSTHROUGH_CLASSES
+                else None
+            )
+            parameters.append(
+                _typed(str(node_id), title, class_type, name, value, spec, is_seed)
+            )
+    return parameters
+
+
+def default_pins(parameters: list[Parameter]) -> list[tuple[str, str]]:
```
lindkvis, line 179: **Default pins miss primitive-driven width/height.** Width/height are matched by input name, but when wired from a `PrimitiveInt` they're connected inputs and skipped, and the primitive's `value` is never featured.

The shipped `Flux2-Klein-t2i` built-in sets size via `PrimitiveInt` 75:68 / 75:69 (1024): its default pins are sampler_name, steps, cfg, models, noise_seed, with no width or height. The main size controls of the flagship template end up under "All N parameters".

pixlstash/services/workflow_parameters.py
```diff
@@ -0,0 +1,382 @@
...
+    return node_id, name
+
+
+def _check_value(parameter: Parameter, value: Any) -> None:
+    label = f"{parameter.node_id}.{parameter.name}"
+    kind = parameter.kind
+    if kind in (INT, SEED):
+        if isinstance(value, bool) or not isinstance(value, int):
+            raise ValueError(f"{label} must be a whole number")
+    elif kind == FLOAT:
```
lindkvis, line 434: **NaN / ±Infinity pass `_check_value`.** `nan < min` and `nan > max` are both False, and the body parser accepts `NaN`/`Infinity` literals. `apply_values(graph, params, [{node_id:'3', name:'cfg', value: float('nan')}])` returns `cfg=nan`; `Infinity` passes any float with no max. Once #1307 wires this in, that reaches ComfyUI instead of a 400. `math.isfinite` check suggested.

lindkvis (commented):
Review at `3eaf934c`. Top items were reproduced against the PR tree (shipped built-ins plus small synthetic graphs). The first five comments (picture fields in the form, seed typing, credential regex, the delete test hitting the real inbox/trash, Flux2-Klein default pins) look worth fixing before merge; the rest are lower severity or cleanup. New unit tests pass.

pixlstash/services/workflow_parameters.py
```diff
@@ -0,0 +1,382 @@
...
+        value = item.get("value")
+        _check_value(parameter, value)
+        graph[key[0]]["inputs"][key[1]] = value
+    return updated
+
+
+def _node_order(node_id: str) -> tuple:
+    """Numeric ids in number order, so node 10 does not sort before node 9."""
+    return tuple(
+        (0, int(part), "") if part.isdigit() else (1, 0, part)
```
lindkvis: **`isdigit()` then `int()` can raise.** `isdigit()` is True for characters `int()` rejects (e.g. superscripts). A node id like `"²"` makes `sorted(graph, key=_node_order)` raise `ValueError`, which `_parameters_of` doesn't catch (only `WorkflowGraphError`), so GET /parameters and PUT /pins return 500 instead of 422. `str.isdecimal()` avoids it.

pixlstash/routes/comfyui.py
```diff
@@ -1233,6 +1304,159 @@ def shas_by_id(session):
...
+            ) from exc
+
+    def _read_parameters(request: Request, workflow_name: str, *, typed: bool) -> dict:
+        """Load a workflow file and describe its parameters for the API, or raise 4xx.
+
+        With *typed* the parameters are typed from ComfyUI's ``object_info``;
+        when it cannot be reached they are still described, untyped, and
+        ``comfyui_error`` says why.
+        """
+        name, document, _inputs = _read_picture_inputs(workflow_name)
```
lindkvis: **UI-format files with an unreducible graph get 422 instead of `readable: false`.** `_read_picture_inputs` runs `detect_workflow_io` and converts any failure to 422. A UI workflow with a subgraph instance missing from `definitions.subgraphs` raises `MissingSubgraphDefinitionError` here, although `api_graph()` is None and `describe_parameters` would return `[]` with `readable: false` as documented.

pixlstash/routes/comfyui.py
```diff
@@ -1233,6 +1304,159 @@ def shas_by_id(session):
...
+                    "[comfyui] Parameters of %s are untyped, ComfyUI not reachable "
+                    "at %s: %s",
+                    name,
+                    comfyui_url,
+                    exc,
+                )
+                comfyui_error = str(exc)
+        parameters = _parameters_of(name, document, object_info)
+        return _describe_parameters(
+            name, document, parameters, object_info is not None, comfyui_error
```
lindkvis: **`typed` doesn't match its documented meaning.** `object_info is not None` is False for every UI-format file even with ComfyUI reachable, and True when none of the graph's classes exist in object_info. `GET .../canvas.json/parameters` answers `typed: false, comfyui_error: null`, and a client following integration_architecture.md ("`typed: false` means ComfyUI could not be reached") shows an offline banner for a healthy ComfyUI.

pixlstash/services/workflow_parameters.py
```diff
@@ -0,0 +1,382 @@
...
+        seed = SEED_FIELD_RE.search(name) if is_seed is None else is_seed
+        return SEED if seed else INT
+    if isinstance(value, float):
+        return FLOAT
+    if is_model_field:
+        return MODEL
+    return STRING
+
+
+def _number(value: Any) -> Optional[int | float]:
```
lindkvis, line 390: **Offline typing depends on the stored JSON value.** An integer-valued FLOAT input (Flux2-Klein built-ins store CFGGuider `cfg` as 1 or 5) is typed `int` offline and `float` online, so the offline form offers whole numbers only and `apply_values` with that description refuses 1.5. Models (extension rule) and seeds (name vs `control_after_generate`) shift the same way, so default pins differ by ComfyUI availability.

pixlstash/routes/comfyui.py
```diff
@@ -1233,6 +1304,159 @@ def shas_by_id(session):
...
+        if "pins" not in payload:
+            raise HTTPException(status_code=400, detail="pins is required")
+        requested = payload["pins"]
+        pins = None
+        if requested is not None:
+            try:
+                pins = workflow_parameters.validate_pins(parameters, requested)
+            except ValueError as exc:
+                raise HTTPException(status_code=400, detail=str(exc)) from exc
+        replace_parameter_pins(hub, name, pins)
```
lindkvis: **Pins write races with DELETE.** PUT /pins reads, validates, then writes the hub row without holding `workflow_inbox.INBOX_LOCK` or re-checking the file exists. If DELETE trashes the file and runs `replace_parameter_pins(None)` before this commit, the row survives, and per the delete route's own comment it comes back into force if a same-named file is imported later.

pixlstash/services/workflow_parameters.py
```diff
@@ -0,0 +1,382 @@
...
+        and len(value) == 2
+        and isinstance(value[0], str)
+        and isinstance(value[1], int)
+    )
+
+
+def _bound_inputs(document: dict) -> set[tuple[str, str]]:
+    """``(node_id, field)`` of every input a run fills."""
+    bound = set()
+    for targets in workflow_bindings.run_targets(document).values():
```
lindkvis: **The graph is reduced three times per request.** `_read_picture_inputs` runs `detect_workflow_io`, then `_bound_inputs` calls `run_targets(document)` (detects again, since `detected=` isn't passed), then `detect_workflow_io(document)` again. Compute once and pass it through.

pixlstash/routes/comfyui.py
```diff
@@ -1233,6 +1304,159 @@ def shas_by_id(session):
...
+        """
+        name, document, _inputs = _read_picture_inputs(workflow_name)
+        object_info, comfyui_error = None, None
+        if typed and workflow_parameters.api_graph(document) is not None:
+            user = server.auth.get_user_for_request(request)
+            comfyui_url = getattr(user, "comfyui_url", None) if user else None
+            comfyui_url = (comfyui_url or DEFAULT_COMFYUI_URL).rstrip("/")
+            # ponytail: one object_info fetch per read; cache it if forms open slowly.
+            try:
+                object_info = fetch_object_info(comfyui_url)
```
lindkvis: **No cache on `/object_info`.** Every GET /parameters fetches and parses the full object_info, even for a graph with no parameters, and holds a threadpool worker up to the 15 s timeout when the ComfyUI host drops packets. A short TTL cache keyed on `comfyui_url`, or skipping the fetch when `describe_parameters(document)` is empty, would avoid it (acknowledged in the ponytail note).

pixlstash/services/workflow_parameters.py
```diff
@@ -0,0 +1,382 @@
...
+
+def _node_order(node_id: str) -> tuple:
+    """Numeric ids in number order, so node 10 does not sort before node 9."""
+    return tuple(
+        (0, int(part), "") if part.isdigit() else (1, 0, part)
+        for part in str(node_id).split(":")
+    )
+
+
+def _is_link(value: Any) -> bool:
```
lindkvis: **Duplicated link predicate.** `_is_link` re-implements `workflow_hash._is_link` but drops its `not isinstance(value[1], bool)` guard, so `['4', True]` is a link here and a widget value in the hash. `api_graph` also repeats the prompt-envelope unwrap from `run_targets`, `detect_workflow_io` and `workflow_inputs._raw_node`. Suggest exporting `workflow_hash._is_link` and one envelope helper, as this PR already did for `SEED_FIELD_RE`, `find_input_spec` and `node_title`.

pixlstash/routes/comfyui.py
```diff
@@ -1233,6 +1304,159 @@ def shas_by_id(session):
...
+            return workflow_parameters.describe_parameters(document, object_info)
+        except WorkflowGraphError as exc:
+            logger.warning(
+                "Failed to read the parameters of workflow %s: %s", name, exc
+            )
+            raise HTTPException(
+                status_code=422, detail="This workflow's graph could not be read."
+            ) from exc
+
+    def _read_parameters(request: Request, workflow_name: str, *, typed: bool) -> dict:
```
lindkvis: **Small cleanups.** `typed` is keyword-only but the single caller always passes True (the False branch is dead and untested); `api_graph(document)` is evaluated three times per request; and the comfyui_url resolution block is now pasted a seventh time (976, 1330, 1470, 1559, 1721, 1891, 2010). Suggest dropping the flag, computing `readable` once, and a `_comfyui_url(request)` helper.

pixlstash/services/workflow_parameters.py
lindkvis: Fixed in 32523196. `workflow_io.py` now has one loader-class to picture-field map, `PICTURE_INPUT_FIELDS` / `picture_fields()` (`PixlStashPictureLoader` → `picture_ids`, `Image Load` → `image_path`, plus the `INPUT_IMAGE_FIELDS` entries). `_bound_inputs` binds each detected input through it. I did not switch `run_targets` to the map: a run would then write a path into `picture_ids`, which expects ids. Test: `test_a_loaders_own_picture_field_is_bound_whatever_it_is_called`.

pixlstash/services/workflow_parameters.py
lindkvis: Fixed in 32523196. A passthrough primitive is now a seed when an input it feeds is one, decided the way `detect_seed_targets` decides it: from the consumer's `control_after_generate` when object_info describes the consumer, and from the name rule only when it doesn't. Test: `test_a_primitive_is_a_seed_when_its_consumer_rerolls_the_input` (your `CustomSampler.rng` case, which also checks it lands in the default pins).

pixlstash/services/workflow_parameters.py
lindkvis: Fixed in 32523196. The credential check is a case-blind substring match again (`api_?key|access_?key|private_?key|secret|passw(or)?d|token|auth`). It carries a short exception list for names that only contain the word: `tokens` / `max_tokens`, `token_normalization`, `tokenizer`, `author`. Tests are parametrized over all of your names, plus the exceptions that must stay.

tests/test_workflows_api.py
lindkvis: Fixed in 32523196 for both fixtures. The #1305 `edit_workflow` fixture had the same hole. `_isolate_workflow_folders` now redirects `workflow_inbox.workflow_inbox_dir` to the test's tmp dir and replaces `send2trash` in both `workflow_inbox` and the route with a plain remove. I checked: the delete tests no longer add anything to the trash. I did not empty the copies earlier runs left there.

pixlstash/services/workflow_parameters.py
lindkvis: Fixed in 32523196. A parameter now carries `drives`, the inputs a passthrough primitive is wired into, and `default_pins` features a primitive that drives a featured name. Test: `test_flux2_klein_pins_its_size_primitives_by_default` reads the shipped built-in and checks `75:68` and `75:69` are pinned.

pixlstash/services/workflow_parameters.py
lindkvis: Fixed in 32523196. Numbers must be finite (`math.isfinite`), both for typed parameters and for the looser untyped check. Test: `test_a_non_finite_number_is_refused_typed_or_not` covers NaN, inf and -inf, typed and untyped.

pixlstash/services/workflow_parameters.py
lindkvis: Fixed in 32523196: `isdecimal()`. Test: `test_a_node_id_that_only_looks_numeric_sorts_without_raising` (node id `²`).

pixlstash/routes/comfyui.py
lindkvis: Fixed in 32523196. The parameters route no longer goes through `_read_picture_inputs`. It loads the file (`_load_stored_workflow`) and runs detection only when `api_graph()` finds an API graph. A UI file whose reduction would raise now answers `readable: false`. Test: `test_a_ui_file_the_reduction_refuses_is_still_just_unreadable`, which first asserts that detection really raises on that file.

pixlstash/routes/comfyui.py
lindkvis: Fixed in 32523196 by making the contract say what the code can know. `comfyui_error` is the outage signal. Top-level `typed` means object_info was read, and it is also false, with no error, for a file with nothing to set, which is never sent to ComfyUI. Each parameter now carries its own `typed`, so a class ComfyUI doesn't know no longer reads as typed. The integration doc and the response model docstrings say this. The UI-file test asserts `typed: false, comfyui_error: null`.

pixlstash/services/workflow_parameters.py
lindkvis: Fixed in 32523196 on the value side, which is what the run depends on (#1307 checks run values against the untyped description). An untyped parameter's value is now held to the broad type of the stored value only (bool, number, text), so a `cfg` stored as `1` takes `1.5`. Test: `test_an_untyped_number_takes_any_number_so_an_integer_cfg_takes_a_fraction`. The guessed kind and the default pins can still differ between offline and online. I left that: offline there is no way to know the real type, and the per-parameter `typed` flag tells a form the kind is a guess.

pixlstash/routes/comfyui.py
lindkvis: Fixed in 32523196. `PUT /pins` now loads, validates and writes under `workflow_inbox.INBOX_LOCK`, the lock the delete takes while it trashes the file. A PUT that starts after the trash gets a 404, and one that starts before it finishes before the delete forgets the pins. No test: it is a timing window, like the inbox locking in #1356.

pixlstash/services/workflow_parameters.py
lindkvis: Fixed in 32523196. `describe_parameters` takes `detected=` and passes it to `run_targets`, and `_bound_inputs` reads `detected.picture_input_classes` instead of detecting again. The route detects once and hands the result to both the untyped and the typed description.

pixlstash/routes/comfyui.py
lindkvis: Partly done in 32523196. The route now describes the file untyped first and only fetches object_info when there is something to type, so a file with nothing to set never waits on ComfyUI. Test: `test_a_workflow_with_nothing_to_set_does_not_ask_comfyui`. I left the TTL cache out for now, as the ponytail note says: the recipe preflight does the same uncached fetch, and I'd rather add one cache for both paths once the form shows whether it is slow. Leaving this open for you to decide.

pixlstash/services/workflow_parameters.py
lindkvis: Fixed in 32523196. `workflow_hash.is_link` is exported (with its bool guard) and used here. `api_graph` moved to `workflow_io` and is shared by `detect_workflow_io`, `workflow_inputs._raw_node` and this module. `run_targets` keeps its own unwrap because it needs the `["prompt"]` path prefix as well.

pixlstash/routes/comfyui.py
lindkvis: Fixed in 32523196. The `typed` flag and `_read_parameters` are gone, since the GET is the only reader. `readable` is computed once. `_comfyui_url(user)` replaces all six pasted URL blocks in `comfyui.py`, the websocket proxy's included.

PR #1363, opened 2026-09-15
Recipes: record how each picture was made

### LINKED ISSUE #1311: Recipes: record how each picture was made
Add `recipe_asset`, `recipe_instance`, `generation` and `generation_input`, and a pass that fills them for existing pictures.

When it's done, the model shelf can show how many pictures used a model, including pictures imported before this change.

lindkvis:
Closes #1311.

## What changed
- **`workflow_recipe_instance`** (hub table): the instance's parameters as a document, with model and image names replaced by references. It is keyed by `(library_uuid, instance_hash)`. A row stays while a picture (Scrapheap included) or a ghost in that library carries the hash. The covered-ghost cascade removes it after that. Forgetting a library deletes its rows.
- **`generation`** (picture id, seed stored as text because ComfyUI seeds go past SQLite's INTEGER range) and **`generation_input`** (the resolution lock), both vault tables. Migration 0118 adds them and both cascade with the picture: what outlives a destroyed picture stays the ghost's job.
- **Backfill:** 0118 clears the scanned marker on pictures that have a workflow, so the ComfyUI extraction reads them once more. On that pass it never replaces existing keys with NULL (for example, a file on an unplugged drive). It also never re-files asset names that were forgotten. Scan progress still counts those pictures as read.
- **Shelf:** model rows gain `pictures_verified` (the loader named the model by its sha256) and `pictures_by_filename` (the filename matches). It is computed from the recipe keys pictures already carry, so it covers older imports straight away. There is no UI for it yet.

## Deliberate choices a reviewer may question
- **The issue's `recipe_asset` is the existing `workflow_recipe_asset`**, unchanged. The companions plan says the two should merge. Matching a recipe's models to the shelf happens when the counts are read, so the result corrects itself when models are added or removed.
- **The table is named `workflow_recipe_instance`, not `recipe_instance`,** to match the other hub tables. It is keyed by library, which is narrower than the 2026-08-20 "hub-side with the recipe" ruling. A global row could not be checked against any single vault's pictures, the same argument that keyed ghosts by library.
- **The hub now holds the prompt of every kept picture, not only of ghosts.** That follows the 2026-08-20 ruling: the prompt is already readable in its own library. Detaching a library keeps these rows, exactly as it keeps ghosts.
- **No `generation_input` rows are backfilled.** A `LoadImage` names a file in ComfyUI's input folder, and a Picture Loader names ids from whichever library built it. Neither identifies a picture, so any lock written from them would be invented.
- **Known gaps, which fail toward keeping less:**
  - Pictures whose file is unreadable during the backfill get a generation with no seed and no instance row, and are not retried. Retrying would re-open missing files on every planning cycle.
  - A cascade drain that races a new picture with the same instance, or a purge's ghost write, can leave that picture or ghost without an instance row.
  - A future `HASH_VERSION` bump must not rely on the keep-keys rule, and the code says so.
- **Shelf cost:** each shelf list now runs one grouped vault query and one read of the hub's asset table. That is small next to the queries the list already runs.


lindkvis (commented):
Reviewed the migration, extraction revisit, cascade, hub instance table and shelf counts. Merges cleanly with develop and with #1362; 0118 follows the current head (0117). Two comments inline: one is a credential leak into the new hub table, the other a cost on erase.

pixlstash/services/workflow_hash.py
```diff
@@ -553,6 +552,41 @@ def document_from_reduction(nodes: dict[str, ReducedNode]) -> dict:
...
+    """``value`` with every filename-shaped string in it swapped for a reference."""
+    if isinstance(value, dict):
+        return {
+            key: _nested_assets_as_references(str(key), item)
+            for key, item in value.items()
+        }
+    if isinstance(value, list):
+        return [_nested_assets_as_references(name, item) for item in value]
+    asset = structural_widget_value(name, value)
+    return asset_reference(asset) if asset is not None else value
```
lindkvis, line 601: **Nested credential-named keys are stored in `workflow_recipe_instance.document`.** `reduce_api_graph` drops `SECRET_FIELD_RE` names only at the top level of a node's inputs. This recursion keeps every other nested value as-is, so the credential check never runs on nested keys.

Reproduced against this head:
```python
g = {"1": {"class_type": "Power Lora Loader (rgthree)", "inputs": {
    "lora_1": {"lora": "x.safetensors", "api_key": "test-leak", "seed": 42},
    "api_key": "test-top"}}}
instance_document_from_reduction(reduce_api_graph(g))
# {"1": {..., "inputs": {"lora_1": {"lora": "asset:77ca...", "api_key": "test-leak", "seed": 42}}}}
```
The top-level `api_key` is dropped, but the nested one is written to the hub. Before this PR a nested value only fed the hash. Now it is stored and kept indefinitely. A nested `seed` also gets in, even though the docstring says seeds stay null.

Suggested fix: in this function, skip dict keys that match `SECRET_FIELD_RE` and null nested keys that match `_SEED_RE` / `_OUTPUT_PATH_RE`. Extend `test_a_credential_widget_reaches_no_tier_including_the_instance` with a nested case.

pixlstash/services/workflow_ghost_service.py
```diff
@@ -426,20 +467,46 @@ def dequeue(session: Session):
...
+    """Erase every ghost one library holds, and re-judge what they were keeping.
+
+    An instance row a ghost alone was keeping has lost its reason with the
+    ghost, and no picture delete will ever queue its hash, so the erase queues
+    the library's retained hashes for the cascade itself. Returns how many
+    ghosts were erased.
+    """
+    erased = erase_picture_ghosts(hub, library_uuid)
+    try:
+        requeue_library_ghosts(vault_db, hub, library_uuid)
```
lindkvis: **Erasing ghosts now re-queues every instance hash in the library, not just the ones the ghosts covered.** Ghosts are gone by the time `requeue_library_ghosts` reads `retained_instance_hashes`, so it returns every `workflow_recipe_instance` row. That is about one per distinct prompt in the library. So one click re-checks the whole library's instances against the vault, when only the erased ghosts' hashes can have lost cover.

Suggested fix: read `SELECT DISTINCT instance_hash FROM workflow_picture_ghost WHERE library_uuid = ?` before the DELETE (or have `erase_picture_ghosts` return the hashes) and queue only those. `requeue_library_ghosts` can keep the full sweep for restore, which really does need it.

pixlstash/services/workflow_hash.py
lindkvis: Fixed in 15c03034. The nested recursion now applies the rules a top-level input already had. A key matching `SECRET_FIELD_RE` is dropped, and a seed, `filename_prefix` or output-path key is stored as null. The instance hash is still computed from the raw widgets, so no existing key changes. `test_a_nested_credential_seed_or_output_path_is_never_stored` covers the nested credential, seed and output path. It fails if the secret filter or the seed nulling is removed.

pixlstash/services/workflow_ghost_service.py
lindkvis: Fixed in 15c03034. `erase_library_ghosts` now reads the ghosts' instance hashes before the erase (`hub/workflows.ghost_instance_hashes`) and queues only those. `requeue_library_ghosts` keeps the full sweep for restore. The erase test now adds a second instance that has no ghost and asserts that only the erased ghost's hash is queued. Swapping the full requeue back in makes it fail.

Issue #1357, opened 2026-09-15
Tagger rewrite silently deletes manual tags added while a new picture's auto-tag pass is pending


lindkvis:
## Summary

When a tag is added to a newly imported picture while its background `TagTask` is still pending or running, the tagger's write step deletes every `Tag` row on that picture and re-inserts only (a) the model's output and (b) tags that have a `label_source='human'` POS row in `tag_prediction`. A manually added tag outside the hardcoded `ANOMALY_LABELS` vocabulary never gets that human ledger row, so the tagger silently deletes it. The operation log does not record the deletion, so it cannot be undone and nothing shows that it happened.

The API returns success, and a read right after the add shows the tag. The tag disappears minutes later, when the tag task commits. The window depends on how long the tagger takes: here JoyCaption was the active full-pass plugin and its first cold load took about 275 s.

Seen on v1.11.2rc1 (desktop AppImage). The relevant files are unchanged between `v1.11.2rc1` and `origin/main`.

## Reproduction shape

1. Pick a tagger that is slow to start (e.g. JoyCaption on first load, or any large batch).
2. Import pictures with `POST /api/v1/pictures/import` and wait until `/pictures/import/status` reports completion.
3. Right away (within seconds to a few minutes), add a tag that is **not** in `DEFAULT_SMART_SCORE_PENALIZED_TAGS` / `DEFAULT_TAG_MERGES` with `POST /pictures/{id}/tags` (here: a manually set anomaly tag the deployed PixlStash tagger supports but the smart-score table does not list). Also add a tag that *is* in the table, as a control.
4. Read the tags back with `POST /pictures/tags/bulk_fetch`: both are present.
5. Wait for the `[TAG_TIMING]` log line for that batch, then read the tags again. The unlisted tag is gone. The listed tag is still there.

## Observed evidence (one library, 198 pictures, two import batches)

Server log timeline (local time):

| Event | Batch 1 (63 pictures) | Batch 2 (100 + 35 pictures) |
|---|---|---|
| Import completed | 17:19:29 | 23:21:18 / 23:21:27 |
| Manual tags added (`operation` rows `pictures.tags.add`) | 17:19:51-52 | 23:21:50-52 |
| Client re-verify (`bulk_fetch`): all tags exact | ~17:20:12 | ~23:22:12 |
| JoyCaption cold load | 17:19:39 -> 17:24:14 (274.9 s) | already loaded |
[...]

PR #1364, opened 2026-09-15
Retag no longer keeps a stale in-flight task's tags

### LINKED ISSUE #1361: Retag preserves stale model tags from an in-flight task
## Problem

The tagger write introduced in PR #1358 treats every existing non-sentinel `Tag` row as a tag written by another writer and preserves it. That is not true when an older tag task is already in flight.

## Reproduction

1. A background tag task captures a picture and starts inference.
2. The user requests a retag (possibly with a different model). The reset deletes the tags and writes the new retag sentinel.
3. The older task finishes first, clears the sentinel, and writes its stale model tags.
4. The newly requested task writes afterward. Because the stale rows are now treated as manual/foreign tags, it only adds its own tags.

The final `Tag` rows contain the stale and fresh model outputs (`stale-model-tag` + `fresh-model-tag`) instead of only the fresh result. This is especially likely because the older task can already be occupying the tagger worker when the reset queues the new interactive task.

## Impact

A user's explicit retag can silently retain tags that the selected fresh model no longer produces, and model selection can result in a union of outputs from two models.

## Suggested direction

Track tag provenance or add a generation/token to the retag sentinel and reject writes from tasks started before the reset. The tagger should preserve only tags known to have been added by a human/other durable writer, not output from an earlier tag task.

See [the preservation logic](https://github.com/Pikselkroken/pixlstash/blob/6d9aad75ec2234692b99a49893bc8c82b20d70ae/pixlstash/tasks/tag_task.py#L630-L640).,--label,bug,--label,priority:high

lindkvis:
Fixes #1361.

A tag task that read a picture before a retag could finish afterwards, clear the new sentinel and write its tags. The retag's own task then kept those rows as another writer's, so the picture ended up with tags from both runs.

**Change:** `TagResetRegistry` (in memory, on `VaultDatabase.tag_resets`) keeps a reset generation per picture. Writers that replace tags in place of the tagger call `mark_reset` after they commit: `reset_pictures_tags`, sidecar sync in the reference-folder scan, the metadata import route, and duplicate import with sidecar tags. `TagTask` reads the generation before its pictures are loaded; the finder and `retag_picture_interactive` pass it in. On the writer thread, the task drops its tag and prediction writes for any picture reset since then. The sentinel stays, so the next pass tags the picture.

Deliberately not covered:
- **Manual replace, clear-all and undo do not mark a reset.** They keep #1357's behaviour: a manual edit made while tagging is pending gets the tagger's tags added beside it.
- **`reset_generation` defaults to "now".** That default is correct for any caller that reads its pictures just before building the task. Both production callers pass an explicit value.
- **`TagPredictionBackfillTask` is unchanged.** It can still write prediction rows for a picture reset mid-backfill. That problem predates #1358 and belongs in its own change.
- **`_resolve_pending_predictions` is left unfiltered.** It recomputes status from the current Tag rows and skips human labels, so it cannot bring stale output back.

The new test in `test_operation_log.py` reproduces the issue and fails when the generation check is disabled.


pixlstash/tasks/reference_folder_scan_task.py
```diff
@@ -886,6 +886,7 @@ def apply_caption_updates(
                                 )
                     session.flush()
                 session.commit()
+                self._db.tag_resets.mark_reset(retagged)
```
lindkvis, line 889: **Only one of the four writers that mark a reset is covered by a test.** The new test goes through `reset_pictures_tags`. Delete this `mark_reset`, or the ones in `reference_folders.py` import_metadata or `_import.py`, and nothing goes red: no test touches `tag_resets` apart from the fake in `test_vram_batch_budget.py`. The same is true of the finder's capture-before-read in `missing_tag_finder.py`. Move it after the read (or drop it and use the constructor default) and the #1361 window reopens with the suite still green.

Suggested: one parametrised test in the same module that builds a stale task, runs each sidecar path, and asserts `_add_tags_unless_reset` writes nothing. Show each case failing once with its `mark_reset` removed.

lindkvis (commented):
The ordering holds. Resets and the task's two guarded writes all run on the single writer thread, and marking after the commit means the only race makes a task look stale when it isn't, never fresh when it's stale. Both production `TagTask` constructors capture the generation before they read. The generation is global but compared per picture, so resetting other pictures never drops a write. A reset landing between the tag write and the prediction write is also safe, because the reset deletes the rows the tag write just added. Merges cleanly with develop.

One comment inline, on test coverage.

pixlstash/tasks/reference_folder_scan_task.py
lindkvis: Covered in be80a1d7. Each writer now has a test: a tag task reads the picture, the writer replaces its tags, and then the task's late write must write nothing. I added each test next to an existing one that already sets up that path:

- scan sidecar sync and the metadata import route: `tests/test_reference_folder_sidecars.py`
- duplicate import with a sidecar: an extra assertion in `tests/test_server.py::test_duplicate_import_with_sidecar_replaces_existing_tags`
- finder capture order: `tests/test_pipeline_stage_contracts.py::test_the_tag_finder_captures_the_reset_generation_before_its_read`. A reset lands right after the finder's read, and the task must count as older than it.

I didn't use one parametrised test because the three writers need different environments (a reference folder with a scan, the route, the import upload). I checked each one locally: removing that writer's `mark_reset`, or passing `reset_generation=None` from the finder, turns its test red.

PR #1366, opened 2026-09-15
Workflows: run a workflow

### LINKED ISSUE #1307: Workflows: run a workflow

lindkvis:
Closes #1307. Builds on #1362 (#1306, workflow parameters), whose commits are in this diff until it merges.

**Backend.** `POST /api/v1/comfyui/workflows/{workflow_name}/run` (owner only) runs a saved API-format workflow and fills each picture input by its #1305 mode:
- **Selection** input: runs once per selected picture and stacks each output with that picture.
- **No Selection input:** runs once, filed into the view context.
- **Picker** inputs come from the request. **Fixed** inputs come from the stored setup (409 if the picture has left the library).

Outputs are collected from the detected save nodes. `values` from the #1306 parameters are applied, and `seed_mode` can be `keep`. Every refusal happens before anything is uploaded. A ComfyUI failure partway through a batch returns `status: "partial"` with the runs that did start. Uploads are named `pixlstash-<id>-<pixel_sha>`, so same-named files no longer overwrite each other in ComfyUI's input folder. The list gains `runnable`.

**Frontend.** A run panel in the inspector rail replaces the selection pill's ComfyUI menu and the toolbar's T2I menu. The grid stays usable while it is open. The pill offers workflows with a Selection input and the toolbar offers the rest. The Run button states the count ("Run 12 times"). Overlay and Remix are unchanged and still use `run_i2i`/`run_t2i`.

**Not done:**
- "Record the resolution lock as today": I couldn't tell what it refers to, so it is left out.
- The #1306 parameter form is not in the panel yet; the #1306 session plans to add it here.

**Pre-push review findings, deliberately left:**
- `runnable` can be true for a workflow the route still refuses, e.g. a loader PixlStash can't fill. The route says why.
- An old binding on an unrecognised loader is ignored when other picture inputs are detected. Documented.
- A toolbar run no longer sends `source_picture_id` for a single selected picture. Toolbar runs now take no selection by design.
- There is no cap on selection size (same as `run_i2i`). The count is shown before running.


lindkvis:
Review by Qwen 3.8 Flash Next:

Review — PR #1366 "Run a saved ComfyUI workflow from the right-hand rail"
Verdict: request changes. The feature is well-scoped and the surrounding hygiene is good (changelog fragment + authz matrix updated in the same PR), but there are two issues I'd block on — one security-relevant validation gap, one CI/test-coverage mismatch that would let regressions through silently.

I should be upfront: the diff is large enough that it overflowed my working context, so line-level citations below are by file/function rather than line number. Everything marked ✅ I confirmed in the diff; items marked ⚠️ need a quick check on your side because I couldn't re-read the exact lines.

Blocking
1. ✅ Body-supplied input values skip the allow-list that setup enforces (pixlstash/comfyui/workflow_setup.py::_apply_input vs workflow_run.py)
Setup validates Select/Combo values against the node's declared options and returns 400 for anything not in the list. The run path only applies that check to values coming from the stored setup (if chosen and chosen not in values). Values resolved from the request body — i.e. by_picture / one_for_all inputs, which is the whole point of the new route — are written straight into the prompt dict unvalidated.

Net effect: a caller can put arbitrary text into any input the workflow exposes, including combo/Select nodes whose option set _apply_input exists precisely to police. Not remotely exploitable beyond the owner (the route is owner_only), but it defeats an intentional invariant and lets a hand-crafted body push values ComfyUI will resolve however it likes.

Fix: make run go through the same helper as setup rather than re-implementing value handling — one _apply_input(node_def, value) used by both paths. If that's hard because run resolves lazily per-picture, at minimum apply the option check whenever input_type in ("select", "combo") regardless of whether the value came from setup or body.

2. ⚠️ The CI change looks copy-pasted from #1306 — new run tests may never run
The ci.yml hunk adds tests/test_workflow_parameters.py, which is #1306's test file, not a run/panel test. Two possibilities, both need resolving before merge:

The new tests live in a file like tests/test_workflow_run.py that isn't in the list → CI never executes them, and the PR is effectively untested on CI.
There are no new backend tests for run at all, only frontend ones.
Please confirm which test files this PR adds and that each appears in the CI job's path list. Given #1306 already landed test_workflow_parameters.py, adding it again here is a strong hint of the first case.

Should fix
3. ✅ Authz matrix justification doesn't match what the route does
The matrix declares the new POST /api/v1/comfyui/workflows/{workflow_name}/run as owner_only, with the rationale that "a Fixed input names a picture by content from the stored setup, not from the body, so no body-id scope check could cover it." That's true of Fixed inputs — but the route also accepts picture ids in the body for by_picture/one_for_all. So the stated reason describes roughly half the route's behaviour.

This isn't a privilege escalation (owner_only is strictly tighter than picture_scoped, and pictures are owner-scoped), so I'm not calling it a vuln. But the authz matrix is explicitly consumed by an adversarial review, and a rationale that doesn't match the handler is worse than no rationale — the reviewer either trusts prose that's wrong or burns time re-deriving it. Update the wording to cover both input modes. Arithmetic itself checks out: 327 declared, +1 owner_only, nothing retargeted.

4. ✅ Duplicate-sha content resolution differs between setup and run
Setup and the new _resolve_sha_map pick different winners for the same content hash when multiple pictures share a sha (one favours lowest id, the other takes the newest). Since Fixed inputs resolve by content, "the Run button shows how many runs will start" can silently disagree with what the setup preview showed for duplicated uploads.

Fix: extract one shared resolver used by both paths rather than two implementations that must be kept in sync by hand. While there: _resolve_sha_map issues a query per sha in a loop — collapse to a single WHERE sha IN (...).

5. ✅ User-input errors surface as 500
The handler raises ValueError for bad workflow name / missing inputs / deleted pictures and the route lets it fall through to a generic 500 (or catches broad Exception). The sibling setup endpoints map these to 4xx. Do the same here, and specifically: when picture ids no longer resolve, return a structured error listing which ids are gone so the panel can refresh itself instead of showing "something went wrong".

6. ✅ Unbounded fan-out
One request starts len(pictures) × runs_per_picture ComfyUI generations with no cap and no throttle. Owner-only means it's self-inflicted rather than cross-tenant, but selecting a large library (or the whole library) is one click away in the new panel. Add a max-batch constant and reject above it with 400 + the limit in the message.

7. ✅ Frontend: parameters fetch has no cancellation
The panel fetches /parameters on open and writes into state without an AbortController or effect-cleanup guard. Switching workflows quickly can land the older response last, so the panel shows workflow A's inputs while Run targets workflow B. Add an abort + a request-identity check before committing to state.

8. ✅ Frontend: no in-flight guard on Run
Nothing disables the Run button (or otherwise serialises submission) while the POST is pending, so double-clicking enqueues two batches — directly at odds with the changelog's promise that the button reflects how many runs will start. Disable while pending and keep it disabled until the count refreshes.

lindkvis:
Replying to the review above, point by point. Fixes are in b8b2f1b4. Several of the mechanisms named (`workflow_setup.py`, `workflow_run.py`, `by_picture` / `one_for_all`, `_resolve_sha_map`, `runs_per_picture`) are not in this diff, so I checked each observation against the code that is.

1. **Body values skip option checks: not changed.** `values` are the only body-supplied settings. `workflow_parameters.apply_values` checks them by type, and picture ids never become values. Option and range checks without ComfyUI's `object_info` would guess. ComfyUI validates combo options and min/max itself on `POST /prompt`, and a rejection there is a 400/502 before the first run is queued. So a hand-made body cannot push an out-of-list combo value into a running graph. Typing at run time would add an `object_info` fetch and a new failure mode. Left for the #1306 form, which reads typed parameters already.
2. **CI: already gated.** The new backend tests are in `tests/test_workflows_api.py` and `tests/test_workflow_io.py`, both already in the backend job's file list. The `ci.yml` line adding `tests/test_workflow_parameters.py` is #1362's, carried here because this branch builds on it.
3. **Matrix rationale: fixed.** It now covers both ways the route reads pictures: body ids for the selection and Picker inputs, and stored content for Fixed inputs. A scoped declaration would cover only the first, so the route takes owner-only for both.
4. **Duplicate resolution: made structural.** Both already resolved to the oldest copy (`ORDER BY id DESC`, last one wins), and both used one `IN` query, not a query per sha. They were still two copies of the rule, though. Both now read through `_oldest_kept_by_sha`, and a test runs a duplicated Fixed picture through setup and run and asserts the same copy.
5. **Errors as 500: mostly not the case; missing ids now named.** Bad names are 400/404, value errors 400 and deleted pictures 404; there is no broad `except`. The 404 for pictures that left the library now lists their ids ("Pictures not found: 3, 5").
6. **Unbounded fan-out: fixed.** `MAX_RUNS_PER_REQUEST = 200`; above it is a 400 naming the limit, before anything uploads.
7. **Fetch race: already guarded; now tested.** The panel doesn't fetch `/parameters`. Its `/inputs` read carries a request counter and drops a response for a workflow no longer chosen. A new test lands the older read last and asserts the panel keeps the current workflow's inputs.
8. **Double-click on Run: already guarded; now tested.** `run()` returns while a request is in flight, and the button's `loading` state disables it. A new test double-clicks and asserts one request. Both new frontend tests fail with their guard removed.

PR #1368, opened 2026-09-15
Drop backfill predictions for pictures retagged mid-backfill

### LINKED ISSUE #1367: Prediction backfill writes stale rows onto a picture retagged mid-backfill
## Problem

`TagPredictionBackfillTask` reads its pictures, runs the built-in tagger, and then writes predictions through `TagTask._write_predictions_from_tags`. Nothing checks whether the picture's tags were reset while inference was running. PR #1364 added that check for `TagTask` (`TagResetRegistry`), but not for the backfill.

## Reproduction (from reading the code; not reproduced)

1. A backfill task picks up a tagged picture that has no predictions and starts inference.
2. The user retags that picture with a plugin tagger. `reset_pictures_tags` deletes its non-human predictions and tags and writes the sentinel.
3. The backfill write lands. `_backfill_predictions` leaves the sentinel out of the applied set, so it writes built-in prediction rows for the picture, all `REJECTED`.
4. The plugin retag writes its own predictions. `_write_predictions_from_tags` deletes stale-version rows only on its own side of the built-in/plugin split, so the backfill's built-in rows survive.

## Impact

A picture the user explicitly retagged keeps prediction rows from a pass made against its old tags. Built-in rows feed the anomaly score, so the smart score may reflect them.

## Suggested direction

Use the same guard as `TagTask`: capture `database.tag_resets.current()` before the finder reads the pictures, and drop any picture in `reset_since(...)` inside `_backfill_predictions` on the writer thread.

lindkvis:
Fixes #1367.

`TagPredictionBackfillTask` now gets the same guard #1364 gave `TagTask`. `MissingTagPredictionFinder.find_task` captures `tag_resets.current()` before it reads the pictures. On the writer thread, the task drops every picture reset since that capture, from both the scored write and the zero-confidence write for unscored pictures. A dropped picture still has tags and no predictions, so a later sweep backfills it against its new tags.

**Depends on #1364.** This branch carries #1364's commit, because the registry comes from there. It targets develop, and that commit drops out of the diff once #1364 merges. The change for this PR is the last two commits.

Tests: a real retag lands mid-inference, once as a plugin reset on the scored path and once as a sidecar replacement on the unscored path. A second test checks that the finder captures the generation before its read. Each filter and the capture ordering were mutated, and a test went red each time.

On review: the task treats `tag_resets` as optional, and without the registry it captures at construction. The finder assumes the registry exists. This copies `TagTask` from #1364. The optional path is only for test fakes, because a real `VaultDatabase` always has the registry.


pixlstash/tasks/tag_prediction_backfill_task.py
```diff
@@ -158,6 +166,41 @@ def _run_task(self) -> dict:
...
+    def _backfill_predictions_unless_reset(
+        self,
+        session: Session,
+        label_scores_by_pic_id: dict,
+        attempted_picture_ids: list,
+        model_version: str,
+    ) -> int:
+        """``_backfill_predictions`` minus pictures retagged since the read (#1367).
+
+        Their scores were taken against the replaced tags, and a retag by a
```
lindkvis, line 178: This isn't quite the reason. The scores come from the image alone, and `_backfill_predictions` reads the applied tags again inside this same write, so neither input is stale. The harm is mixing the two populations. Once a plugin retag owns the picture's tags, this pass writes built-in rows onto it: all `REJECTED` while the sentinel is pending, and next to the plugin's rows once its task has finished. `_write_predictions_from_tags` never deletes across the built-in/plugin split, so those rows persist. As written, the docstring suggests that a manual tag edit also needs `mark_reset`, and it doesn't.

tests/test_tag_prediction_backfill.py
```diff
@@ -188,3 +192,72 @@ def test_backfill_pins_to_the_pixlstash_tagger_even_when_a_plugin_is_active(tmp_
...
         assert workflow.overrides == ["pixlstash_tagger"] * 3
         assert workflow.active_model_version("pixlstash_tagger") == "v7"
+
+
+@pytest.mark.parametrize("scored", [True, False])
+def test_backfill_drops_pictures_retagged_since_it_read_them(tmp_path, scored):
+    """A retag mid-backfill must not get rows from the old pass (#1367).
+
+    Scored: a plugin retag leaves only the sentinel, and the stale scores would
+    become REJECTED built-in rows. Unscored (decode failure): a sidecar import
```
lindkvis, line 202: The unscored case doesn't show a bug. Without the filter, the task writes `wolf` at 0.0 as `CONFIRMED`, which is the same row a fresh pass after the sidecar import would write after the same decode failure. The test proves the filter runs, but not that it prevents harm. I couldn't find a harmful unscored case: a pending plugin sentinel leaves `tags_by_pic_id` empty, and a finished plugin retag has already written its pairs, which `existing_pairs` skips. Keeping the filter as defence is fine. I'd point this case at a plugin retag whose task finished before the write, on the scored path (the case only the registry catches), or drop the "would confirm them" claim.

pixlstash/tasks/missing_tag_prediction_finder.py
```diff
@@ -34,6 +34,7 @@ class MissingTagPredictionFinder(SimpleMissingFinder):
     def __init__(self, database, engine_getter: Callable):
         super().__init__(database)
         self._engine_getter = engine_getter
+        self._reset_generation = 0
```
lindkvis, line 37: Nit: this is never read, because `find_task` always overwrites it before `_create_task` runs. Harmless, and 0 is the safe default, but a one-line comment that `_create_task` depends on `find_task` having set it would make the hand-off through instance state explicit.

pixlstash/tasks/tag_prediction_backfill_task.py
```diff
@@ -158,6 +166,41 @@ def _run_task(self) -> dict:
...
+        plugin tagger never deletes built-in rows, so they would outlive it.
+        Must run on the writer thread, where the resets also run.
+        """
+        if self._tag_resets is not None:
+            stale = self._tag_resets.reset_since(
```
lindkvis, line 183: Nit: this filter-and-log block duplicates `TagTask._reset_since_capture`. Fine at two copies. If a third task needs the guard, move it onto `TagResetRegistry`.

lindkvis (commented):
The guard looks correct. The finder captures the generation before its read, `mark_reset` runs after the reset commits, and the filter runs on the writer thread, so the worst case is a spurious drop that the next sweep picks up. A simpler check for the pending sentinel at write time would miss a plugin retag whose `TagTask` has already finished, so the registry is the right tool. My comments are about the docstring rationale and one test scenario, not the code.

PR #1369, opened 2026-09-15
Keep the budget stub test out of the real CI step summary

lindkvis:
The "backend shard exceeded its CI budget" summary (3600s vs 3000s at /4, 2700s vs 750s at /16) was not a real breach. Those are the synthetic inputs of `test_test_time_budget_fires_on_over_budget_input`: it calls `_enforce_test_time_budget` with stub sessions, and `_append_step_summary` wrote them into the real `GITHUB_STEP_SUMMARY` of whichever gate shard ran the test.

The test now points `GITHUB_STEP_SUMMARY` at a tmp file and asserts the breach landed there. It goes red if the redirect is removed. The budget and the durations map are unchanged.

A pre-push review suggested putting the guard in the shared function instead. I left it in the test: the production hook has to write to the real summary, and this is the only stub caller. The subprocess test already redirects the variable.


lindkvis:
## Review: approve

The redirect is correct and the assertion it adds is real.

`monkeypatch.setenv` at `tests/test_ci_shards.py:1439` covers every call the test makes into `_enforce_test_time_budget` (lines 1444, 1466, 1473, 1483), and monkeypatch's finalizer restores the environment on teardown whether the test passes, fails or raises — so there is no path where a synthetic breach reaches the shard's real summary. The function has exactly two side channels: the terminal reporter and `_append_step_summary` (`tests/conftest.py:654`, `657-669`). The reporter here is a stub and the `::warning` annotation never went to stdout, so the step summary was the whole leak.

I mutated the redirect out to check the assertion is load-bearing. It fails with the variable unset, and with `GITHUB_STEP_SUMMARY` pointed at a real file it reproduces the original bug — the file comes back holding a genuine-looking "spent 3600s against a ceiling of 3000s" report. Redirect restored: the file the variable names stays empty, and the whole file passes with the variable set.

**On keeping the guard in the test rather than in `_append_step_summary`** — agreed, and not only because it is the smaller diff. Writing to the real step summary is that function's entire job, and the sibling test `test_test_time_budget_annotates_without_failing_the_run` asserts exactly that it happens. A "don't write under pytest" guard inside the shared function would have hollowed that test out. There is one stub caller in the repo, so there is no class of siblings to sweep; if a second ever appears the redirect has to be repeated, but a file-local autouse fixture today would be scaffolding for a hypothetical.

Nothing else CI-relevant: no workflow, permission or duration-map change, and the test stays in its existing gated file, so shard classification is unaffected.

PR #1370, opened 2026-09-15
Read-only MCP server

### LINKED ISSUE #1316: Read-only MCP server
An MCP server that lets an agent search the library and read pictures, tags and recipes. No write tools.

Every tool goes through the existing token scopes.

Depends on: #1311

lindkvis:
Closes #1316.

Adds `pixlstash-mcp`, a stdio MCP server with six tools: `search_pictures`, `list_pictures`, `get_picture`, `view_picture`, `list_tags` and `get_recipe`.

It has no authority of its own. It is an HTTP client of a running server, and every tool is a fixed `GET` on an existing route, sent with the token in `PIXLSTASH_TOKEN`. Token scope is therefore enforced by the auth middleware and the authz gate, exactly as for any other client. It adds no routes, no registry entries and no new dependency: the protocol surface is small enough for the stdlib.

- Picture ids are coerced to int, so a tool argument cannot reshape the request path.
- Redirects are refused, because urllib would otherwise copy the Bearer header to whatever host the redirect names.
- Malformed JSON-RPC gets an error reply and does not end the session. Batches get `-32600`.

**Connecting it.** `API Tokens → Connect AI agent` in account settings mints a
READ token and shows the finished `claude mcp add` line and `mcpServers` block,
so the token is never copied into a JSON file by hand. No new routes: it calls
the existing `POST /users/me/token`.

**Tests.** `tests/test_mcp_server.py` (gated) drives the tools against a real server with a single-picture READ token. Each refusal sits beside an in-scope positive and an owner control. With `AUTHZ_GATE_ENFORCING = False` the picture tools go red, and removing the redirect guard or the int coercion also turns their tests red.

**Review points answered here rather than in code:**
- `search_pictures` is checked for request shape only, not for scope against a live server. A fresh test server's first search loads the text encoder while the planner imports the same libraries, and that fails intermittently (`accelerate` partially initialised). The race predates this PR and the search route's scoping is already covered by the existing authz suites.
- `list_tags` scoping belongs to the route (`SCOPED_LIST`), not to this client. The live test only proves a scoped token gets an answer.
- `get_recipe` pre-flights against ComfyUI, so it is marked `openWorldHint` and the README says so.
- A full-access token is accepted. The README says to use a read-only one and what happens otherwise. The client cannot tell a token's scope without an extra route.
- HTTPS with the auto-generated self-signed certificate will fail verification. The default URL is plain http on loopback. Add a CA option if anyone asks.
- `ruff check pixlstash` reports an F821 in `services/workflow_hash.py`. It is already on develop and not touched here.


lindkvis:
## Review: changes requested — one line

Backend and security review, merged. The module's whole security story is "the token only ever goes to the URL you named", and the redirect guard shows that was on your mind. The proxy is that bug's sibling and it is still open. Both reviewers found it independently.

### Must fix

**The Bearer token goes to `http_proxy` if one is set.** `pixlstash/mcp_server.py:140` — `build_opener(_NoRedirect)` correctly *replaces* the default `HTTPRedirectHandler` (the subclass-skip rule in `build_opener` is the part people get wrong, and this gets it right), but it leaves urllib's default `ProxyHandler` installed. That reads `http_proxy`/`HTTP_PROXY` from the environment, and urllib does **not** bypass loopback — it consults `no_proxy` and nothing else; `proxy_bypass("127.0.0.1:9537")` returns `False`.

So on a machine with a proxy set — corporate boxes, dev containers, anything sourcing a `proxy.sh` — every request goes to the proxy instead of `127.0.0.1`, carrying `Authorization: Bearer <API token>` and bringing library metadata back in cleartext to a third party. Demonstrated against the branch module: with no proxy env the handler list has no `ProxyHandler`; with `http_proxy=http://proxy.example.test:3128` it does, and the fetch fails with `[Errno -2] Name or service not known` — which is itself the proof, because that DNS failure is for the proxy host, not for the loopback address the process was asked to dial. It also produces a misleading error: the owner is told PixlStash is unreachable at an address the process never contacted.

```python
_opener = urllib.request.build_opener(_NoRedirect, urllib.request.ProxyHandler({}))
```

Same class of problem, same size of fix. A regression test can mirror the redirect one: set `http_proxy`, point at a local socket server, assert the hit landed on the real target.

### Should fix

**The README undersells a full-access token.** `README.md:754` says "a full-access token works too, but then the agent can read the whole library." The read-only-ness of this server is not the boundary — the token is. An `ALL` token lives in the agent's own config file, and essentially every coding agent can read its own config and make its own HTTP requests. From loopback an `ALL` token is owner-equivalent: writes, deletes, `/server/restart`, `filesystem/browse`, reference- and model-folder writes, the whole §16.3 host-capability class. The honest sentence is "any agent that can read that file has full owner control of PixlStash — not just read access to your pictures." The code is fine; the accepted risk deserves stating at its real strength.

**Non-ASCII search queries break on Windows.** `mcp_server.py:298-299` uses `sys.stdin` as-is; when stdin is a pipe Python decodes with the locale encoding, which is cp1252 there, so `search_pictures(query="café")` arrives mangled. Replies are safe — `json.dumps` is `ensure_ascii` by default — so it is one-directional. One line in `main()`: `sys.stdin.reconfigure(encoding="utf-8")`. The tests feed `io.StringIO`, so they structurally cannot catch this.

**`/pictures/search` is the one tool path with no proof it names a live route.** `mcp_server.py:222` builds the path and `tests/test_mcp_server.py:244-248` asserts the URL string the code just built, so a route rename breaks the tool with the test still green. The encoder/planner import race genuinely rules out a live search here, but not the cheap version: assert each tool's templated path appears in `server.api.routes`. Race-free, costs nothing, covers all six.

**A request with an `id` and no `method` gets no answer at all.** `mcp_server.py:251` folds "no id" and "no method" into one notification branch. JSON-RPC calls the second an invalid request (-32600); as written the client waits on that id until its own timeout.

### Nits

- `SUPPORTED_PROTOCOL_VERSIONS` advertises `2025-03-26` (`mcp_server.py:43`), the one revision that *mandates* JSON-RPC batching — which `serve` refuses with -32600 (`:313-318`). Refusing batches is fine; advertising the revision that requires them is a small untruth.
- Head-of-line blocking: the loop is serial with a 60 s timeout (`:158`, `:300`), so one slow search stalls a concurrent `ping` behind it. Fine for a single-user read-only server, though some clients have shorter liveness timeouts than 60 s.
- `raise ToolError(...)` from the URLError branch drops the cause (`:169`) — no `from exc`. The branch does log, so cosmetic.
- `list_pictures` returns full metadata for up to 200 pictures (`:225`); the route supports `fields=grid`, which would cut agent context cost noticeably. Judgment call.
- `for _tool in TOOLS` leaks `_tool` as a module global (`:124`).

### Checked and clear

**The token never leaks anywhere else.** It is touched in exactly three places — env read (`:351`), function argument (`:147`), header interpolation (`:155`) — never argv, never a file. Both failure branches (the generic `logger.exception` traceback and the `URLError`→`ToolError` string) were driven with a marked token value and the stdio replies and log output grepped: absent.

**The redirect refusal genuinely holds**, and not by assumption. The handler dump confirms `HTTPRedirectHandler` is gone and `_NoRedirect` is in its place, `redirect_request` returning `None` is the documented refusal contract, the chain then falls to `HTTPDefaultErrorHandler` and raises, and 301/302/303/307/308 all route through the same method. urllib honours no `Refresh` header at all, so there is nothing there to bypass. The branch's own socket-level test asserting the server was hit exactly once is green, and it exercises the real opener rather than a monkeypatched one.

**No tool argument can reshape a URL.** `_picture_id` rejects bools and non-int/str then `int()`s, and the result is interpolated as an `int`, so `/` and `?` cannot survive; `_paging` clamps and type-checks; everything else goes through `urlencode`.

**All six tools are fixed GETs on existing, already-declared routes** — `/pictures`, `/pictures/search` and `/tags` are `SCOPED_LIST`; `/pictures/{id}/metadata`, `/pictures/thumbnails/{id}.webp` and `/comfyui/pictures/{picture_id}/recipe` are `PICTURE_SCOPED`. `registry.py` is untouched, so the server inherits the gate rather than going around it. `GET /tags` additionally filters through `fetch_scope_allowed_picture_ids` in the handler. Query params check out: `tag` is repeatable on both listing routes, `sort=DATE`/`descending=true` are valid.

**The tests close the dead-route trap** for the five live-tested paths — every picture-tool negative sits beside an in-scope positive on the same path shape and an owner control on the same id. Module-scoped fixture with one `Server`, and nothing hand-writes state the background sweeps would overwrite. Protocol edges are covered in one pass through `serve`: notification, client response, -32601, -32602, -32600, -32700, and the session surviving all of it.

Gating is right — `tests/test_mcp_server.py` is in the `backend` job list in alphabetical position. The doc edit sits in the §4 module table outside both `AUTOGEN` regions, so `render_backend_architecture.py --check` stays green. Logging goes to stderr rather than stdout because `main()` never calls `setup_logging`; that is load-bearing for a stdio protocol and worth not breaking later. Test credentials carry the sanctioned `example-` marker and the only address literals are `127.0.0.1`.

### On the stated deferrals

`search_pictures` not scope-tested live: **accept**, with the route-resolution point above attached — its policy is the same `SCOPED_LIST` as `GET /pictures`, which *is* scope-tested here, so what is missing is route coverage, not authz coverage. `list_tags` scoping owned by the route: **accept** — the handler filters and returns `[]` on an empty scope, and a "fewer tags" assertion is not reliably constructible from two flat-colour PNGs. Full-access token accepted: **accept** the design, but see the README wording above. HTTPS self-signed unverified: **accept** — the default is plain HTTP on loopback where there is nothing to intercept, and the failure is a legible `URLError` rather than a silent fallback. `openWorldHint` on `get_recipe`: correct annotation.

The `workflow_hash.py` F821 is already fixed on `develop`; the branch is just behind, and a merge clears it.

lindkvis:
Addressed in e97ff83d (develop merged in 51d5081d, which clears the `workflow_hash.py` F821).

**Must fix — the proxy.** Fixed, and the fix is `build_opener(_NoRedirect, ProxyHandler({}))` as suggested. The regression test you sketched caught something the suggestion did not: with the opener built at module import, a test that sets `http_proxy` cannot reproduce the bug at all, because `ProxyHandler()` reads the environment when it is constructed. My first version of the test passed with the guard removed. The opener is now built inside `http_fetch`, so the environment is read at the point the owner's process actually runs, and removing `ProxyHandler({})` turns the test red. The test asserts the request landed on a local socket server with the Bearer header while four proxy vars point at `proxy.example.test`.

**README.** Rewritten to your sentence: an agent that can read its own config file has full owner control, not read access. It now also says not to give it one.

**Windows non-ASCII.** `sys.stdin.reconfigure(encoding="utf-8")` in `main()`.

**Route resolution.** Added, and it also found two ways to write that test wrong. `app.routes` does not contain the API routes at all (they sit behind `_IncludedRouter`), so the first version matched nothing and still passed; it now goes through `route_inventory.api_endpoint_set`, the same enumeration the authz guardrails use. Then the SPA catch-all `/{full_path:path}` matched every string, and after excluding it the unescaped `.` in `/api/v1/pictures/{id}.{ext}` still matched `/pictures/searchh`. With literal parts escaped, typos in `/pictures`, `/pictures/search` and `/tags` all go red.

One honest limit: `/pictures/{id}/metadataa` still resolves, because `GET /pictures/{id}/{field}` is a real route that takes any field name. The three picture tools are exercised against the live server anyway, so a rename there fails those tests.

**`id` with no method** now answers `-32600`.

**Nits.** `2025-03-26` removed from the advertised versions, since refusing batches while advertising the revision that mandates them was the untruth. `from exc` added on the `URLError` branch. `_tool` no longer leaks as a module global.

Not taken: `fields=grid` for `list_pictures`. The grid projection drops tags and description, which is most of what an agent asks a listing for, and the limit is already capped at 200. Worth revisiting with a real agent's context cost in front of us rather than guessing. Head-of-line blocking also left alone for now — one user, read-only, and a shorter timeout is the easier lever if a client complains.

PR #1371, opened 2026-09-15
Models: which support files a delete leaves behind

### LINKED ISSUE #1314: Models: which support files can be deleted
- Add `family`, `quant` and `weights_id` to models.
- Collect usage evidence in layers, including which models are used together in recipes.
- Before deleting a model, show what still uses it and what would be left orphaned.

Depends on: #1311

lindkvis:
Closes #1314.

## What changed
- **`family`, `quant`, `weights_id` on `model`**, read from the safetensors header at scan. `family` is the tensor layout of a support file (`vae_16ch`, `clip_l`, `t5_xxl`); the shelf serves the base model's family instead when `base_model` folds. `quant` is the dtype holding most parameters, or `mixed`. `weights_id` hashes tensor names and shapes without dtypes. Existing rows get them on the next scan: the fast path reads the header and does not re-hash the file.
- **`POST /models/companions`** (owner_only). It uses recipe co-occurrence from `workflow_recipe_asset`, across every library on the hub, to sort the VAEs and text encoders next to the models being deleted:
  - **orphaned**: every consumer a recipe records is being deleted.
  - **shared**: a kept model still uses it.
  - **unknown**: two shelf rows share its name, or a digest can't be matched while a hash is pending.
  - It also returns `in_use`, `no_evidence`, and `unrecorded` (kept base models no recipe names).
- **Delete prompt** on the shelf appends those answers as sentences. It still offers the delete if the read fails, and it ignores a second Delete press while the prompt is being prepared.

## Deliberate choices
- **"Orphaned" is worded as what the evidence shows:** "ran only with models you are deleting", not "nothing uses it". A kept model with no recorded workflow may still need the file, so the prompt says how many there are. A pre-push review caught the earlier "nothing else uses" wording.
- **Nothing is deleted or pre-selected.** Orphaned files are named, and the owner deletes them separately.
- **The header facts don't feed the delete answer yet.** A checkpoint's `flux1` and a `vae_16ch` are different vocabularies, and joining them needs a compatibility table this PR doesn't invent. The review flagged them as unused scope; they stay because the issue asks for them.
- **Not included:** evidence from ComfyUI's own history and saved workflows, a "delete these too" checkbox, and shelf columns for the new fields.


lindkvis:
## Review: approve with nits — two one-line fixes wanted first

Backend, frontend and security review, merged. Nothing here is architecturally wrong, and the reasoning in `fetch_companions` is careful, the classification is total, and the tests are good — the digest-pending one in particular proves its own branch is load-bearing. Two one-liners and a paragraph-formatting fix, then it is fine.

### Must fix

**A crafted safetensors header crashes the whole model scan.** `pixlstash/utils/adapter_header.py:420` does `count > total * _QUANT_MAJORITY`. `total` is a Python int summed from shape products taken from the file's own JSON header, and `total * 0.5` converts it to a float — a big enough int raises `OverflowError`. Reproduced with a 1,341-byte header, well under the 16 MiB cap: `count_parameters`, `weights_id_from_header` and `family_from_header` all survive it; `quant_from_header` raises.

Nothing catches it. `describe_adapter` has no guard and `ModelFolderScanner` calls it at `model_folder_scanner.py:522` and `:529` with no `except Exception` anywhere in the file. So one crafted or merely corrupt `.safetensors` aborts the folder scan, leaves every model after it in the walk unregistered, and does it again on every rescan. That is not disclosure or RCE — it is a persistent, silent denial of the model shelf, planted by a file downloaded from a model site.

The "no effort protecting the owner from their own plugins" stance does not cover this one: `read_safetensors_header`'s own docstring already treats the header as hostile ("either a corrupt file or one built to make us allocate"), and `count_parameters` was carefully all-integer. This is the first thing in the module to break that posture. Fix is `count * 2 > total` — equivalent and unoverflowable — plus a regression test watched failing first.

**The companion sentences arrive as one block.** `ModelShelf.vue:1713` concatenates `companionsSentences` onto the existing `message`, and `ConfirmDialog.vue` renders `message` as one `<p class="confirm-dialog__message">`. With all five clauses populated that is roughly ninety words in a single unbroken block, with the `warning` box below competing for the same attention. The whole argument for this feature is that the answer is in front of the reader before they agree; a wall of prose is how it stops being. Smallest fix: return the `said` array `companionsSentences` already builds and render one `<p>` per sentence, or keep the string, join with `\n\n` and add `white-space: pre-line`.

### Backend — what I went looking for and did not find

The safetensors header parse is otherwise sound and not new here — `read_safetensors_header` (`adapter_header.py:221`) reads the 8-byte prefix, refuses a zero or >16 MB declared length *before* allocating, detects a short read, catches JSON and non-dict decodes, and never touches the tensor payload, logging path and numbers on every exit. CPython's own 4,300-digit integer-literal limit also kills the degenerate "one enormous number" header at `json.loads`. The three new readers go through `_valid_shape` (`:363`), which rejects a non-dict entry, a non-list shape and a bool-or-negative dimension, so a header crafted with `{"shape": true}` or `{"shape": [-1]}` yields `None` rather than an exception.

`quant_from_header` (`:390`) is one pass over a header already in memory, weighted by parameter count rather than tensor count — the right call, and the 300-f32-norms-vs-200-f8-weights test genuinely goes red if someone counts tensors instead. `weights_id_from_header` (`:423`) is not collision-prone in any way that matters: a collision requires an identical sorted set of `(name, shape)` pairs, which is the definition of the same architecture at the same size. The failure mode is the opposite one — a repack splitting a set that is morally one model — and the docstring says so.

The hub schema change follows the file's own pattern: three nullable `TEXT` columns in the `model` DDL plus a guarded `ALTER TABLE … ADD COLUMN` loop in `_apply_v2`, right below the `icon_sha256` block that established it. `_apply_v2` re-applies on every open, so existing hubs pick the columns up without a version bump. The backfill claim holds too — `has_header_facts` derives from `weights_id IS NOT NULL` (`model_folder_scanner.py:617`), the fast path re-reads only the header (`:522`), and the scanner test monkeypatches `sha256_file` to raise, so a re-hash would fail the test rather than merely go uncounted.

No N+1 — five full table scans and then pure dict work, nothing queried per row. The classification is total, and the one branch that reports nothing (a support file whose consumers include no deleted model, `model_shelf_service.py:638`) is both deliberate and covered: `test_a_deleted_support_file_names_the_kept_models_that_use_it` asserts `orphaned == [] and unknown == []`, which is that branch firing twice. New SQL in `hub/workflows.py` is fully parameterised, batched `IN` lists included.

### Authz — clear, and provably complete

`OWNER_ONLY` is right. `workflow_recipe` and `workflow_recipe_asset` carry no `library_id` at all — they are content-addressed hub tables — so "across every library" is not a widening, there is simply nothing to scope. There is no per-object id to scope on either, and the route returns model names and sizes that `owner_only` `GET /adapters` already serves alongside `folder_path`. It is the strictest non-locality tier, so scoped tokens are refused outright and there is no partial-leak question; `LOCAL_OWNER_ONLY` is for §16.3 host-filesystem authority and this route takes no path, touches no file and writes nothing. As a POST it is also blocked pre-routing for non-`WRITE` scopes and is not in `READ_SAFE_POST_PATHS`.

The gate really does enforce it on this method+path shape: both guardrails pass on the branch tree, and deleting the registry line turns `test_coverage_matrix_document_matches_the_registry` red and makes `test_all_routes_declare_access_policy` error on the boot abort — so `effective_path` matches what the router mounts. Registry and matrix agree in both directions: identical entry counts and identical `owner_only` counts. No `POST /models/*` route shadows `/models/companions`.

One thing worth knowing rather than fixing: the route does disclose something new to an owner-class caller — that *some* library paired model X with model Y in a recipe. The shelf already discloses machine-wide that both exist, and the caller is by construction an unscoped owner, so this is fine today. But it is a co-occurrence oracle if the library pin is ever meant as a confidentiality boundary.

### Backend — nits

**`unrecorded` counts unclassified files and calls them base models.** `model_shelf_service.py:612` counts every kept row whose `file_kind` is not in `_NOT_CONSUMERS` — which is `checkpoint` *and* `unknown`. Including `unknown` is right for orphan detection (a misclassified base model must keep its VAE), but reusing the same set for a number the prompt quotes verbatim — "N other base models have no workflow on record" — means a shelf with a fat Unclassified bucket reports a count of files PixlStash explicitly could not identify. The count also has no join to `model_file`, so rows whose copies are all absent are in it. Both errors point the safe way, hence a nit: either restrict to `checkpoint` or reword to "other models".

**`in_use` is the only unsorted bucket.** `orphaned`, `shared` and `unknown` come from `sorted(candidates, key=name)` at `:636`; `in_use` is appended inside the set loop at `:632`. Sort it so the prompt does not reorder itself between shelves.

**A file that can never yield a `weights_id` is re-opened on every scan.** `has_header_facts` flips only once `weights_id` is written, so a registered `.safetensors` whose header later becomes unreadable, or one with no tensors, takes the `describe_adapter` path at `model_folder_scanner.py:522` forever and logs a warning each time. Rare and cheap, but it is an unbounded retry with a log line attached.

**`family` can carry arbitrary header text.** `adapter_header.py:478` returns `f"vae_{conv_in[1]}ch"`, letting a header put up to ~4,300 characters into `model.family`, which reaches the shelf UI. Not an injection — it is escaped on the way out — but the column is documented as a closed vocabulary, and the CLIP and T5 branches right beside it already do the right thing by looking the width up in a dict and returning `None` on a miss.

### Frontend — nits

**Pressing Delete does nothing visible for a round-trip.** `ModelShelf.vue:1706` awaits `fetchModelCompanions` *before* the prompt opens, and `deletePromptOpen` (`:1647`) silently swallows every press until it settles. Milliseconds on a healthy hub; on a stalled read the user presses Delete, sees nothing, presses again, still nothing. The guard is right — it must not open a second prompt over the first — but it needs a busy/disabled state so the silence is legible.

**The failure default is correct and I would keep it.** Offering the delete when the read fails is right, and the user *is* told: `companionsSentences(null, …)` returns "PixlStash could not check…" (`utils/modelShelf.js:1273`) rather than quietly omitting the section, which is the trap a lesser version would fall into. Both directions tested.

**`nameList([])` would emit " and undefined".** `utils/modelShelf.js:1247`: `names.length <= 3` is true for zero and `=== 1` is false, so the join branch runs on an empty array. Every call site is behind an `if (x.length)` so it is unreachable today; one `if (!names.length) return "";` keeps it that way.

Singular/plural reads well otherwise — two names give "a and b", three "a, b and c", four-plus "a, b, c and 2 more", and is/are, it/them, "it stays"/"they stay" are all correct.

**Design tokens: nothing to check.** The `ModelShelf.vue` diff is script-only, the two JS files have no markup, and no `<style>` block is touched anywhere in the PR. If you take the must-fix, the per-sentence version should use the existing `--space-*` stack on the container rather than margins on the paragraphs.

### On the unused scope

`family`, `quant` and `weights_id` are correct, cheap, well tested, carried through `CheckpointHashTask._merge` — and read by nothing that decides anything. `family` also deliberately mixes two vocabularies in one column (`family_of(base_model)` returns `sdxl`, `family_from_header` returns `vae_16ch`, and the route serves whichever exists), so every future consumer has to disambiguate a column by inspecting its value.

My recommendation is to cut them to a follow-up: three permanent hub columns and a forced header re-read of every registered file on every existing shelf, in service of a question this PR explicitly cannot answer yet. If closing #1314 whole matters more, shipping them is not harmful — just note you are paying the migration now for a join you have not designed.

### Tests

Identity, not counts, on both sides: exact `companion_ids(...)` lists, exact `used_with` dicts, exact merge tuples. `test_companions_answers_the_owner_and_refuses_a_share_token` asserts owner 200, character-token 403 and unauthenticated 401, and keeps the in-scope positive control next to the negative, so the 403 cannot be passing because the credential was missing — it proves the route resolves before trusting the refusal.

`test_a_digest_the_shelf_cannot_match_yet_makes_its_companions_unknown` is the best test here: it asserts `unknown`, hashes every row, then asserts the same call now returns `orphaned`. That flip is the only thing that could prove `digests_are_complete` (`model_shelf_service.py:545`) is load-bearing rather than decorative.

One fragile assertion: in "opens one prompt however often Delete is pressed while it is being prepared", the third `pressDelete()` is dispatched synchronously right after `answer(...)`, so the `await` continuation has not run and `deletePromptOpen` is still `true`. It holds deterministically, but for a microtask-ordering reason rather than because a press *after* the prompt settles should be ignored. Insert an `await` before that third press and it goes red for correct behaviour — or at least leave a comment saying why it is written that way.

`ruff check` and `ruff format --check` clean on all eight touched backend files.

PR #1372, opened 2026-09-15
Recipes: read A1111 generation data

### LINKED ISSUE #1312: Recipes: read A1111 generation data
Pictures from A1111 and its forks carry parameters, not a graph. Map them onto the same recipe identity.

- Detect the format.
- Classify each field as structure, asset, parameter or volatile.
- Short model hashes are resolved when unambiguous and marked uncertain otherwise.

Depends on: #1311

lindkvis:
Closes #1312.

A1111/Forge PNGs carry a `parameters` text, not a graph. `services/a1111_recipe.py` parses it and builds the graph it describes as a reduction. `hub/workflows.record_reduction` then files it into the same topology, recipe and instance tables as a ComfyUI graph, and the seed goes into `generation`.

- **Fields** are classified by name. Structure is a node: LoRA, embedding, hires fix, refiner, img2img source. Assets are model names, fields named for a model, and checkpoint/embedding hashes. Volatile is seeds, version and hash summaries. Everything else is a parameter, unknown fields included. Parameters are not named in the recipe, because A1111 leaves out fields that hold their default. LoRA tags are cut from every value, and credential-named fields are dropped.
- **Short hashes** are stored as written and resolved when read (`digests_with_prefix`). If exactly one shelf digest starts with the hash, it counts as verified. Otherwise the picture falls to the unverified by-filename tier. Resolving at scan time would give one model's pictures different recipes depending on when the model reached the shelf. Model ghosts judge short hashes by the same prefix rule.
- **Migration 0119** clears the scanned marker on PNGs that have no workflow keys, so existing A1111 pictures get read once.

Known limits:
- A1111 names have no extension, so `.safetensors` is assumed. A `.ckpt` model then reads as a ghost, and it can match a same-stem shelf file in the unverified tier.
- `Lora hashes` is dropped: for `.safetensors` LoRAs it is not a file sha256 prefix.
- A model named inside a compound value (ControlNet) still reaches the instance document.
- JPEG/WebP are not read (their data is in an EXIF sub-IFD).
- 0119 re-opens every PNG without a workflow once, so scan progress dips after upgrade.


lindkvis:
## Review: approve with nits

The design is right and the hard parts — short-hash resolution, the four-bucket classification, the migration — are done properly. The one thing I would fix first is a one-line change in `parse_infotext`, so a mis-split compound value cannot overwrite `Model` and put the wrong checkpoint into `workflow_recipe_asset`.

Every input string below is one I actually ran against the branch's parser.

### Must fix

**A duplicate key overwrites the checkpoint — `a1111_recipe.py:150-156`.** `_PARAM_RE.findall` is fed to `fields[key] = ...`, so the **last** occurrence of a key wins. An unquoted compound value containing a comma gets shredded into fake pairs, and if one of them is `Model:` it silently replaces the real checkpoint:

```
a castle
Negative prompt: bad
Steps: 20, Sampler: Euler, CFG scale: 7, Seed: 1, Size: 512x512, Model hash: aabbccddee, Model: my_real_checkpoint, ControlNet 0: Module: canny, Model: control_v11p_sd15_canny.safetensors, Weight: 1, Version: v1.10.1
```

gives

```
assets: [('ckpt_name', 'control_v11p_sd15_canny.safetensors'), ('ckpt_sha256', 'aabbccddee')]
```

`my_real_checkpoint` is gone, and the recipe now names a checkpoint whose filename and digest belong to two different models. On the shelf that credits the ControlNet model in the by-filename tier and gives the real checkpoint nothing.

In fairness `_PARAM_RE` is character-for-character A1111's own `re_param`, and A1111 quotes any value containing `,`, `:` or `\n` before writing it, so this only bites on infotext written by something that bypasses `quote()`. But A1111 only repopulates a UI from its parse; here the result is a durable asset row. A1111's field order always puts `Model` before `extra_generation_params`, so first-wins is strictly safer and costs one line — `fields.setdefault(key, ...)` — plus a test with the string above.

### Nits

**An empty model name manufactures a `.safetensors` ghost — `a1111_recipe.py:335`.** `Refiner: [abcdef0123]` (a hash with no name) parses to name `""`, and `_asset_name` appends the extension unconditionally, giving asset `('ckpt_name', '.safetensors')`. That ends with `SHELF_MODEL_SUFFIX`, so it lands in `model_ghost_names` and shows in the forget-a-model UI as a ghost called `.safetensors`. Guard the empty string.

**`parse_infotext` reads only the last line — `a1111_recipe.py:140-142`.** Anything appended after the fields line silently drops the whole picture, and because the marker still goes down it never gets another look until `HASH_VERSION` moves. Adding a `Template: something else` line after a valid infotext makes `parse_infotext` return `None`. A1111 itself never does this (it escapes newlines), so the exposure is re-saves by other tools. The tolerant version is still one line: walk `lines` backwards for the last one whose parse yields a `Steps` field.

**Embedding nodes exist only when `TI hashes` was written — `a1111_recipe.py:224-230`.** The same generation produces two different **topologies** depending on whether the user had hash summaries enabled — measured, the two variants do not share a topology hash. Inherent to the format and I would scope it out, but it belongs in the known-limits list next to the `Lora hashes` note.

**The credential drop over-matches, and silently merges instances — `a1111_recipe.py:188-195`.** `SECRET_FIELD_RE` is unanchored, so `Token merging ratio` is dropped from every picture that has it, as the comment says. The consequence worth stating: the field is then absent from `instance_widgets`, so two pictures at token-merging 0.2 and 0.5 collide on the instance hash. That is the recoverable direction so I would leave it, but it is a behaviour rather than just a log line. The pattern is effective on the cases that matter — `Example API key`, `Civitai token`, `… secret` all match.

**Parser cost at the ceiling — `a1111_recipe.py:85`.** The quadratic is real and the cap bounds it, but loosely: a 19,000-char fields line takes 0.82 s, 10,000 takes 0.22 s, a realistic ADetailer-heavy ~3,000 takes ~0.02 s. Since 0119 re-opens every PNG, I would drop `_MAX_FIELDS_LINE` to ~8,000.

**The skip log has no picture context — `a1111_recipe.py:145-147`.** `reduce_a1111` only sees metadata, so the length-skip line cannot name the picture. Log at the caller, or accept it.

**Smaller observations, no action needed.** An unquoted value truncates at the first comma (`ADetailer prompt: a face, detailed` keeps only `a face`) — parameter-only and matches upstream. Any PNG whose `parameters` last line contains `Steps:` is filed as a recipe (`"Steps: first do this, then that"` yields a four-node graph with an empty prompt) — harmless, since only A1111 and Fooocus write that chunk and Fooocus is excluded. `Hires upscaler: 4x-UltraSharp` is recorded as `4x-ultrasharp` with no extension so it can never match a shelf row — deliberate, but those asset rows are write-only. Swapping positive and negative prompts does *not* collide on the instance hash: topology and recipe equal, instance differs.

### Checked and sound

**Short-hash resolution (`workflow_hash.py:298-315`).** `bisect_left` plus a two-element window is the correct way to answer "one or more than one", since all digests sharing a prefix are contiguous. `DIGEST_PREFIX_RE` accepting only 10/12/64 is right — 8 hex is dropped, 12 and 64 kept. Collision math is comfortable: 10 hex is 40 bits, so a 10,000-model shelf has roughly 5e-5 chance of any prefix pair, and a collision degrades to the filename tier rather than mis-attributing.

**Ghosts and shelf counts use the same rule** — verified rather than taken on the PR body's word. `hub/workflows.py:322-329` treats *any* match as "not a ghost"; `model_shelf_service.py:454-458` requires exactly one for verified. That asymmetry is correct, not a bug: two matches mean the shelf does hold a model with that prefix (not a ghost) but we cannot say which (not verified), so the picture falls to the filename tier. The `judge_digests` gate still holds off while any model has a NULL `sha256`, and resolve-at-read is the right choice. One widening worth noting without objecting: a 10/12-hex value in *any* ComfyUI `*_sha256` widget is now ghost-judgeable where before only 64-hex was.

**Migration 0119** meets every rule here: descriptive name, `__all__` after `depends_on`, `down_revision = 0118_add_generation_tables` which is the current head on `develop` so strictly increasing and single-headed, no `op.add_column`, and nothing but a targeted marker reset — the documented "reset the column, let the finder pick it up" pattern. The `downgrade` `pass` is a deliberate no-op with a reason, not swallowed error handling. Cost is bounded: one full `picture` scan at migrate time, then one re-read per PNG, after which the stamp goes back down. The progress dip is the existing task machinery and 0118 set the precedent.

**Integration with `record_reduction`.** `record_api_graph` now delegates, so there is exactly one path to the document, instance document and keys. The graph shape fits the existing tables — `_node` produces asset-only `widgets` and `instance_document_from_reduction` fills parameters in for anything not in `widgets`, as it already does for ComfyUI. No meaningful collision risk between sources: every class type is `A1111*` and `graph_key` hashes class types.

**Error handling.** No bare `pass`. `reduce_a1111` sits inside the existing `try` in `_record`, so a parser crash lands in `_stand_down` (unmarked, retried) rather than being mistaken for "no workflow" — the right side of the rule the docstring states.

### On the stated known limits

**Accept as scoped:** `.safetensors` assumed so a `.ckpt` reads as a ghost — honest rather than wrong, since the shelf only holds `.safetensors`. `Lora hashes` dropped — correct call, AutoV3 hashes tensor data rather than the file; worth one user-facing line that LoRAs therefore never reach the verified tier. JPEG/WebP unread — fine here, but flag for whoever picks it up that 0119 is PNG-only, so EXIF support needs its own migration.

**Accept, but open a follow-up issue:** the ControlNet compound value leaking a model name into the instance document. It is the only limit that makes an existing guarantee partly untrue rather than merely incomplete — `document_from_reduction` exists so that forgetting a model name is a row delete with no stored graph rewritten, and `_nested_assets_as_references` only swaps strings carrying a model extension, so `Model: control_x [hash]` survives verbatim. Bounded by the instance row's lifetime, so not a blocker, but it should not stay an undocumented footnote.

`ruff check` and `ruff format --check` clean on all eight touched Python files.

lindkvis:
Merged `develop` and addressed the review. All in `11d9e67`.

**Must fix — duplicate key overwrites the checkpoint.** Fixed: `_parse_fields` now uses `fields.setdefault`, so the first occurrence wins. Your exact ControlNet string is the test (`test_a_shredded_compound_value_never_replaces_the_checkpoint`); it fails with `fields[key] =` restored.

**Empty model name.** Fixed: `_asset_name` returns `None` for an empty name instead of `.safetensors`, and the call sites drop it. `Refiner: [abcdef0123]` now files the hash and no name.

**Only the last line is read.** Fixed: `parse_infotext` walks backwards for the last line that parses. A `Template:` line appended after a valid infotext keys identically now.

**Parser cost.** `_MAX_FIELDS_LINE` is 8,000, with your measurements in the comment.

**Embedding topology depends on `TI hashes`.** Scoped out as you suggest, and now in the known limits in `backend_architecture.md`.

**Credential drop merges instances.** Left as is, and the consequence is now stated in the comment rather than only the log line.

**ControlNet compound value.** Follow-up issue opened: #1375. Also named in the known limits.

**Skip log has no picture context.** Left as is: `reduce_a1111` only sees metadata, and threading a picture id through for one debug line costs more than it buys.

**JPEG and WebP** are read in this PR after all, so no separate migration is needed: `ImageUtils._user_comment` pulls `UserComment` out of the Exif sub-IFD and lets `piexif` decode it, and 0119 now covers `.png`, `.jpg`, `.jpeg` and `.webp` (renamed to `0119_rescan_pictures_for_a1111_recipes`). A JPEG, a WebP and a PNG carrying the same text land on the same instance hash, which is the test.

**Short-hash resolution is now shared with `fetch_companions`.** `develop` grew `_recipe_asset_index` and a second digest reader while this was open, and it did an exact lookup. Both callers go through `_models_for_digest` now: one match is the model, several are ambiguous, which `fetch_companions` already handles the way it handles a shared filename. The digest list is sorted once per call, not per asset row.

**LoRAs never reaching the verified tier** is now a changelog line.

The merge commit also carries the above, rather than being a pure resolution — worth knowing when reading the diff.

PR #1373, opened 2026-09-15
Stacks: Keep recipes only

### LINKED ISSUE #1315: Stacks: reclaim the space held by near-duplicates
Offer to delete the pixels of the non-cover pictures in a stack, keeping a thumbnail and what makes each one different.

- Only pictures whose recipe can reproduce them are eligible. Say honestly which are not.
- A consent screen before anything is deleted, which can also turn on keeping ghosts.
- One undo for the whole batch.

**Open:** what the consent screen leads with. `docs/design/keep-cover-only.md` rules out a headline reclaim figure.

Depends on: #1311, #1309

lindkvis:
Closes #1315.

Adds **Keep recipes only** next to Keep cover only. It is a mode of the same planner and routes (`keep_recipes` on `/stacks/keep-cover-only{,/preview}`), so it keeps the same rules: one dry run, disjoint buckets, soft delete to the Scrapheap, one undo.

**What moves.** A non-cover copy moves only if the purge could leave a ghost of it. That means:
- a recipe instance in the hub for this library;
- a `pixel_sha`;
- no model ghost or forgotten model name in its recipe;
- a thumbnail file;
- a ghost the retention setting keeps. Under `covered`, some picture that stays must share its instance hash.

Every other copy stays live in its stack and is counted under one reason. A stack with nothing to move is a new bucket, `nothing_reproducible`. Copies that stay are left out of the metadata union, and character loss is judged over the copies that leave.

**Consent.** The dialog lists why pictures stay. When some stay only because their ghost would not be kept, it offers "Keep a ghost of every picture deleted from the Scrapheap". Ticking the box re-runs the preview under `on`. The real call sets `workflow_ghost_retention` to `on` only if it moved something, using the same helper as the Privacy PATCH, which now calls it too. If saving fails, the response returns `ghost_retention_saved: false` instead of a 500.

**Deliberate choices a reviewer may question**
- **The headline is the picture count, not a reclaim figure**, following `keep-cover-only.md`. The issue left this open. Changing it is copy only.
- **The retention setting is not part of the undo.** It is a privacy setting that applies to every library, and the checkbox note says so.
- **The pixels still leave only through the Scrapheap.** Whether a ghost survives is decided when the Scrapheap is purged. A later cover delete or setting change can still remove it, which is how the setting already works.
- **Known gaps:**
  - Like model ghosts, only `.safetensors` names and digests are judged, so a missing `.ckpt` is not caught.
  - The hub reads, including the forgotten-name `json_each` scan, run inside the vault DB task. This is marked `ponytail:`.
  - A hubless vault reports every copy as "no workflow recorded".


lindkvis:
## Review: changes requested — two items, neither of which loses data outright

Backend, frontend and security review, merged.

**Fix first:** the move predicate promises "could be made again" but only judges `.safetensors`. A copy whose recipe loads a `.ckpt`, `.gguf` or `.pt` checkpoint that is nowhere on the shelf passes the model check, moves, and ends up as a ghost nobody can re-render. The PR body lists this as a known gap; on a destructive action with that wording it needs either the extra predicate or a change of wording, not a footnote.

I traced the whole predicate adversarially against what the purge actually does (`workflow_ghost_service.apply_purge_to_hub` / `_prepare_ghosts`) and **could not construct a state where a picture is destroyed with nothing left behind**, other than the `.ckpt` case. Everything else fails in the conservative direction. The pixels still only leave through the Scrapheap, undo is one `Ctrl+Z`, and the default retention (`None`) means the Scrapheap never empties on its own, so every residual risk below needs a deliberate purge before it bites.

### Backend — must fix

**1. The model check is blind to every model format that is not `.safetensors`.** `recipes_missing_a_model` (`pixlstash/hub/workflows.py:439`) builds its "missing" set from `model_ghost_names`, and that only ever judges a name ending in `SHELF_MODEL_SUFFIX`, the literal `".safetensors"` (`pixlstash/services/model_folder_scanner.py:73`; the docstring at `hub/workflows.py:288` says so — "it scans `.safetensors` alone"). The rows are in `workflow_recipe_asset` either way, they are just never compared against the shelf.

So: a copy with an instance row, a `pixel_sha`, a thumbnail and a cover sharing its hash, whose `CheckpointLoaderSimple` names a `.ckpt` the user does not have, is classified as movable. It goes to the Scrapheap, the purge writes its ghost, the original is deleted from disk, and the recipe it left behind cannot be run. That is the one concrete state where the action's own promise fails.

For the model-ghost *screen* that blind spot is defensible — calling a `.ckpt` a ghost would forget the name of a model still on disk. For *deciding what to delete* the same conservatism points the other way: an unjudgeable model name should count as `model_missing`, not as fine. The cheap version is one predicate in `recipes_missing_a_model` — flag any structural hash with a loader asset whose name has a model-ish extension that is not `.safetensors`. If you would rather not, the dialog and `changelog.d/1315.md` should stop saying "could be made again" and say what is actually checked.

Same shape, narrower: `_model_ghost_names` suspends digest judgement entirely while any shelf model still has a NULL `sha256` (`judge_digests`). On a library where `MissingCheckpointHashFinder` has not finished, a recipe identified only by a download node's `expected_sha256` is never flagged. Worth a docstring sentence at minimum.

**2. Preview and confirm are two independent plans, and this mode's inputs move on a background timer.** `preview_keep_cover_only` and `post_keep_cover_only` each build their own `RecipeCheck` and call `plan_in_session` fresh. For Keep cover only that was fine — locked sets, character links and member counts do not change in the seconds a user spends reading a dialog. Keep recipes only added `no_thumbnail` as an eligibility condition, and thumbnails are generated lazily by `MissingThumbnailFinder` while the dialog is open.

Concretely: preview says 3, the user reads the staying-reason rows for twenty seconds, five more thumbnails land, they press **Move 3 to the Scrapheap**, and 8 move. Same if a model is added to the shelf, or an import lands a picture that covers a previously-uncovered hash. It is recoverable — the receipt states 8 and offers Undo — but `docs/design/keep-cover-only.md` acceptance criterion 1 exists precisely so the button's number is the number acted on, and this is the first input that drifts on its own.

Smallest honest fix: send the previewed moving ids (or just the count) on the real call and refuse with a 409 when the plan has grown. Alternatively accept it and write the exception into the design doc — but it should be a decision, not a silent property.

### Backend — should fix

**3. `pictures_staying_*` undercounts, and the comment justifying it is wrong.** `keep_cover_only_service.py:955-958` sums the staying buckets over `[*eligible, *skipped_nothing]` only, with the comment "the only ones whose copies were judged one by one". Not true: `plan_in_session` judges every stack with ≥2 members and no lock, including ones later skipped for `character_on_copy`, and `_plan_one_stack` does populate `staying_copies` on that branch. A stack's `staying_picture_ids` row can list copies the headline "N pictures stay in their stacks" never counts. The scoping decision is defensible; the comment is not.

**4. The retention write happens after the pixels have moved.** `pixlstash/routes/stacks.py` sets `workflow_ghost_retention: on` only once `pictures_moved` is truthy, but the plan was made *under* `on`. If `persist_server_config` raises, the response degrades to `ghost_retention_saved: false` and the setting reverts to `covered` on restart — at which point a purge drops exactly the ghosts of the uncovered copies this run moved *because* the user ticked the box. The degrade is honest and the in-memory value stays `on` for this process, so it needs a failed save *and* a restart *and* a purge. But the ordering is backwards for a destructive action: write the setting first (trivially reversible), then move. The "only if something moved" rule survives by writing it between the plan and the delete.

**5. The `ponytail:` comment names the right ceiling but the wrong blast radius.** `keep_cover_only_service.py:600`. Both hub reads are whole-hub and unscoped no matter how small the selection: `model_ghost_names` does a `SELECT DISTINCT` over all of `workflow_recipe_asset` plus four model queries, and `forgotten_asset_counts(hub)` with no `topology_hash` does the `json_each` pass over every recipe document. On top of that `_staying_reasons_in_session` calls `ImageUtils.find_thumbnail` once per copy — and that call has a side effect: it `os.replace`s legacy thumbnails, so a "read-only" preview writes to disk.

The part the comment misses is that this runs inside the serialised vault DB task, so it is not just this request that gets slow — it holds the worker and every background finder queues behind it. And the dialog re-runs the preview on every tick of the keep-every-ghost box.

### Backend — checked and correct

Right library: `filed_instance_hashes` filters on `library_uuid = ?`, taken from `vault.library_uuid`. The planner is *stricter* than the purge (which only needs a non-null `instance_hash` + `pixel_sha` + a thumbnail on disk) — the safe direction — and both sides use the same `ImageUtils.find_thumbnail`, so the thumbnail answer cannot diverge. `covered` is judged against live pictures (`Picture.deleted.is_(False)`, a real indexed column) excluding the whole moving set, which is what `GhostPurgeContext.surviving_instance_hashes` measures after its DELETE; copies in a stack later skipped for character loss are still excluded from the covering set, which is conservative. `off` leaves `covered` empty, and a hubless vault or missing `library_uuid` returns `no_recipe` for every copy — both fail closed. No `pixel_sha` correctly folds into `no_recipe`. Buckets are disjoint, the sum assertion in `preview_in_session` was updated to include the fifth, and per-copy reasons are first-match over `STAYING_REASONS`.

The union narrowing is the subtle part and it is right: `apply_metadata_union_in_session` writes every tag and the best score onto *every* picture handed to it, so passing a staying copy would have written the cover's metadata onto a non-cover. `[cover_picture_id, *copy_ids]` fixes that, and `_character_loss` is judged over the same set. `member_ids` still covers the whole stack for the op-log capture — a harmless superset.

Residual risk, correctly documented: under `covered`, a copy moves because the cover shares its hash; if the user later deletes the cover and purges, the cascade destroys the copy's ghost too. That is the setting's existing contract, not something this PR introduces.

`ruff check` and `ruff format --check` clean on all seven touched Python files.

### Security — clear to merge

The interesting question, a data route writing a global privacy setting, comes back as an acceptable consented action rather than an escalation.

**The refactor did not weaken the PATCH.** `apply_ghost_retention` (`workflow_ghost_service.py:209`) is a byte-for-byte lift of the five lines it replaced. The `GHOST_RETENTION_CHOICES` validation still sits above the call at `config.py:1202` — it was never inside the moved lines. Scope is unaffected by construction: `PATCH /server-config/ghost-retention` is `owner_only` and the gate runs before the handler body, so moving application code out of a handler cannot touch authorization. There was no audit write in those lines, so none was lost.

**No privilege is gained.** `POST /stacks/keep-cover-only` is `owner_only` and so is the PATCH — same tier, same principal. Whoever can reach the stacks route can already set the setting by hand, so this is convenience, not a boundary crossing. The gating is as narrow as claimed: `keep_recipes` must be on (so `keep_every_ghost` alone does nothing), the box must be ticked, `pictures_moved` must be truthy, the setting must not already be `on`, and the value written is the module constant, never anything from the body. The preview path writes nothing — `recipe_check` is a pure dataclass constructor. The failed-save handling is right and worth noting approvingly: the pictures have already moved, so a 500 would read as "nothing happened"; returning `ghost_retention_saved: false` and logging with the moved count is the honest answer and swallows nothing.

Two nits. The helper is now a shared write chokepoint but validates nothing itself — both current callers are safe, but a one-line membership check would make it safe by construction. And the coverage-matrix row (`docs/authz-coverage-matrix.md:342`) describes the soft-delete and metadata union but not that the route can now write a global privacy setting; CI checks the matrix's arithmetic rather than its prose, so nothing goes red, but a reader auditing "which routes write server-config" would miss this one. One clause fixes it.

Same honest caveat as the PATCH it shares the helper with: an `ALL` token pinned to library A changes a setting governing every library. This PR does not widen that.

### Frontend — should fix

**6. The two danger items are near-indistinguishable.** `SelectionMenu.vue` and `ImageGridContextMenu.vue` put "Keep recipes only (3 stacks)" directly under "Keep cover only (3 stacks)", same `.ctx-item--danger`, same count, and the **same glyph** — both render `KEEP_COVER_ONLY_ICON` (`mdi-layers-minus`). One word apart, in a danger group, where one moves several times as many pictures. The tooltip distinguishes them; the visible row does not. `docs/design/keep-cover-only.md` picked that glyph deliberately as the inverse of `mdi-layers-plus`; a sibling wearing it identically undoes the point. Wants a distinct glyph and a `lead-designer`/`ui-ux-expert` call. Also, the pair now breaks the documented escalating-severity order, since Keep recipes only is the *less* destructive of the two and sits second.

**7. The checkbox disappears mid-flight when you untick it.** `KeepCoverOnlyDialog.vue:144`: `showEveryGhost` is `keepRecipes && (keepEveryGhost || (!figuresUnknown && …))`. Unticking sets `keepEveryGhost` false and `ImageGrid.vue:3959` immediately nulls the preview for the re-run, so `figuresUnknown` is true and both halves are false — the checkbox and its note unmount, then reappear when the preview lands. A control vanishing under the cursor in a destructive confirm is the wrong moment for a layout jump. Keep it mounted and disabled.

**8. Two small accessibility slips.** `KeepCoverOnlyDialog.vue:299-300` gives the section `role="status"` with `aria-labelledby="keep-recipes-staying-title"`, but that id's element is `v-if="stayingReasons.length"` — when the box shows with no staying reasons the reference dangles. Separately, an interactive `v-checkbox` inside a `role="status"` live region gets re-announced on every preview update. Move the checkbox and note out of the live region, next to it.

### Frontend — nits

- The note (`KeepCoverOnlyDialog.vue:326`) correctly says the box sets a Settings › Privacy toggle "for every library", but does not say `Ctrl+Z` will not turn it back off — the property the design doc calls out as most likely to be assumed wrongly. One clause.
- The lede's "each leaves its thumbnail and recipe behind for as long as your ghost setting keeps them" is true, but under `covered` the thing that takes the ghost away is deleting the *cover* — the picture they kept on purpose. A reader will not derive that from "your ghost setting".
- The `covered` staying row — "N have a prompt no kept picture shares, and your ghost setting only keeps recipes that one does" — is accurate and hard to parse. One more pass.

### Frontend — correct

**No stale-plan race.** `loadKeepCoverOnlyPreview` (`ImageGrid.vue:3965`) nulls the preview, sets `loading`, and increments `keepCoverOnlyRunToken` before the request; both success and failure paths bail on token mismatch. The body reads `keepEveryGhost.value` synchronously before the await, so a rapid toggle produces the last-ticked state in the last-issued request and out-of-order responses are dropped. `figuresUnknown` disables confirm for the whole in-flight window. Double-click is held by `keepCoverOnlyBusy`, and `setKeepEveryGhost` refuses while busy.

**Every figure comes from the server, none by subtraction.** `keepRecipesOnlyStayingCount` is the plain sum of the four disjoint counts; `keepCoverOnlySkippedCount` adds `nothing_reproducible` as a fifth summand. `keepCoverOnlyMenuLabel`'s new `label` parameter keeps both menu items rendering from one function.

**Design tokens.** `--space-2`, `--text-sm`, `--text-xs` and `--leading-body` all exist in `docs/design/design-tokens.css`; no hardcoded hex, ad-hoc radius, raw `rgba` shadow, or off-ramp font size.

**Wiring.** `ImageGridContextMenu.onAction` forwards the payload only when given, `SelectionBar` forwards `$event`, `ImageGrid` binds `openKeepCoverOnly` in both places, so the plain item still arrives with `options === undefined`. The operation store gets both the icon and the destructive-duration rule for the new op type. Menu gating is identical to the sibling, and the new menu tests assert position, count label, absence at zero stacks, and disabled+reason on both menus.

### Tests

`tests/test_keep_cover_only.py` is already in the `backend` job file list (`.github/workflows/ci.yml:514`) — pre-existing file, this PR only appends. Gating requirement met.

The new tests use the file's existing per-test `_env()`/`_teardown()` shape rather than a module-scoped fixture. Against standing guidance, but it is the shape the whole file already has; converging it is a separate change.

The assertion guarding the data-loss path is `test_keep_recipes_only_moves_only_what_could_be_made_again`: `preview["picture_ids_moving"] == [pics["covered"]]` and `body["picture_ids_moved"] == [pics["covered"]]` — identity, not counts, over a six-picture fixture with one representative of each staying reason. Loosen any one of the four conditions and an extra id appears and it goes red. `test_retention_off_...` and `test_a_missing_thumbnail_file_or_pixel_sha_keeps_the_copy` pin the retention and file-level conditions the same way. Both directions covered. The gap is must-fix 1: there is no test for a non-`.safetensors` model, because there is no behaviour to assert yet.

lindkvis:
Answered in f6e96331. Both must-fixes are fixed; of the six should-fixes, five are fixed and one is answered without a code change.

**1. Unjudgeable model formats — fixed.** `recipes_missing_a_model` now flags a recipe naming any asset whose extension is in `MODEL_EXTENSIONS` but is not `.safetensors`, alongside the model-ghost set. The reasoning is in the docstring: for the ghost *screen* an unjudgeable name must not be called a ghost (it would forget a model still on disk); for deciding what to *delete*, "we do not know" has to keep the picture. The `judge_digests` suspension is recorded there too, as the window it is. Test: `test_a_model_the_shelf_cannot_judge_keeps_the_copy`, red when the predicate is removed.

**2. Preview and confirm — fixed, the 409 shape you suggested.** The confirm sends `expected_picture_ids` (the preview's `picture_ids_moving`); `keep_cover_only_in_session` raises `KeepCoverOnlyPlanGrew` → **409**, nothing written, when the plan would move an id that list does not carry. A plan that *shrank* still runs: moving fewer than consented to is inside the consent. The grid warns and re-previews rather than retrying. Only Keep recipes only sends it, since it is the mode whose inputs move on their own. Test: `test_a_plan_that_grew_since_the_preview_is_refused`, which writes a thumbnail between the preview and the confirm.

**3. `pictures_staying` comment — fixed.** Both comments now say what the scoping actually is: a locked stack was never judged copy by copy, a character-skipped stack was but does not change at all, so its copies are named in its own row rather than in a total about a collapse that did not happen. The mutation and the preview scope it identically.

**4. Retention write ordering — fixed, setting first.** The route passes a `before_write` callback; the service calls it after the plan, only when something will move, and **before the first write**. A failure raises out with nothing moved, so `ghost_retention_saved` is gone — there is no half-consented state left to report. Test: `test_keeping_every_ghost_is_in_force_before_anything_moves` makes the write raise and asserts nothing moved; it fails if the call is moved after the commit.

**5. `ponytail:` blast radius — comment corrected, and the write removed.** The comment now says both reads are whole-hub whatever the selection, that they run inside the serialised vault task and so queue every background finder, and that the dialog re-runs on each tick of the box. The side effect is gone: the planner no longer calls `find_thumbnail` (which `os.replace`s legacy bitmaps) but a read-only `_has_thumbnail` that looks in the same places. The reads themselves stay inside the task for now — moving them out is a change to how the planner is called, and the comment names it as the upgrade.

**6. Two look-alike danger items — fixed, past `lead-designer` and `ui-ux-expert`.** Keep recipes only gets `mdi-image-refresh` (a frame with a circular arrow: it states that the picture could be made again, without claiming the app will re-make anything). `mdi-layers-remove` was rejected because its cross reads as more severe than the sibling's minus, on the less destructive item. It now **leads** the danger group, above Keep cover only: it moves a strict subset of the same pictures, so escalating severity puts it first, and the cheapest slot goes to the safer item. Applied at all four naming sites including the receipt icon. The order comments in both menus and `keep-cover-only.md` are updated, and the menu tests assert the new order and that the two glyphs differ.

**7. Checkbox vanishing mid-flight — fixed.** "Has this selection had copies staying for want of a ghost" is now latched in the dialog and reset on open, so the box stays mounted and is disabled while the re-run lands. Test asserts it survives `preview: null, loading: true`.

**8. Accessibility — fixed.** The checkbox and its note moved out of the `role="status"` region into their own panel, so a preview update no longer re-announces a control, and the region renders only with its title, so `aria-labelledby` cannot dangle.

**Nits — all three taken.** The note now ends "Ctrl+Z does not turn it back off". The lede branches on the live `ghost_retention`: under `covered` it says the recipe lasts "for as long as its cover stays — delete the cover for good and they go with it", which is the thing a reader could not derive. The `ghost_not_kept` row leads with the consequence: "N share their prompt with no picture that stays, so your ghost setting would keep nothing of them."

**Security nits — both taken.** `apply_ghost_retention` now validates the position itself and raises `ValueError` otherwise, so the shared chokepoint is safe by construction rather than by its callers. The coverage-matrix row for `POST /stacks/keep-cover-only` says it can write the server-wide `workflow_ghost_retention`, and that the PATCH it shares that write with is `owner_only` too.

**Not changed:** the per-test `_env()` shape in `tests/test_keep_cover_only.py`, agreed as a separate change.

**Unrelated, but it will redden this PR's `checks` job:** `pixlstash/services/workflow_hash.py:590` uses `_SEED_RE`, which is defined nowhere. It came in with 15c03034 (on `develop`), so `ruff check` is red on `develop` itself and that line raises `NameError` if it is reached. Not touched here.

Local runs: `tests/test_keep_cover_only.py` (33) + `test_workflows_api.py` + `test_architecture_guardrails.py` = 137 passed; frontend 3510 passed. Each new assertion was checked by breaking what it guards.

Issue #1376, opened 2026-09-16
Insert a LoRA loader into a workflow that has none


lindkvis:
#1310 puts a shelf LoRA into a workflow by patching the loaders the graph already has. A workflow with no LoRA loader is refused with "This workflow has no LoRA loader, so there is nothing to put a LoRA into."

That refusal is the placeholder for the interesting half: let the owner *add* one from PixlStash, interactively.

- Splice a loader between the model source's MODEL/CLIP outputs and their consumers, so the whole downstream chain keeps working.
- Handle the shapes a naive splice breaks on: model-only loaders with no CLIP, several consumers of one MODEL output, GGUF/UNET chains, a graph whose CLIP comes from a separate loader.
- Show where it would go and what it would rewire before it happens, rather than mutating a graph the owner cannot see.
- Decide which loader to insert: a plain `LoraLoader` (works anywhere, needs the file on that ComfyUI) or `PixlStashAdapterLoader` (resolves by digest, needs ComfyUI-PixlStash installed) - the same choice #1310 made per existing slot.
- Stacking more than one adapter is the natural extension once insertion exists.

PR #1377, opened 2026-09-16
Put a LoRA from the shelf into a workflow run

### LINKED ISSUE #1310: Put a LoRA from the shelf into a workflow
Before a run, add or swap a LoRA from the model shelf into the workflow.

- `detect_lora_targets` / `apply_adapter`, the same shape as `detect_seed_targets` / `apply_seeds`.
- Resolve a shelf model to a name the target ComfyUI understands.
- Decide what happens when the graph has no LoRA loader.

**Decision needed first:** patch a plain `LoraLoader` (works on any ComfyUI, needs the file there) or substitute the PixlStash adapter loader node (resolves by shelf id, needs the node installed).

Depends on: #1307

lindkvis:
Closes #1310.

**The decision the issue asked for:** patch the loaders a graph already has — a core `lora_name` loader gets a filename this ComfyUI lists, a ComfyUI-PixlStash loader gets `adapter_sha256`. Nothing is substituted or inserted, so it works on any ComfyUI. A graph with no LoRA loader is refused by name; #1376 is the interactive inserter.

**Every route that runs a workflow takes it**, through one resolver so they refuse the same things in the same words:

| Surface | Route | Slots read from |
|---|---|---|
| Run panel | `POST /comfyui/workflows/{name}/run` | `GET .../inputs` |
| Edit with ComfyUI | `POST /comfyui/run_i2i` | each row of `GET /comfyui/workflows` |
| Generate variants | `POST /comfyui/run_recipe` | `GET /comfyui/pictures/{id}/recipe` |

Backend: `detect_lora_targets` / `apply_adapter` beside `detect_seed_targets` / `apply_seeds`. Frontend: one composable, `useLoraSwap`, behind all three surfaces. Slots are found by field name, not class, so every pack that copies ComfyUI's `lora_name` widget (numbered stacker slots included) works with no list to maintain.

**In a replay the swap is applied before the pre-flight judges the graph.** A picture made with a LoRA that has since left this ComfyUI is exactly the one worth re-running with another, and pre-flighting the file it no longer uses refused it. A test pins both directions: swapped it runs, unswapped it is still refused by name.

**One slot per run, and a slot is a node and a field.** A graph chaining a style LoRA and a character LoRA would come back loading the chosen file twice with the other gone, and a stacker holds several LoRAs on one node, so `lora_node_id` and `lora_field` name the slot; each surface asks only when there is more than one.

**The filename match is by name, not content** — `object_info` carries no digests, so a file of that name on that machine is all ComfyUI can be asked for. Documented as such rather than dressed up; the digest slot is the exact one.

Things a reviewer may want to push back on:
- A stacker holding its slots as dicts under its own key (rgthree's Power Lora Loader) reads as no slot, so such a workflow is told it has no LoRA loader. `lora_name_1` / `lora_name_2` stackers do work. Reading one pack's own widget shape belongs with #1376.
- The picker offers every adapter on the shelf, not only `kind = lora`: LyCORIS formats load through the same loaders.
- A recipe whose pre-flight fails for some *other* reason still blocks the Remix dialog before a LoRA can be chosen. Only the LoRA being replaced stopped blocking.

Tests: service unit tests, API tests on both run routes and the recipe route, panel and dialog tests. Every new assertion was anchored by mutating what it guards and watching it go red.


lindkvis:
Independent backend and frontend reviews ran on this before merge. Fixed in 067bd604:

- **A stacker's slots were indistinguishable.** `lora_name_1` / `lora_name_2` share a node, so `lora_node_id: "7"` swapped every LoRA on it. A slot is now node **and** field (`lora_field`), and whatever the two leave must be exactly one slot.
- **LoRA filenames were readable by share-link tokens.** `GET /comfyui/workflows` is `ANY_TOKEN` by design; its rows' slots no longer carry `value`. Test covers both directions.
- **A swap reaching no slot ran the graph's own LoRA.** Now a 500, and the write goes through `api_graph`, so a `{"prompt": graph}` document is reached.
- **A LoRA carried into the next run.** Generate variants kept the last picture's choice and opens with focus on Generate; a recipe choice rode into a template with the same slot. The choice now resets whenever the graph changes. The three hand copies became one composable, since each bug had been found three times.
- Smaller: empty `lora_node_id` treated as absent, a UI-format file refused as that, a missing loader class named as a missing node, a replay against an unreachable ComfyUI now a 502 like the other routes, "no LoRA loader" only said once it is known, a11y roles, a visible label on the second select, the shelf list sorted with base model.

Deliberately not changed:
- **Stale copies' `relpath` as candidate names**: a gone copy's path is still the file's name, and a match is against what ComfyUI lists, so it can only ever find the file there.
- **Ambiguity refused on the first ambiguous candidate**: candidates are the same file's paths and share a basename.
- **`lora_node_id` sent without `adapter_sha256` is ignored**: nothing to put in the slot.
- **Sync `object_info` / hub calls in async handlers**: the routes already upload and submit synchronously; moving them belongs to that wider change.
- **`unknown` adapters not offered in the pickers**: the shelf lists them apart from adapters; the API still accepts them.

Not component-tested: the overlay's two wiring lines (reset on paging to another picture, the body spread). The rules they call are covered by `useLoraSwap.test.js`; mounting the overlay's ComfyUI menu needs a `v-menu` stand-in the existing overlay specs avoid.

Issue #1321, opened 2026-09-13
Unified screens: Models


lindkvis:
Bring the Models screen onto the shared shell. Restyle only.

Depends on: #1317

Issue #1319, opened 2026-09-13
Unified screens: Duplicates review


lindkvis:
Bring Duplicates review onto the shared shell. Restyle only.

Depends on: #1317

PR #1384, opened 2026-09-16
Unified screens: one inspector

### LINKED ISSUE #1317: Unified screens: one inspector
Overlay metadata, grid stats, model detail and duplicates evidence all use the shared inspector.

Restyle only: no flow changes. Design: unified screens in the design system project.

Depends on: #1301

lindkvis:
Closes #1317.

The lightbox's side panel (Description, Faces, Tags, Metadata) now uses `AppInspector`, the same pane as grid stats and the workflow inspector. It's a restyle with no flow changes.

- `AppInspector` gets a `lightbox` prop. The pane is a translucent dark surface in both themes, and it's a fixed-height column whose sections scroll themselves. Its content stays mounted while closed (`v-show`), because the overlay opens the description editor as it opens the pane.
- The pane is 288px (`--stats-panel-w`) instead of 320px, and the body uses the inspector's padding and gap.
- The Metadata box uses the inspector's tab band one level down (small caps, olive underline) and `dl.inspector-kv` for picture information.
- `max-width` is added to the inspector's transition, so a pane slides shut instead of snapping. This applies to Stats too.

**Model detail and duplicates evidence:** neither pane exists yet. Building one is a new feature, not a restyle, so they're left to #1321 and #1319, which depend on this PR and can use `AppInspector` directly.

Vitest is green on the related suites, and the new lightbox test fails if the content remounts or stops hiding. eslint and the build pass. I did not run the e2e tests.

**Pre-push review, answered here**
- *44px bottom padding dropped.* Nothing is positioned over the bottom of the pane (checked in `ImageOverlay.css`), so the inspector's own padding replaces it.
- *Closing resets tag-list scroll.* True. The old zero-width aside reflowed its lists too, and I kept it simple.
- *Pane is now a named landmark, `aria-hidden` when closed.* Intended: before this, the closed pane's controls could still be reached with Tab at zero width.
- *Values are larger (`--text-sm`) in a narrower pane, and the Metadata tab band is 36px.* Both follow the inspector spec, which gives the nested tabs the same band.
- *The ≤720px 78% rule is dead.* It was already: the pane is `display: none` below 900px.
- Fixed from the review: the stale transition comment in the receipt, a comment typo, and the test now checks the same node survives close and reopen.


frontend/src/components/views/OverlayMetadataPanel.vue
```diff
@@ -479,70 +475,53 @@ watch(
...
 .metadata-tab-btn.active {
   color: rgb(var(--v-theme-on-dark-surface));
-  background: rgba(var(--v-theme-dark-surface-primary), 0.2);
+  border-bottom-color: rgb(var(--v-theme-dark-surface-primary));
 }
 
 .metadata-tab-btn:hover:not(.active) {
   color: rgb(var(--v-theme-on-dark-surface));
-  background: rgba(var(--v-theme-on-dark-surface), 0.16);
+  background: rgba(var(--v-theme-on-dark-surface), 0.08);
```
lindkvis: **Non-blocking, but the one I'd most like changed — the hover wash on the metadata tabs is half the documented value.**

`.metadata-tab-btn:hover:not(.active)` goes from `rgba(var(--v-theme-on-dark-surface), 0.16)` to `0.08`. `docs/design/visual-language.md:752–753` states the rule directly:

> **Hover is an ink wash** (2026-09-13). A transparent control (row, tab, bar button, segment, menu row) takes `--hover-wash`, the text colour at 16% per theme.

The old `0.16` *was* that value spelled by hand for a surface that's dark in both themes — which is the right call here, since `--hover-wash` is per-theme and would be wrong on this pane. That's the same argument this PR makes two rules up for `--active-bar`. For what it's worth, `0.16` is also the most common on-dark alpha in the codebase (45 occurrences) against 24 for `0.08`, none of which is a tab hover.

Why it matters beyond consistency: this pane is a 60%-opaque dark surface over an arbitrary photograph, so an 8% wash is the weakest hover cue in the app and can be close to invisible over a bright image.

**Fix:** restore `0.16`.

---

frontend/src/components/widgets/OverlayActionReceipt.vue
```diff
@@ -194,8 +194,8 @@ defineExpose({
   display: flex;
   justify-content: center;
   pointer-events: none;
-  /* `--sidebar-width` flips instantly (0 to 320px) while `.overlay-sidebar`
-     slides its width over 200ms. Without this the pill jumps mid-glide. */
+  /* `--sidebar-width` flips instantly (0 to the pane's width) while
```
lindkvis, line 197: **Non-blocking — the pane's width transition no longer matches the receipt that tracks it.**

The receipt slides `left`/`right` over `--dur-2` (200ms) specifically so it doesn't jump while the sidebar glides. Before this PR the sidebar was `transition: width 0.2s ease` — also 200ms, so they matched. The pane now inherits `AppInspector`'s `width var(--dur-1)` (150ms), so they're 50ms apart and the receipt finishes 50ms after the pane settles.

The comment edit on this line removes the "200ms" figure, so the coupling was clearly noticed — the durations just weren't re-synced with it.

Against the token file, `--dur-2` is documented as the duration for "panels, expand/collapse" (`design-tokens.css:368`), and a 288px pane collapsing is exactly that. `AppInspector` using `--dur-1` for width is pre-existing rather than introduced here.

**Fix:** either move `AppInspector`'s `width`/`min-width`/`max-width` transition to `--dur-2`, which matches the token's documented intent, or change the receipt to `--dur-1`.

---

frontend/src/components/views/ImageOverlay.css
```diff
@@ -588,38 +588,23 @@
   padding: 0 var(--space-2);
 }
 
-.overlay-sidebar {
+/* Width, collapse and surface come from AppInspector; this places it over the
```
lindkvis, line 591: **Non-blocking — the rule is correct, but this comment states a mechanism that isn't the one at work.**

The comment says "Two classes, so it outranks the inspector's own `height: 100%` whichever stylesheet loads last." Two classes alone would *tie*: `.overlay-main > .overlay-sidebar` is (0,2,0), and `AppInspector`'s `.inspector` is (0,1,0) plus one attribute from its scope id — also (0,2,0), decided by source order, which is exactly the thing the comment claims not to depend on.

What actually wins is that Vue stamps this file's scope id on the last compound too, giving `.overlay-main > .overlay-sidebar[data-v-…]` at (0,3,0) against `.inspector[data-v-…]` at (0,2,0).

Why it's worth fixing rather than leaving: the comment invites a later editor to simplify the selector back to `.overlay-sidebar`, believing the two classes were the mechanism. That would tie — and with `top` and `bottom` both set, the resulting `height: 100%` would over-constrain the box and run the pane past the bottom edge.

**Suggested wording:** "Three selector parts once Vue stamps this file's scope id on the last one, so it outranks `AppInspector`'s own `.inspector` rule regardless of stylesheet order."

---

frontend/src/components/widgets/AppInspector.vue
```diff
@@ -98,6 +106,32 @@ const emit = defineEmits(["update:modelValue"]);
   flex-direction: column;
 }
 
+/* The lightbox is dark in both themes, so its pane is a translucent
```
lindkvis, line 111: **Non-blocking, and as much a question as a request — the lightbox variant doesn't re-ink `AppInspector`'s own tab band.**

`.inspector--lightbox` overrides background, border and text colour, but not the tab band: `.inspector-tab` still uses `var(--hover-wash)`, `var(--active-text)` and `var(--active-bar)` (lines 170–177, untouched). Those are the per-theme tokens — and this PR's own comments twice explain why per-theme tokens are wrong on a surface that's dark in both themes.

Latent rather than live: `ImageOverlay` passes no `tabs`, so the band never renders today. But `lightbox` is now a public prop, and the next caller passing both `lightbox` and `tabs` gets light-theme olive on a dark surface.

Either add the three on-dark overrides under `.inspector--lightbox .inspector-tab`, or say in the prop docblock that `lightbox` doesn't currently support `tabs` — whichever you'd rather the next caller discover.

---

frontend/src/components/views/ImageOverlay.vue
```diff
@@ -820,9 +820,14 @@
...
         />
 
-        <aside
+        <!-- The shared inspector (shell contract rule 8), in its lightbox
+             variant: dark in both themes, laid over the canvas. -->
+        <AppInspector
           class="overlay-sidebar"
-          :class="{ open: sidebarOpen, hidden: chromeHidden }"
+          :class="{ hidden: chromeHidden }"
+          label="Sidebar"
```
lindkvis: **Non-blocking — `label="Sidebar"` names the chrome rather than the subject.**

Two things, both bearing on the PR body's accessibility claim.

**(a)** The prop's own docblock says the label is "The accessible name of the pane ("Stats", "Inspector")", and the component's tab-band comment says "Tabs name the SUBJECT of the pane, never the view." "Sidebar" is the view. The pane holds Description, Faces, Tags and Metadata, so "Picture details" or "Details" would tell a screen-reader user what they're entering.

**(b)** More concretely: the overlay root (line 2) is a plain `<div class="image-overlay">` with no `role="dialog"` / `aria-modal="true"`, and the grid behind it stays mounted — including `StatsSidebar`, itself an `AppInspector` labelled "Stats" (`StatsSidebar.vue:1154–1157`). With both panes open the landmark list now shows two `complementary` regions, "Stats" and "Sidebar", with nothing saying which belongs to the thing actually on screen.

A distinct name mostly resolves that. Making the overlay a real modal would resolve it properly, but that's a bigger change and out of scope here.

---

lindkvis (commented):
## Frontend + design review — one inspector

**No blocking findings.** Saying that plainly rather than padding it: the load-bearing decisions hold up under inspection.

### What I checked and found correct

- **The `v-if="open || lightbox" v-show="open"` keep-mounted decision is correct and necessary.** `openSidebarFromTeaser` (`ImageOverlay.vue:2810`) and the `T` shortcut (`:2488`) both set `sidebarOpen = true` and then reach into a panel ref. Both reach their `.focus()` inside a `nextTick`, which runs after the component update queue flushes, so `display: none` is already gone. A plain `v-if` would have broken the teaser path.
- **`aria-hidden` and `v-show` agree in both directions.** Closed → the content subtree is `display: none`, so nothing inside is tabbable while the `<aside>` carries `aria-hidden="true"`. Open → both flip in the same render. No window where focusable content is hidden-but-tabbable or tabbable-but-hidden. The PR body's claim holds.
- **The flex chain the tag list depends on survives the extra nesting level:** `.inspector` (definite height from the overlay's absolute box) → row flex item `.inspector-content` (stretched, `overflow: hidden`) → `.inspector-body` (`flex: 1; min-height: 0`) → `.sidebar-section--tags`.
- **`max-width` in the transition is a fix, not a regression.** `.inspector--collapsed` already set `max-width: 0` without transitioning it, so the used width clamped to 0 instantly and the *closing* animation never played on Stats or the Workflow inspector. It now does. That's a real behaviour change on two screens this PR doesn't otherwise touch, and it's the change most worth an eyeball before merge.
- **Class removals are complete.** `.sidebar-section { margin-bottom: 6px }` is gone from all four places it existed, and every surviving `.sidebar-section` element is a direct flex child of `.inspector-body`, so the new `gap: var(--space-3)` covers them. `.metadata-info-grid` and `.metadata-info-label` have zero remaining references in `frontend/src`, `frontend/e2e` or `docs/`.
- **The token claims check out.** `--stats-panel-w` is real (`design-tokens.css:282`, 288px) and used in both places claimed; padding and gap are `--space-3`/`--space-4`/`--space-5`. Three off-token literals were *removed* (the 6px margin, 6px tab padding, 10px tab-panel padding), each of which carried an explicit "no token" comment. No hardcoded hex, no ad-hoc radius, no raw `rgba(0,0,0,…)` shadow and no off-ramp `em`/`px` font size in the added lines.

### Four notes that can't anchor inline

**`--sidebar-width` and the pane's width are two spellings of one number at ≤720px.** `--sidebar-width` no longer feeds the pane's width — `AppInspector` owns that — so at desktop widths the two agree only because both read `--stats-panel-w` (`ImageOverlay.css:31`, `AppInspector.vue:71`). At ≤720px they're two independent `78%` literals: `--sidebar-width: 78%` on `ImageOverlay.css:886`, and `width`/`min-width`/`max-width: 78%` just below. `--sidebar-width` is what `OverlayActionReceipt` positions against, so changing one without the other drifts the receipt off the pane edge. Not a defect today; a one-line comment tying them together, or having the width rules read `var(--sidebar-width)`, would close it.

**The label/value gap inside a metadata row drops from 2px to 0.** `.metadata-info-item` loses its `display: flex; flex-direction: column; gap: var(--space-1)`, and the shared `dt`/`dd` rules set `margin: 0` on both, so name and value now sit on consecutive line boxes with nothing between them. This is consistent with `WorkflowInspector.vue:133–160`, which uses the same `dl.inspector-kv > div > dt + dd` shape with the same zero gap — so it's plausibly deliberate, and "shows its values the same way" is the changelog's phrasing. Asking rather than flagging: intended, or should `.inspector-kv dt` carry a small `margin-bottom` in the shared component so every pane gets it?

**Doc line says `title`, the component reads `tooltip`.** `docs/frontend_architecture.md:1450` still documents `tabs` as `[{ value, label, icon, disabled, title }]`. The prop docblock says `tooltip?` (`AppInspector.vue:54`), the template reads `tab.tooltip` (`:31`) and the test fixture uses `tooltip` (`AppInspector.test.js:12`). Pre-dates this PR, but the line is being rewritten here anyway — free fix.

**Template comment left badly wrapped.** `AppInspector.vue:11` runs to ~105 characters against ~80 on the lines around it. Prettier doesn't reflow HTML comments so lint passes; it just reads as an unfinished edit.

### The new test is sound, with two small gaps

`AppInspector.test.js:74–91` does real work and can fail for the right reasons: drop `v-show` and the `display: none` assertion goes red; narrow `v-if` back to `open` and `content.attributes("style")` is `undefined` so `.toContain` throws. The same-element identity check on the probe is the right way to prove "not a remount" — good call adding it after the pre-push review.

Two suggestions, neither a blocker: it never asserts the *open* case clears `aria-hidden` (line 83 covers only the closed case), which is one line and is the half of the landmark contract the PR body actually claims. And `mountInspector` passes `tabs: TABS`, so the pane under test renders a tab band the real caller never has — harmless, but it doesn't exercise the shape `ImageOverlay` actually mounts (`tabs: []`).

### Not verified

No tests were run and nothing was rendered, so every layout, contrast and truncation statement above is from reading the CSS cascade rather than pixels. Three things specifically need eyes:

- **Text truncation in the narrower pane.** The pane goes 320px → 288px while the metadata type ramp goes *up* (`dt` 11px → 12px, `dd` 12px → 13px). The key/value column lands at roughly the same width (~132px) because `.metadata-tab-panel` lost its 10px horizontal padding, so the net is the same column carrying ~8% larger text under `white-space: nowrap; text-overflow: ellipsis`. Whether any real value now ellipsises that didn't before needs looking at.
- **Contrast of the dt/dd inks over a photograph.** Both moved the safe direction on paper (label 0.6 → 0.7, inactive tab 0.5 → 0.7), but nothing measured them against a 60%-opaque dark surface over live image content.
- **The `.overlay-sidebar.hidden` state.** When `chromeHidden` and `sidebarOpen` are both true — reachable, since `onCanvasClick` (`ImageOverlay.vue:2798`) sets `chromeHidden` without touching `sidebarOpen` — the pane is `opacity: 0; pointer-events: none` but still in the tab order and *not* `aria-hidden`. Unchanged by this PR (`ImageOverlay.css:603–606`), so not written up as a finding; noted only because the PR body raises the landmark/`aria-hidden` question and this is the one state where the two don't agree. Whether it's tracked elsewhere is outside what I read.

---

frontend/src/components/views/OverlayMetadataPanel.vue
lindkvis: Fixed in c63978b2: the hover wash is back to 0.16.

frontend/src/components/widgets/OverlayActionReceipt.vue
lindkvis: Fixed in c63978b2: AppInspector's width, min-width and max-width transitions now use `--dur-2`, matching the token's expand/collapse intent and the receipt's glide. Stats and the workflow inspector get the same 200ms.

frontend/src/components/views/ImageOverlay.css
lindkvis: Fixed in c63978b2: the comment now says the win comes from the scope id stamped on the last selector part (three parts against two), not from source order.

frontend/src/components/widgets/AppInspector.vue
lindkvis: Documented rather than built, in c63978b2: the `lightbox` prop docblock now says its tab band isn't re-inked for the dark surface, and a lightbox pane that needs `tabs` has to add the on-dark tab colours first. No caller passes both today, and on-dark tab colours nobody has looked at seemed worse than a stated limit.

frontend/src/components/views/ImageOverlay.vue
lindkvis: Fixed in c63978b2: the pane is labelled "Picture details". Making the overlay a real modal dialog is left out of scope, as you suggest.

lindkvis:
Notes from the review body, in c63978b2:

- **78% at ≤720px:** the pane's width rules now read `var(--sidebar-width)`, so the pane and the receipt share one number. (That rule is still unreachable today, because the pane is `display: none` below 900px.)
- **Doc said `title`:** now `tooltip`.
- **Wrapped comment:** reflowed.
- **Test:** the lightbox test mounts with `tabs: []`, as the overlay does, and asserts `aria-hidden` clears on open. Forcing `aria-hidden` on turns it red.
- **Label/value gap:** left at 0 on purpose. It matches `WorkflowInspector`'s `dl.inspector-kv`. If the shared `dt` should get a margin, that's a change to every pane and belongs with the inspector's own design, not this PR.
- **`.overlay-sidebar.hidden` still tabbable when the chrome is hidden:** true, and unchanged here. Not fixed in this PR.

PR #1385, opened 2026-09-16
Task activity indicators use accent instead of selection olive

### LINKED ISSUE #1343: Design Cleanup: The task indicator pulse on the Stats Sidebar needs to be a different colour
Image

It is currently green, which has no contrast against the selected task icon, but also, green is selection, not indication of anything.

lindkvis:
Closes #1343.

The Tasks-tab pulse, busy Tasks icon, per-task running dot and the toolbar stats-toggle activity dot were `primary` (olive), which is the selection colour and blended into the selected Tasks tab. They now use `accent`, which is what visual-language §12 already prescribes for attention dots and what SideBar's existing attention dots use.

Review notes, answered rather than acted on:
- **Light-theme contrast:** amber on the light sidebar is ~2.9:1 against ~4.7:1 for olive. That's the token §12 names and the sidebar's other dots already sit at it. Raising it is a palette decision, not this issue's.
- **Amber also reads as `warning`:** same trade-off. The manual deliberately makes accent the attention colour.
- **Row running dots changed too:** they're the same "work is live" signal. Leaving them olive would keep the mixed meaning the issue objects to.
- **Not done:** dot sizes stay 6/7px rather than `--badge-size-dot`. `docs/design/v1.8.0-preview.html` is a frozen mockup and still shows olive.


frontend/src/components/panels/StatsSidebar.vue
```diff
@@ -2462,8 +2462,8 @@ defineExpose({ focusTasksTab });
 }
 
 .tm-status-dot--running {
-  background: rgb(var(--v-theme-primary));
-  box-shadow: 0 0 5px rgba(var(--v-theme-primary), 0.55);
+  background: rgb(var(--v-theme-accent));
```
lindkvis, line 2466: **Non-blocking — contrast, for the record rather than to hold this up.** Please check my arithmetic.

A pulsing activity dot isn't text, so 4.5:1 doesn't apply. It is a graphical object needed to understand the content, so it owes **WCAG 2.2 SC 1.4.11 Non-text Contrast, 3:1**. `visual-language.md` §4 already works in those terms ("the 3:1 UI floor").

From the values shipped in `main.js` — `accent: #c47a1e`, `primary: #567309`, same in both themes — against the real backdrop (`sidebar`/`toolbar`: light `#f0ede9`, dark `#23282f`):

| Backdrop | before (`primary`) | after (`accent`) |
|---|---|---|
| light `#f0ede9` | 4.67:1 — passes | **2.93:1 — fails 3:1** |
| dark `#23282f` (default) | **2.72:1 — fails 3:1** | 4.34:1 — passes |

So this moves a 1.4.11 failure from the default theme to the light theme rather than introducing one. §4 states dark is the default, so the outgoing olive was already failing for most users, and blocking on contrast would block a net improvement.

It also isn't local to these four rules: the three shipped attention dots in `SideBar.css` (658, 1387, 1410) are `accent` on `sidebar` too — the identical 2.93:1 in light. The whole attention-dot pattern fails 1.4.11 in the light theme today; this PR brings four rules into line with that pattern. Fixing it here would mean inventing a per-component colour for four dots, which is the drift the token system exists to prevent.

**Suggested follow-up covering all seven dots:** either an `--attention-dot` alias resolving to `accent` in dark and a deepened amber in light (`#9e6727`, the value §4 still documents, measures 3.9:1 on `#f0ede9`), or keep the fill and add a 1px full-opacity `on-surface` ring, which satisfies 1.4.11 through the boundary at no cost in either theme. The first is cleaner and is a design call rather than an inline one.

Cross-check on the method: `style.css:78` records "the deep one measures 2.32:1 on a dark panel" for this same olive, and recomputing `#567309` against `#313337` returns 2.32:1.

---

frontend/src/components/panels/TbGlobalActions.vue
```diff
@@ -73,8 +73,8 @@ const statsTitle = computed(() =>
   width: 7px;
```
lindkvis: **Non-blocking — agreeing with the deferral, recording it so it survives the merge.**

§12 gives the attention dot as `--badge-size-dot` (8px) at `--radius-pill`. These three dots are 7px, 7px and 6px at `border-radius: 50%`. The token file names this exact value as the thing it replaces:

```
frontend/src/styles/design-tokens.css:243
  --badge-size-dot: 8px;   /* the no-count activity dot (on-grid; was an ad-hoc 7px) */
```

After this PR there are two populations of attention dot: the three in `SideBar.css` (658, 1387, 1410) on the token at 8px/`--radius-pill`, and these three off it at 6–7px/`50%` — two of which differ in size from each other with no stated reason. The colour split is now resolved; the geometry split isn't.

Your reasons for leaving it hold up: a size change is pixel-moving, which §12's own "One band" precedent treats as UI/UX-gated rather than a designer's inline call; `top: 7px; right: 7px` on this rule is positioned against the current 7px diameter, so 8px silently re-centres the dot and wants a look in situ; and folding geometry into a colour diff makes a clean revert harder.

Suggestion: one issue naming these three selectors, landed together with the contrast follow-up — they touch the same declarations and should be a single geometry-plus-colour pass rather than two.

---

frontend/src/components/panels/StatsSidebar.vue
```diff
@@ -2504,17 +2504,19 @@ defineExpose({ focusTasksTab });
...
 .tm-tab-icon--busy {
   animation: tm-dot-pulse 1.4s ease-in-out infinite;
-  color: rgb(var(--v-theme-primary));
+  color: rgb(var(--v-theme-accent));
 }
 
 .tm-tab-pulse {
   width: 6px;
   height: 6px;
   border-radius: 50%;
```
lindkvis: **Question, not a finding.** Was leaving `border-radius: 50%` deliberate, or just out of scope along with the size?

§12 specifies `--radius-pill` for the attention dot, and the three `SideBar.css` dots use it. Unlike the size, this one is free — `--radius-pill` is `999px` and renders identically at 6–7px — so it could come along now if you want it, or wait for the geometry pass. Either is fine; nothing needs changing on my account.

---

lindkvis (commented):
## Design review — activity indicators to `accent`

The colour decision is right, correctly sourced and cleanly tokenised. **One blocking finding: the sweep stopped one rule short, in the same file, 40 lines below the last hunk.** Three inline notes besides, none of them asking for changes here.

### Blocking — the import landing-glow is still olive, and its own comment says it's amber

`StatsSidebar.vue` 2562–2575 (untouched lines, hence here rather than inline). `@keyframes tm-import-glow` still carries five `--v-theme-primary` stops — the `box-shadow` at 0%/35%/100% and the `border-color` at 0% and 35%.

Its own header comment 19 lines above already says what it should be:

> "Landing punctuation for the async import (#459): a one-shot spring-pop + **accent** glow-and-settle when the FLIP flight from the import dialog lands the chip on this row (mirrors the `gridNewPulse` landing-pulse pattern, visual-language §10)."

The manual names the colour twice. §10: "**Landing pulse (one-shot)** — a single glow-and-settle when something new arrives. Reference: `gridNewPulse` — a one-shot `--v-theme-accent` glow ring". §12: "when a badge appears or its count ticks up because work *just landed*, play the one-shot landing pulse (§10, `gridNewPulse` model, **accent glow**)". The reference implementation it says it mirrors, `.thumbnail-card-new` / `gridNewPulse` in `ImageGrid.css:812–830`, is already accent at every stop.

**Why this belongs in this PR rather than a follow-up.** A task-manager row announcing that work *just landed* is a live-work indicator by this PR's own title, so it's the fifth member of the class the other four changes fix. It's in the file this PR edits, inside the same `<style>` block. And the new comment at 2507–2508 ("Live work wears the accent, not primary") now sits 54 lines above a rule that does the opposite while claiming in its own comment to do the same — leaving the next reader to work out which one states the intent.

**Fix:** five token swaps, `primary` → `accent`, on the three `box-shadow` stops and the two `border-color` stops at 0% and 35%. Leave the 100% `border-color` as `on-surface`; that's the resting border, not the glow, and it's already correct.

### The premise checks out, on both halves

§12 prescribes `accent` for the attention dot verbatim, in the shape table and in the rationale under it ("the activity dot is accent"). The second half — that olive is the selection mark — holds too, and the collision is specific rather than rhetorical: `style.css:77` records "SELECTED IS OLIVE (2026-09-13): tile, row, tab, focused group", and `AppInspector.vue:140–143` draws `.inspector-tab--active` with `border-bottom-color` resolving to `primary` in light. Before this change the Tasks tab's *busy* icon and the *selected* tab's underline were the same olive on the same control. The new comment states this accurately.

### Completeness: 4 of 5

Every looping or one-shot indicator animation in `frontend/src`:

| Selector | Animation | State |
|---|---|---|
| `.thumbnail-card-new` (`ImageGrid.css:812`) | `gridNewPulse` | already accent |
| `.thumbnail-placeholder` (`ImageGrid.css:974`) | `skeletonPulse` | skeleton, not an indicator |
| `.tb-stats-btn--busy::after` | `tb-stats-pulse` | **changed** |
| `.tm-status-dot--running` | `tm-dot-pulse` | **changed** |
| `.tm-tab-icon--busy` | `tm-dot-pulse` | **changed** |
| `.tm-tab-pulse` | `tm-dot-pulse` | **changed** |
| `@keyframes tm-import-glow` | one-shot | **still primary — blocking finding above** |

The three non-animated attention dots in `SideBar.css` (658, 1387, 1410) were already correct. Residual `v-theme-primary` in `StatsSidebar.vue` is otherwise out of population: sparkline fill (2125, 2292) and `.tm-comfy-fill` (2481), a progress-bar fill. So: five live-work indicators wore `primary` before this PR, four are changed here, one remains.

### Tokens are clean

All eight touched declarations use `rgb(var(--v-theme-accent))` / `rgba(var(--v-theme-accent), <a>)`. No hex, no `rgba(0,0,0,…)`, no new value, and the right token — `accent`, not `accent-bright`, which `main.js` declares "not a text/fill token". The `box-shadow` glows aren't elevation-token violations: §10 defines the attention pulse as a named hand-rolled pattern whose glow "lives with the pattern", and the radii and alphas carry over unchanged.

### Two pieces of documentation drift, neither caused by this PR

No action wanted on this branch; both cost the next reader.

- §12 ends "the stats toggle's attention dot moves to the ⋯ trigger while folded". The shipped code says the opposite — `TbGlobalActions.vue:96–99` cites amendment #2 in `toolbar-responsive-decisions.md`: "Settings and Stats never fold … The activity dot stays first-class on the Stats button at every width."
- §4 still documents `accent` as `#9e6727` (light) / `#b85c0c` (dark) and tabulates ratios from those values, while `main.js` ships `#c47a1e` for both. Checking this PR's contrast *from the manual* gives numbers that don't correspond to the product — §4's table would say 4.51:1 where the shipped value gives 2.93:1.

### Nothing in CI guards any of this

`grep` for the four changed class names across `frontend/src/**/*.test.js` and `frontend/e2e/` returns zero hits. No unit test, no e2e assertion, no visual-regression check — so a green gate says nothing either way here, and a future revert to olive would be equally silent. Not a request to add one; noted so the green tick isn't read as confirmation.

---

frontend/src/components/panels/StatsSidebar.vue
lindkvis: Arithmetic agrees with the subagent pass I ran before pushing (2.93:1 light, 4.34:1 dark). No change here: accent is the token §12 names, and the light-theme shortfall belongs to the whole attention-dot pattern (the SideBar.css dots too), so it wants the `--attention-dot` alias or ring as its own design call.

frontend/src/components/panels/TbGlobalActions.vue
lindkvis: Agreed, left out for the reasons you give (the 7px offsets are tuned to the 7px diameter). Best landed with the contrast follow-up as one geometry-plus-colour pass.

frontend/src/components/panels/StatsSidebar.vue
lindkvis: Only out of scope along with the size. Since it renders identically, it's in 87af8a44: `.tm-status-dot`, `.tm-tab-pulse` and `.tb-stats-btn--busy::after` now use `var(--radius-pill)`. `.legend-dot` stays as it is, since it's a chart legend swatch rather than an attention dot.

lindkvis:
Blocking finding fixed in 87af8a44: all five `primary` stops in `@keyframes tm-import-glow` (box-shadow at 0/35/100%, border-color at 0/35%) are now `accent`. The 100% resting border stays `on-surface`. The two doc-drift notes (§12's folded-⋯ sentence, §4's stale accent hexes) are left for a docs pass, as you asked.

frontend/src/components/panels/StatsSidebar.vue
lindkvis: Tracked in #1413, covering all the amber indicators including the stats-toggle icon, which now pulses whole since e4f33f07 and has the same 2.93:1 on the light toolbar.

frontend/src/components/panels/TbGlobalActions.vue
lindkvis: Since e4f33f07 the toolbar dot is gone: the stats toggle pulses its whole icon instead. The two sidebar dots that remain off `--badge-size-dot` are tracked with the contrast work in #1413.

PR #1386, opened 2026-09-16
Plugin pickers in Settings › Models are dropdowns

### LINKED ISSUE #1344: Design Cleanup: The plugin lists in Settings -> Models cannot be radio buttons
It takes up far too much space and looks busy. They are single selection and should just be drop down lists instead. Add a separate settings icon for configuring the current model.

Image

lindkvis:
Closes #1344.

The tag and description plugin pickers under Settings › Models › Auto-tagging were radio tables. Single selection doesn't need a row per plugin, so each is now a field that opens the app's menu (`.ctx-menu` rows, `menuitemradio`, olive check on the chosen one), with a gear beside it that configures the chosen plugin.

- Each plugin row, and the field, leads with a loaded icon, and hovering shows the plugin's description and whether it is loaded. So readiness can be compared without choosing one, which saves.
- Not a native `<select>`: its popup can take neither the menu styling nor a tooltip per option.
- `PluginsTable.vue` → `PluginSelect.vue`; the scroll-box CSS in `BehaviourSection.vue` that existed only for the tables is gone.
- The gear is `aria-disabled` with a guarded handler for None or a no-longer-installed plugin, so its tooltip saying why stays reachable.
- Choosing refocuses the field once the menu's close transition ends. A failed save is announced as an alert.
- A saved plugin that is no longer installed shows as "(not installed)", in the field and in the empty state.


frontend/src/components/widgets/PluginSelect.vue
```diff
@@ -0,0 +1,228 @@
...
+  }
+  return activePlugin.value
+    ? `${activePlugin.value} is not installed`
+    : "Choose a plugin to configure it";
+});
+
+// With nothing chosen this sets no plugin, so the dialog stays unrendered.
+function openSettings() {
+  dialogPlugin.value = activePluginInfo.value;
+  dialogOpen.value = true;
```
lindkvis, line 147: **Non-blocking — the gear is inert, but by coincidence rather than by a guard.**

`aria-disabled` doesn't stop activation; the button still receives click, Enter and Space, so the handler has to refuse the press itself. Here line 115 runs unconditionally. What actually makes the press harmless is `v-if="dialogPlugin"` on line 182 — `activePluginInfo` is `null` in exactly the states that set `aria-disabled` (line 172), so the dialog never mounts.

I traced the state machine and couldn't find a sequence where the dialog opens from a refused press, so **this isn't a live defect** — it's a coupling between two expressions that have to stay in lockstep. Two reasons to tighten it anyway:

1. It leaves `dialogOpen === true` after a refused press. Nothing reads it while `dialogPlugin` is null, so it's inert, but it's state the component believes is open when nothing is.
2. The house pattern is `aria-disabled` **plus** a guarded handler, and it's written down. `UndoControl.vue:15` says so in prose — "So `aria-disabled` + a guarded handler" — and `UndoControl.vue:179`/`:184` are the guards. `ReviewSessionView.vue:376` records the same rule. This is the one `aria-disabled` site in `frontend/src/` that marks a control refused without also refusing it in the handler.

```js
function openSettings() {
  if (!activePluginInfo.value) return;
  dialogPlugin.value = activePluginInfo.value;
  dialogOpen.value = true;
}
```

The two existing tests (`PluginSelect.test.js:158`, `:172`) keep passing; a third asserting `dialogOpen` is still false after a refused press would pin the guard rather than the `v-if`.

**Your tooltip justification checks out**, for the record: `Tooltip.vue:187–199` attaches `focusin` to the activator, and `AppButton` only sets native `disabled` from its own props (`AppButton.vue:15`), neither of which this call site passes — so a keyboard user tabs to the gear and the tip opens. `AppButton.vue:17` also makes the tooltip the accessible name for an icon-only button, which `PluginSelect.test.js:162`/`:176` assert.

---

frontend/src/components/widgets/PluginSelect.vue
```diff
@@ -0,0 +1,228 @@
...
+  const status = p.is_loaded ? "Loaded" : "Not loaded";
+  return p.description ? `${status} · ${p.description}` : status;
+});
+
+// A native <select> speaks strings, so None travels as "" and maps to null.
+const options = computed(() => [
+  { value: "", label: noneLabel.value },
+  ...capablePlugins.value.map((p) => ({
+    value: p.name,
+    label: p.display_name,
```
lindkvis: **Non-blocking — a plugin with no `display_name` renders as a blank option.**

No fallback here, and `display_name` is optional all the way down: `TaggerPlugin.display_name` defaults to `""` (`pixlstash/tagger_plugins/base.py:80`), and `GET /taggers` serves the raw attribute rather than the schema's `display_name or name` (`routes/taggers.py:155` vs `base.py:174`). A third-party plugin that doesn't set it arrives as `""`.

In the old table a blank name was a blank cell in a row that still had a radio, a Loaded indicator and a gear beside it — obviously a plugin. In a closed `<select>` there's nothing else: the field just reads empty, indistinguishable from a rendering fault. The gear's accessible name degrades with it, since line 105 would produce `" settings"`.

**This is the class rather than the instance.** Every other tagger-plugin label in the frontend already has the fallback: `SelectionBar.vue:43`, `SelectionMenu.vue:177` and `:211`, `TbTagPanel.vue:284`, `ImageOverlay.vue:112`, `OverlayDescriptionPanel.vue:81`, `OverlayTagsPanel.vue:76`, and `ImageGrid.vue:1871`. That's 8 sites with it and `PluginSelect` without. The gap was carried over from `PluginsTable` rather than introduced here, but this is the surface where it shows worst and the natural moment to close it.

**Fix:** `label: p.display_name || p.name` here, and `${activePluginInfo.value.display_name || activePluginInfo.value.name} settings` on line 105.

---

frontend/src/components/widgets/PluginSelect.vue
```diff
@@ -0,0 +1,228 @@
...
+
+<template>
+  <div class="plugin-select">
+    <div v-if="capablePlugins.length" class="ps-row">
+      <AppSelect
+        class="ps-field"
+        :model-value="activePlugin ?? ''"
+        :options="options"
+        :label="kind === 'tag' ? 'Tag plugin' : 'Description plugin'"
+        hide-label
```
lindkvis: **Non-blocking, but the most consequential of my notes — the save disables the `<select>`, which drops keyboard focus mid-interaction.**

`:disabled="settingActive"` puts the native `disabled` attribute on the `<select>` (`AppSelect.vue:34`) for the duration of the PATCH. A natively disabled control can't hold focus, so the browser moves focus to `<body>`. `AppButton` treats this as a bug worth solving and carries a refocus watcher for it (`AppButton.vue:108–122`: "A natively-disabled button cannot hold focus, so the browser drops focus to `<body>`, stranding a keyboard user…"). The same hazard is now on the select, unhandled.

It matters more here than on a button because of how a native `<select>` fires `change` from the keyboard: on a **closed** select, arrow keys move the selection and fire `change` on each step in Chrome and Firefox. A keyboard user arrowing from None past the first plugin to the second would fire a PATCH per step — activating plugins they're only passing over — and the first PATCH takes their focus away. A user who opens the popup first and commits with Enter fires one `change`, so this is per-browser and per-interaction-style rather than a certainty. I couldn't run a browser, so please treat the multi-save half as unconfirmed; the focus-drop half follows from the `disabled` attribute and isn't in doubt.

The disable isn't gratuitous — your comment on lines 92–94 is right that toggling it is what re-patches the `<select>`'s DOM value after a failed save, and `PluginSelect.test.js:135–143` covers that. So the fix needs to keep the toggle:

**Suggested:** mirror `AppButton`'s watcher — record whether the select owned `document.activeElement` when `settingActive` went true, and refocus on `nextTick` when it goes false. `AppSelect` would need a small `defineExpose({ focus })`, the way `AppButton.vue:136` already does.

The old radio table had the same `:disabled="settingActive"` shape, so the behaviour is inherited; what's new is that there's now one focusable control instead of a group, so losing it is more disruptive.

---

frontend/src/components/widgets/PluginSelect.vue
```diff
@@ -0,0 +1,228 @@
...
+
+const activePlugin = computed(() => props.settings?.[activeKey.value] ?? null);
+
+/** The chosen plugin's object, or null for None or a name no longer listed. */
+const activePluginInfo = computed(
+  () => capablePlugins.value.find((p) => p.name === activePlugin.value) ?? null,
+);
+
+/** "Loaded · what it does" for the chosen plugin; empty for None. */
+const hint = computed(() => {
```
lindkvis: **Design question, not a defect — other plugins' loaded state is no longer visible before you commit to one.**

The old table showed, for every capable plugin at once, its description (tooltip) and a Loaded / Not loaded indicator. The dropdown shows those for the **selected** plugin only (line 178). Since picking an option saves immediately (`setActive`, line 118), the only way to read another plugin's status is to activate it — browsing the list now mutates config.

Losing the per-plugin **description** to a tooltip-less `<select>` is unavoidable and I'd accept it; a description is a one-time "what is this" that the muted line answers well enough once chosen. The **loaded** state is different: it's the thing you want to compare across plugins before choosing, and it's cheap to keep:

```js
label: `${p.display_name || p.name}${p.is_loaded ? "" : " — not loaded"}`
```

The open dropdown then shows readiness for every plugin without a save, and `.ps-hint` keeps carrying the description.

Worth saying plainly: I think the dropdown is the right call overall. Two radio tables for a single-choice setting was the wrong control, and the explicit "None — no automatic tagging" first option (line 71) is a real improvement in legibility over a None *row*.

---

frontend/src/components/widgets/PluginSelect.vue
```diff
@@ -0,0 +1,228 @@
...
+        icon-only
+        icon-left="cog"
+        :tooltip="gearTooltip"
+        :aria-disabled="activePluginInfo ? undefined : 'true'"
+        @click="openSettings"
+      />
+    </div>
+    <div v-else class="ps-empty">No {{ kind }} plugins registered.</div>
+
+    <div v-if="hint" class="ps-hint">{{ hint }}</div>
```
lindkvis: **Non-blocking — two notes on the status line, both about it not reaching a screen-reader user.**

**It isn't announced when it changes.** `.ps-hint` is the only place the chosen plugin's loaded state now lives, and it changes as a *consequence* of the selection rather than as part of the control the user is on, so it won't be heard. Same for `.ps-error` on line 180 after a failed save — the field silently reverts and a message they never hear appears below. The repo uses live regions where this matters: 24 of the 80 components under `settings/` and `widgets/` carry `role="alert"` or `aria-live` (`LibrariesSection.vue`, `AddToEntityControl.vue`, `SearchResultBar.vue`). Suggest `aria-live="polite"` here and `role="alert"` on line 180.

**Success is silent, next to a control that isn't.** `setActive` surfaces failure via `.ps-error` but says nothing on success. In this same pane the VRAM slider — also an immediate-save control — sets `maxVramGbSuccess = "Saved."` and renders it beside the control (`BehaviourSection.vue:125`, `:313–314`), clearing it on a timer (`:130–132`). The picker now sits two sections above that and behaves differently.

The old radio table had no success affordance either, so this is carried over rather than introduced. Raising it because a dropdown makes it more acute: with a radio you can see your choice is checked; with a select the field shows the new value whether or not the PATCH landed, until the failure path corrects it. Either reuse the `"Saved."` pattern, or say on this thread that silent success is deliberate.

---

frontend/src/components/widgets/PluginSelect.vue
```diff
@@ -0,0 +1,228 @@
...
+      },
+    },
+  };
+  emit("update:settings", next);
+}
+</script>
+
+<template>
+  <div class="plugin-select">
+    <div v-if="capablePlugins.length" class="ps-row">
```
lindkvis: **Question — the empty state hides a saved-but-unavailable plugin.**

The "(not installed)" option (lines 78–85) is only reachable when `capablePlugins.length` is non-zero, because the whole `.ps-row` sits behind this `v-if`. If *every* capable plugin disappears — all failed to load, or the last one was uninstalled — while `active_tag_plugin` still names one, the user sees only "No tag plugins registered.", with no hint that a plugin is still configured and no way to clear it.

The old table behaved identically, so this isn't a regression, and the plugin load-error list in `BehaviourSection` covers the commonest cause. Asking rather than flagging: is it worth one extra muted line in the empty state naming the orphaned setting?

---

lindkvis (commented):
## Review — plugin pickers as dropdowns

**No blocking findings.** The change is careful: the failed-save restore is deliberate and tested, the `aria-disabled` gear really is inert, the CSS is entirely token-based, and the architecture docs were updated in the same change. Seven inline notes, all worth doing but none of which breaks a user on the happy path. The one I'd actually act on before merge is the focus drop on `PluginSelect.vue:163`.

I agree with the control change. Two radio tables for a single-choice setting was the wrong shape, and the explicit "None — no automatic tagging" first option reads better than a None *row* did.

### One note that can't anchor inline

**`.behaviour-pane` is now a class with no rule anywhere.** The hunk at `BehaviourSection.vue:430–440` removes `.behaviour-pane { display: flex; flex-direction: column; height: 100% }`, correctly, since nothing inside the pane scrolls independently now. But `<div class="behaviour-pane">` stays at line 272 (untouched, hence not inline-able) and is the only occurrence of the name left on the branch — `git grep behaviour-pane` returns 1 hit, the markup. No stylesheet, test or e2e selector uses it.

The layout consequence is fine: `.settings-content` in `UserSettingsDialog.vue:394–400` sets `height: 524px; overflow-y: auto`, so the pane scrolls as a whole, and both `SettingsFieldBlock`s pass `top`, so `field-block__control`'s `margin-top: auto` bottom-pinning isn't in play. So this is tidiness, not a regression — drop the class, or leave a line saying it's kept as a test hook.

### Token discipline — clean

New CSS in `PluginSelect.vue` (192–228) and the rewritten block in `BehaviourSection.vue` (433–438): no hardcoded hex, no `rgba(0,0,0,…)` shadow, no ad-hoc radius, no raw `px`/`em` font size. Spacing is `--space-2`/`--space-3` only; type is `--text-xs` and `--text-sm`.

Worth noting the new code is *better* than its neighbours here: secondary text uses `rgba(var(--v-theme-on-surface), var(--opacity-text-secondary))` (the token, asserted by `design-tokens.test.js:55`), while `.settings-tagger-loading` and `.settings-tagger-plugin-dir` in the same file still use a raw `0.55` and `field-block__desc` a raw `0.6`. No change requested — just recording which side of the drift this lands on.

The gear moving from `ghost`/`sm` to `variant="secondary"` at default `md` gives it `--control-h`, the same height as `AppSelect`'s field (`AppSelect.vue:149`), so the row aligns. Reads deliberate and correct for a control paired with an input.

### Checks that came back clean

- **Props/emits.** `kind` has a validator, `plugins` and `settings` have factory defaults, `update:settings` is the only emit and is declared. `v-model:settings` at both call sites matches. `@update:model-value` → `update:modelValue` is correct — the compiler camelizes the handler key and `AppSelect` declares it.
- **Failed-save restore is real, not wishful.** Vue force-patches `value` on every props patch regardless of change, and toggling `disabled` is what triggers it.
- **`onParamsSaved`** is a faithful copy of the old immutable deep-merge; no shared-reference mutation of `props.settings`.
- **Scroll-box CSS removal, grep counts.** `.tagger-cols` 0 hits, `.tagger-section` 0 hits, `field-block__control` 4 hits all inside `SettingsFieldBlock.vue` itself. `PluginsTable`/`TagPluginsTable`/`DescriptionPluginsTable` 0 hits anywhere on the branch, and the whole `.pt-*` class family 0 hits. Nothing was relying on the removed `:deep` override and no e2e spec referenced any of them.
- **Test coverage moved rather than dropped.** `PluginsTable.test.js` had 9 tests; `PluginSelect.test.js` has 14. Every old case has a counterpart. The one without is "re-selecting the active plugin keeps it selected", which was specific to a radio firing `change` on a re-press — a native `<select>` can't fire `change` for the value it already holds, so dropping it is right. New coverage with no old equivalent: the "(not installed)" option, the failed-save restore, the gear's disabled states, and the status line.
- **CI gating.** Frontend tests run as `vitest run` over a glob, so the new file is gated automatically; the `test_ci_shards.py` file-list rule is Python-only and isn't engaged.
- **Docs updated in the same change** — the three table entries replaced by a `PluginSelect.vue` entry, the `BehaviourSection.vue` layout paragraph rewritten to match the removed CSS, and `backend_architecture.md:1313` picked up the rename. `changelog.d/1344.md` exists and describes a genuinely user-visible change.

### Not verified

`frontend/node_modules` isn't installed in the review environment, so neither `vitest` nor the app was run. Specifically unconfirmed: whether arrowing a *closed* `<select>` fires one `change` or one per step, which decides whether the `:163` note means one stray save or several. Nothing was rendered, so the layout-shift nit (`.ps-hint` is `v-if`'d, so choosing None removes a line and the content below jumps up) is derived from markup and may well be imperceptible.

One comparison is still owed: `frontend_architecture.md` §7 points at a published design system readable through `DesignSync`, with `ui_kits/app/` holding specs for the app's screens. That tool wasn't available here, so whether a Settings › Models spec exists there and whether this picker matches it hasn't been checked.

---

frontend/src/components/widgets/PluginSelect.vue
lindkvis: Fixed in ee171d56: `openSettings` returns early when no plugin is chosen, so a press it refuses no longer leaves `dialogOpen` true. A new test checks `dialogOpen` after a refused press. It went red with the guard removed.

frontend/src/components/widgets/PluginSelect.vue
lindkvis: Fixed in ee171d56: both the option label and the gear's name now use `display_name || name`. A new test covers a plugin with an empty `display_name`, and it went red without the fallback.

frontend/src/components/widgets/PluginSelect.vue
lindkvis: Fixed in ee171d56, following AppButton's approach. `AppSelect` now exposes `focus()`. `setActive` records whether the select had focus before disabling it and puts focus back on `nextTick` once the save settles. The disable stays, since it is still what puts the dropdown back after a failed save. The test blurs the select while the save is in flight and asserts focus comes back; it went red without the refocus. Still unconfirmed, as you said: whether arrowing through a closed select fires one save per step. The disable during each save at least keeps them from overlapping.

frontend/src/components/widgets/PluginSelect.vue
lindkvis: Taken in ee171d56: a plugin that isn't loaded now reads "<name> — not loaded" in the list, so you can compare readiness without choosing one. Loaded plugins show just the name. The line under the dropdown still carries the description.

frontend/src/components/widgets/PluginSelect.vue
lindkvis: Partly done in ee171d56: the status line is now `aria-live="polite"` and stays mounted (hidden with `:empty` when blank), so it can announce, and the error line is `role="alert"`. Both are tested. I left success silent on purpose: the dropdown already shows the new value, and a failure both puts it back and is announced, so a "Saved." line would only confirm what is already on screen. Leaving this open in case you'd still like the VRAM slider's pattern here.

frontend/src/components/widgets/PluginSelect.vue
lindkvis: Yes, done in ee171d56: when no plugins of that kind are installed but one is still saved, the empty state now reads "No tag plugins registered. <name> is still selected." It still can't be cleared from there, since there's no control to clear it with. Tested.

lindkvis:
Two notes from the review body, both in ee171d56:
- `.behaviour-pane`: removed. Nothing else used the class.
- Design system: I checked the Claude Design project before the first push. `ui_kits/app/` has no Settings or Models spec, and the forms kit has no select, so `AppSelect` is the house field to use here.

frontend/src/components/widgets/PluginSelect.vue
lindkvis: Following up after a62522bb, which replaced the native `<select>` with the app menu. Two parts of this no longer apply:
- **Status line:** `.ps-hint` is gone. Loaded state is now an icon on the field and on each row. The field's accessible name carries it ("Tag plugin: WD14, loaded"), and choosing returns focus to the field, so a screen reader hears the new value and its state there.
- **Field running ahead of the save:** the field renders the saved setting rather than the browser's own control state, so it changes only once the PATCH succeeds. A failure never moves it and is announced through `role="alert"`.

So a "Saved." line would repeat what the field already shows. I'm keeping success silent on purpose, as the thread allowed, and resolving.

PR #1387, opened 2026-09-16
Filter menu as submenus that fill a filter strip

lindkvis:
Builds the approved **Filter Submenus** design for the toolbar funnel. The funnel icon is unchanged.

**What changed**
- The funnel now opens a root menu of filter kinds: Picture (Media, Faces, Stacks, Sharing), Content (Score, Tags, Tag confidence, ComfyUI model, LoRA) and Problems. Each kind opens its own `.tbm` menu beside the root. The menus stay open while you choose; Esc, a click outside or the funnel closes them. This replaces `GbFilterPanel`.
- A new filter strip under the toolbar shows one chip per active filter, each with an ×, plus "N of M" and Clear all. It disappears when the last chip goes. Chips and the menu's labels share `utils/filterChips.js`.
- Row counts are `GET /pictures/count` over the grid's view plus that one filter (`useFilterCounts`, `buildFilterCountBaseQuery`), so no new endpoint was needed.
- Score is At least / At most plus Include unscored. In the backend, `PredicateFilter` now ORs `unscored` with a min/max range instead of ANDing them, and the store no longer makes the two exclusive. The stats sidebar's bars and the agreement matrix clear the other kind themselves, so they still pick one thing.
- The toolbar's count badge is gone, since the strip shows every filter.

**Where this differs from the design**
- **Sharing** offers only *Shared*. *Not shared* needs an exclusion the listing cannot express on every path (unassigned, set, likeness, smart-score candidates), and a filter that is silently ignored on some views would be worse than not having it.
- **Impossible combinations** lists the two cases the backend detects (face tags with no face, people tags with no humans). The design's indoors/outdoors, day/night and solo/group cases don't exist in the app, and adding them is tag-ontology work.
- **Tag counts** come from `/tags` and cover the whole library, not the view: one request instead of one per tag.
- **No counts during a search**, because `/pictures/count` cannot count search hits. A blank beats a wrong number.
- `OptionRows` is documented as "sort by only". The design explicitly uses option rows here, so the doc now names the filter menu as the second use.

**Pre-push review findings I left as they are**
- The Model and LoRA menus fire one count per visible row (at most 50) and cache them until the menu reopens. That is acceptable for typical model lists; batch it if it shows up.
- Row counts ignore the other active filters. That is on purpose: each count is "this choice on its own". The changelog now says so.
- The Score rows are separate Tab stops rather than one roving radiogroup. Worth a follow-up.
- The panel widths (320/280px) and caps match the existing toolbar panels (the Show menu is 320px). They are not new design values.
- There are no dedicated tests for `FilterStrip.vue` itself. Its logic (`filterChips`) and the menu rules are tested.


frontend/src/components/panels/FilterMenu.vue
```diff
@@ -0,0 +1,742 @@
...
+  const v = store[def.field];
+  return v === def.off ? null : v;
+}
+
+function setPick(kind, id) {
+  const def = PICK_ONE[kind];
+  store[def.field] = id == null ? def.off : id;
+}
+
+// Arrows and clicks both select (update fires); a click on the row that was
```
lindkvis: **Blocking — arrowing to a pick-one option makes the next Enter a dead keystroke.**

The `justChanged` latch is built for the click path, where `update:modelValue` and `pick` fire in one synchronous `select()` call. But the two emit sites don't agree:

```js
// OptionRows.vue
function select(option) {
  if (option.id !== props.modelValue) emit("update:modelValue", option.id);
  emit("pick", option.id);                     // always
}
function onKeydown(event) {
  const id = arrowStep(event, props.options, props.modelValue);
  if (id !== undefined && id !== props.modelValue) emit("update:modelValue", id);
}                                              // no pick, ever
```

So an arrow press sets `justChanged` and nothing ever clears it.

Repro, in any of the four pick-one submenus (Media, Faces, Stacks, Sharing):

1. Open Stacks, nothing selected.
2. Press Down → `update:modelValue` → `justChanged = "stacked"`, filter applied, chip appears.
3. Press Enter on that same row → `select()` sees `option.id === modelValue`, so no `update` — only `pick` → `repick` finds `justChanged === "stacked"`, clears the latch, returns **without** calling `setPick(kind, null)`.

The keyboard user's natural "arrow to it, Enter to confirm" gesture does nothing at all, and the *second* Enter is what toggles the filter off — the opposite of what they were reaching for. Arrow-then-click behaves the same way.

It's silent, it only affects keyboard users, and it makes the widget feel broken rather than wrong. `FilterMenu.test.js:135` exercises click-then-click, which is exactly the path the latch handles correctly, so nothing catches it.

**Suggested fix** — the latch only needs to survive one synchronous emit pair:

```js
function changePick(kind, id) {
  justChanged = id;
  queueMicrotask(() => { justChanged = null; });
  setPick(kind, id);
}
```

The click path sets and consumes it within the same synchronous `select()`, so the microtask always runs after. Alternatively give `OptionRows` a `deselectable` prop so the toggle-off rule lives with the widget and every future consumer inherits it.

Worth a test for arrow-then-Enter alongside the existing click-then-click.

---

frontend/src/components/panels/FilterMenu.vue
```diff
@@ -0,0 +1,742 @@
...
+          <button
+            class="tbm-ghost"
+            type="button"
+            :disabled="!scoreActive"
+            @click="clearScore"
+          >
+            Clear
+          </button>
+        </div>
+        <div class="tbm-section" role="radiogroup" aria-label="At least">
```
lindkvis: **Blocking — the Score submenu declares `role="radiogroup"` but implements none of the contract.**

Two hand-rolled radiogroups ("At least" here, "At most" at :167), six `role="radio"` buttons each, with `aria-checked` but **no `tabindex` management and no arrow-key handling**. That's 12 sequential tab stops, and the arrow keys a keyboard or screen-reader user will reach for do nothing.

This isn't just a WAI-ARIA-practices point — it contradicts this repo's own written contract. `docs/frontend_architecture.md` §5, "Fields and pick-one":

> Both are `radiogroup`/`radio` with `aria-checked` and one roving tab stop moved by the arrow keys (`utils/radioGroup.js`).

A user who has learned the arrow keys from `Segmented`, from `OptionRows` in the Sort menu, and from the agreement matrix in this PR's *other* file — `StatsSidebar.vue:1034–1040` implements a roving tabindex so "20 cells don't cost 20 tab presses" — will find them inert here.

I've read your note that this is a known follow-up, and I'm disagreeing with that specific call rather than overlooking it. Two reasons it should land in this PR: the role attributes already *promise* the behaviour, so shipping as-is is a stated-but-unimplemented contract rather than a missing feature; and the machinery already exists, so the fix is small.

**Cheapest first:**

- Use `OptionRows` for both groups, with the star row in the `#meta` slot or as the `label`. That's one component for all three pick-one shapes in this menu and it brings `radioGroup.js` with it.
- If the star rendering can't fit `OptionRows`, call `arrowStep(event, options, value)` on the group's `@keydown` and `tabStopId(options, value)` for `:tabindex` directly — the two exports in `utils/radioGroup.js` are exactly this case.

The `At most` rows already carry native `disabled`, which both `arrowStep` and `tabStopId` skip, so that part needs no extra work.

---

frontend/src/utils/filterChips.js
```diff
@@ -0,0 +1,229 @@
...
+ * @param {{ allPicturesView?: boolean }} [view] "No character" only exists in
+ *   All Pictures, so a stale flag elsewhere is not shown as a filter.
+ * @returns {Array<{ key: string, kind: string, value: string, remove: Function }>}
+ */
+export function filterChips(store, { allPicturesView = true } = {}) {
+  const chips = [];
+  const push = (key, kind, value, remove) =>
+    chips.push({ key, kind, value, remove });
+
+  if (allPicturesView && store.unassignedOnlyFilter) {
```
lindkvis, line 128: **Blocking — the funnel can stay lit with an empty strip and no way to clear it.**

`filterChips` correctly hides the "No character" chip outside All Pictures (`allPicturesView && store.unassignedOnlyFilter`), but `filterStore.isActive` (`useFilterStore.js:95`) counts `unassignedOnlyFilter` unconditionally, and `clearAll()` (`FilterMenu.vue:533`, `FilterStrip.vue:87`) only iterates chips. The two disagree.

Repro:

1. In All Pictures, tick **Problems → No character**. Chip appears, funnel lit.
2. Select any **set** in the sidebar. `handleSelectSet` sets `selectedCharacter = null` but — unlike `handleSelectCharacter`, which clears the flag at `useAppNavigation.js:124–126` — leaves `unassignedOnlyFilter` set.
3. Now: no chip, so the strip doesn't render at all; the funnel is still lit; the Problems submenu's "No character" box is disabled and shows unchecked (`FilterMenu.vue:239–241`); and its **Clear is disabled**, because `problemsActive` requires `allPicturesView` (`:627–631`). No control anywhere turns it off.
4. Navigate back to All Pictures and the filter silently returns.

The flag doesn't change results in a set view (`_appendSelectionParams` takes the set branch), so this is an indicator and reachability bug rather than a wrong-results one. But it's a **regression from what this PR replaces**: the old funnel badge showed a count for it, and `GbFilterPanel`'s reset was gated on `filterStore.isActive`, so it was always clearable.

**Suggested fix:** now that `filterChips` is the single source of truth for "which filters are on", derive the funnel's lit state from it rather than from `isActive` — pass the same `allPicturesView` down and use `chips.length > 0` at `Toolbar.vue:230`. That removes the class of divergence rather than this one instance. Clearing the flag in `handleSelectSet` too would patch this instance, but leaves `isActive` and `filterChips` free to drift again.

---

pixlstash/utils/query/predicate_filter.py
```diff
@@ -303,12 +304,18 @@ def predicates(self) -> list[ColumnElement]:
...
+            score_range.append(Picture.score >= self.min_score)
         if self.max_score is not None:
-            preds.append(Picture.score <= self.max_score)
+            score_range.append(Picture.score <= self.max_score)
         if self.unscored:
-            preds.append(or_(Picture.score.is_(None), Picture.score == 0))
+            # With a range, unscored is added to it ("3-4 plus the unrated"),
+            # alone it is the unrated only.
+            unrated = or_(Picture.score.is_(None), Picture.score == 0)
+            preds.append(or_(and_(*score_range), unrated) if score_range else unrated)
```
lindkvis, line 316: **Non-blocking, but a contract question worth deciding rather than inheriting — `min_score=0` and `max_score=5` beside `unscored=1` now widen to the whole view.**

The AND→OR swap is right for the star values the new menu offers. But two spellings that are already legal on the API change meaning in a direction that's probably not intended:

```
# before (AND)
min_score=0 & unscored=1  ->  score >= 0 AND (score IS NULL OR score = 0)   -> score 0 only
max_score=5 & unscored=1  ->  score <= 5 AND (score IS NULL OR score = 0)   -> score 0 only

# after (OR)
min_score=0 & unscored=1  ->  (score >= 0) OR (score IS NULL OR score = 0)  -> every picture
max_score=5 & unscored=1  ->  (score <= 5) OR (score IS NULL OR score = 0)  -> every picture
```

`score >= 0` and `score <= 5` cover every rating that exists, so ORing the unrated in makes the score filter a no-op rather than a narrowing. `max_score=5` isn't hypothetical: `tests/test_predicate_filter.py:200–207` documents it as *"the 'any rating' spelling the sidebar sends"*.

The new Filters menu can't produce it — `STAR_ROWS` is `[null,1,2,3,4,5]` so `min_score` is never 0, and the sidebar clears `unscored` when it sets a range — so this is reachable via the API, an older client, or a bookmarked URL rather than through the new UI.

Not necessarily a defect: "any rating or unrated = everything" is arguably the consistent reading of the new rule. It's worth a deliberate decision plus one test pinning it, because it's the one input where old and new differ by the whole library instead of by a handful of rows.

**Suggested:** add `PredicateFilter(min_score=0, unscored=True)` to the new test with the expected set written out, and a clause in the `unscored` description at `routes/pictures/_listing.py:279–285` saying a range that already spans every rating plus `unscored` is everything.

**Separately, the grouping is correct** — I checked it specifically, since an OR spliced into an AND chain is the classic bug here. `or_(and_(*score_range), unrated)` is built from ORM column expressions, not `text()`, so SQLAlchemy calls `self_group(against=and_)` on each ANDed element and a `BooleanClauseList` with an `or_` operator returns a `Grouping`. It renders with the outer parens present. The manual-paren warnings elsewhere in this file (`:349–355`) are explicitly scoped to `text()` fragments, which is what makes this construction safe. Two sibling clauses in the same ANDed list already depend on that behaviour under green tests (`:213`, `:485`).

---

frontend/src/composables/useFilterCounts.js
```diff
@@ -0,0 +1,62 @@
...
+
+  watch(baseQuery, reset);
+
+  /**
+   * The count for one extra filter, or `undefined` while it loads or failed.
+   * Safe to call from a template: the request starts outside the render.
+   *
+   * @param {string} extra pre-encoded filter params, "" for the view's total
+   */
+  function count(extra) {
```
lindkvis, line 67: **Non-blocking — count requests are never cancelled or capped, and a failure is cached forever.**

The correctness half is handled and I want to say so first: the cache is keyed by the exact encoded filter string, and `generation` (`:19`, `:39`, `:45`) discards any response resolving after a `reset()`, so **a stale count cannot land on the wrong row.** That was the race I went looking for and it isn't there.

What remains is resource use:

1. **No cancellation.** `getPictureCount` (`api/pictures.js:82`) takes no `AbortSignal`, so nothing stops in-flight requests when the menu closes or the view changes — `generation` only stops them *writing*. With `MAX_SHOWN = 50` in `FilterChecklistMenu.vue:66` and `count-for` invoked per rendered row, opening Model or LoRA fires up to 50 `GET /pictures/count` in one microtask batch with no concurrency limit. Closing the funnel immediately stops none of them.
2. **Typing multiplies it.** Narrowing the field renders a different 50 rows, each asking for its own count. `requested` dedupes by query string, so the ceiling is the number of distinct models rather than 50 — but it's reached by typing rather than by a deliberate action.
3. **Every reopen pays again.** `reset()` on `props.open` (`FilterMenu.vue:537–548`) empties the cache, so closing and reopening re-issues the whole batch even when the view hasn't changed.
4. **A failure is sticky.** The `.catch` at `:47–50` logs but doesn't remove the key from `requested`, so a count that fails once shows no number until the next `reset()`.

In order of value for effort: key the cache on the base query and reset only when the base query actually changes (fixes 3, and makes 1 mostly moot); cap concurrency (a queue of 4–6, or fetch only for rows in the viewport); delete the key in `.catch` so a later render retries (fixes 4).

I've left this non-blocking because I couldn't measure what one `/pictures/count` actually costs. The backend lane's answer sharpens it: on the **default grid** the count is one `SELECT count(picture.id)` on its own connection with no ORM rows loaded — genuinely cheap. But on three branches it goes through `run_task`, i.e. queued on the **single serialised DB worker thread**: `character_id=UNASSIGNED`, `only_deleted`, and `sort=CHARACTER_LIKENESS` — which takes *two* queued tasks per request. The frontend does put `sort=CHARACTER_LIKENESS` into the count base query when that sort is active (`useGridFetch.js:468–477`), so that path is live, and 50 rows becomes 100 queued tasks sharing a thread with background tagging and embedding writes.

Worth skipping the per-row counts when `sort=CHARACTER_LIKENESS` is in the base query — the numbers are identical to the unsorted ones for every filter except the character facet.

---

frontend/src/components/panels/FilterMenu.vue
```diff
@@ -0,0 +1,742 @@
...
+  const n = chips.value.filter((c) => KIND_OF_CHIP[c.kind] === kind).length;
+  return n || "";
+}
+
+function openKind(kind) {
+  sub.value = kind;
+  const row = rowRefs[kind];
+  // Line the submenu's top up with the row that opened it.
+  subTop.value = row ? Math.max(0, row.offsetTop - 8) : 0;
+  nextTick(() => {
```
lindkvis, line 540: **Non-blocking — opening a submenu can put focus on "Clear".**

`querySelector` with a selector list returns the first element in **document order** that matches *any* selector — it doesn't try the selectors in the order written. The header's "Clear" button (`.tbm-ghost`) comes before the option rows in every submenu, so whenever that kind already has a filter on (i.e. Clear is enabled), focus lands on Clear and the next Enter wipes the kind.

**Fix:** try the selectors in sequence, or scope to the body — `subRef.value?.querySelector('.tbm-section')?.querySelector(...)`.

Smaller, related: the `document.activeElement?.tagName !== "INPUT"` guard on the next line is effectively dead. At the moment it runs, `activeElement` is still the row button; the search field in `FilterChecklistMenu`/`FilterConfidenceMenu` is focused afterwards by each component's own `onMounted` → `nextTick`, which is queued later. It works out, but the guard isn't what makes it work.

---

lindkvis (commented):
## Review — filter submenus and the filter strip

Reviewed in two lanes, frontend and backend, merged here. **Three blocking findings, all frontend; the backend change is sound.** Six inline notes plus the items below.

### Cross-lane reconciliation, since the frontend change depends on it

The store change and the chip label "3+ or unscored" rest on the backend actually ORing `unscored` with the range. It does, and the clause is correctly grouped against the surrounding AND chain — checked specifically, because an OR spliced into an AND chain without its own group is the classic bug in this shape. Details in the inline note on `predicate_filter.py`. So the two halves of this PR agree.

The backend lane found **no blocking issues**: no route added, no authz file touched, `("GET", "/api/v1/pictures/count")` unchanged at `registry.py:841`, and the new usage is more calls to an existing route with the same parameter vocabulary, so it opens no new input space. The caller enumeration for the AND→OR change came back clean too — `PredicateFilter` has 8 construction sites, 6 of which can receive `unscored` with a range, but under the old AND the result was empty for every combination except `min_score=0`/`max_score∈{0..5}`, and "nothing" is not a result any caller can have relied on. The one substantive change is the inline note. Two sites confirmed deliberately unaffected: `picture_stats._agreement_scope` rebuilds its scope with the score fields cleared, and `fetch_smart_score_candidate_ids`'s early-return tests for presence rather than semantics.

Worth naming one caller explicitly: **`GET /pictures/export`** uses the same parser, so an export with `unscored=1&min_score=N` now writes out the union rather than the empty set.

### Blocking, in summary

1. **Arrow-then-Enter is a dead keystroke** in all four pick-one submenus (`FilterMenu.vue:570`). The `justChanged` latch assumes `update:modelValue` is always followed by `pick`; `OptionRows.onKeydown` never emits `pick`.
2. **`role="radiogroup"` with none of the contract** on the Score submenu (`:138`) — 12 tab stops, inert arrow keys, against a documented repo rule and a roving-tabindex implementation in this PR's own other file.
3. **The funnel can stay lit with an empty strip and no way to clear it** (`filterChips.js:108`) — `isActive` and `filterChips` disagree about `unassignedOnlyFilter` outside All Pictures.

### `OptionRows` is widened past its documented rule with no doc change

No diff line to anchor this to. `docs/frontend_architecture.md` §5 still reads "`OptionRows` for **sort by and only sort by**". On `develop` there are exactly three call sites and all three are sort pickers (`Toolbar.vue:114`, `ShelfSortPanel.vue:32`, `WorkflowShelf.vue:118`). This PR adds three non-sort ones (`FilterMenu.vue:102`, `FilterConfidenceMenu.vue:38` and `:50`), doubling the population and inverting the rule.

The widening is the right call — the filter pick-ones are structurally identical to a sort pick-one, and inventing a fourth shape would be worse. The problem is only that the rule now says one thing and the code does another, and CLAUDE.md asks for the doc update in the same change. Two short edits: the §5 "Fields and pick-one" paragraph, and the `OptionRows.vue:36–37` docstring, which repeats the old rule verbatim.

I note the PR body says `OptionRows` is documented as "sort by only" and that the doc now names the filter menu as a second use — that edit doesn't appear to have landed in §5 or the docstring.

### On the panel widths

The description says 320/280px match the existing toolbar panels. Checking every one on `develop`: `TbExportPanel` 392px, `TbTagPanel` min 340px, `TbOverflowMenu` min 220px, the view panel 264px — and `GbFilterPanel`, the panel being replaced, 376px. `.tbm` itself sets no width. So these are the fifth and sixth numbers rather than two of the existing four.

That may well be fine — two menus side by side have a different width budget from one — but it's a design decision rather than a match, and `design-tokens.css` is explicit about this failure mode: `--stats-panel-w`'s comment says naming a width is what "stops a fifth call site picking a fifth number". Either reuse 264/340, or name these as tokens so the seventh number has somewhere to go.

### Smaller notes

- **The submenu footer always names the first option** (`FilterMenu.vue:114–117`). With "Video" selected, Media still reads `On the strip as "Media images"` — the one thing on screen contradicting the chip the user can see.
- **The checklist's keyboard highlight can move off-screen** (`FilterChecklistMenu.vue:111–119`, `:128–132`). `move()` never scrolls; ~11 rows fit in the 320px list, so Down past that ticks a row the user can't see. Needs `scrollIntoView({ block: "nearest" })`. For screen readers the highlight is also invisible — no `aria-activedescendant`.
- **The same widget selects on arrow in one submenu and not the next.** `FilterConfidenceMenu.vue:50–60` binds `:model-value` and `@pick` but no `@update:model-value`, so arrows move focus and Enter applies; one section up, the Tag rows use `v-model`. Manual activation is defensible here — just make it explicit rather than implied by a missing listener.
- **Two non-modal `role="dialog"` elements side by side, no focus trap.** No other `.tbm` panel declares a role. `role="group"` with the same `aria-label` matches what the markup actually is. The `aria-expanded`/`aria-controls` wiring is good and should stay.
- **`.fm-check--kbd` draws something that looks exactly like the focus ring** (`App.css:1040–1042`) — same colour and width as the global `:focus-visible`, minus the gap, on an element that isn't focused. A keyboard user sees two focus rings.
- **The "of N" label is computed and fetched twice** — `FilterStrip.vue:57–85` calls `getPictureCount` directly with its own sequence guard while `FilterMenu.vue:463–473` goes through `useFilterCounts`. Same query, two requests, two copies of the formatting, and they can disagree mid-flight.
- **Chips scroll horizontally with the scrollbar hidden** (`FilterStrip.vue:113–114`). Nothing indicates more chips exist off-edge. `Clear all` is correctly pinned so it never scrolls away.
- **Smart score and Resolution are strip-only** — chips and `Clear all` cover them, but neither has a row in the menu, so once removed they can only be set again from the stats sidebar. Reads deliberate; worth a comment in `filterChips.js` so the next reader doesn't hunt for the missing rows.
- **`activeCount` now has no consumer** (`useFilterStore.js:99–125`) — the funnel badge was its last one. Its test still explains itself with "miss activeCount and the toolbar badge undercounts", naming a badge that no longer exists.

### What holds up

Stated so it doesn't have to be re-derived: `buildFilterCountBaseQuery` really is filter-free and really does match the grid (`stack_leaders_only=true` matches what the grid stream sends, and the `unassignedOnly: false` fall-through re-appends `project_id`, so dropping "No character" doesn't drop the project scope). The `--filter-strip-h` sweep is complete — five consumers of `--selbar-height`, the four below the strip all updated, the fifth being the strip's own `top`. The stats-sidebar claim checks out: `isAgreementCellActive` excludes `unscoredOnly` and `onAgreementCellClick` clears it before setting the range. `useGridScoring` matches the new store semantics in both directions. No leftover references to `GbFilterPanel` anywhere.

### Test coverage

`FilterMenu.test.js` pins six real behaviours, including that each row's count is the view plus exactly one filter with the other active filters asserted *out*. `filterChips.test.js` covers construction, ordering and the `allPicturesView` gate. Deleting `GbFilterPanelScope.test.js` is right — it tested native `<select>` wiring that no longer exists.

Two gaps worth closing:

1. **`FilterStrip.vue` has no test and it owns a race guard.** The `request` sequence number at `:70–77` is the kind of code that's written once, is never obviously wrong, and silently stops working. One mount test with two overlapping `countBaseQuery` changes, asserting the earlier response is discarded, covers the only non-trivial logic in the file.
2. **The `off` sentinel is tested for one of four pick-one kinds.** The deleted test's most valuable case guarded `null` being coerced. The three sentinels are now `"all"` (media, stacks), `null` (faces) and `false` (sharing), and `FilterMenu.test.js:135` only exercises `"all"`. Parameterising over all four restores the invariant that file existed to protect — and it's the same test that can't currently see blocking finding 1.

On the backend side: no new test file, so the gating rule is satisfied by construction. But `test_predicate_filter.py` is listed in `DEFERRED_FROM_GATE` (`tests/test_ci_shards.py:230`), so the one test covering the semantics this PR changes runs only in the informational sweep — a later revert of the OR wouldn't turn a PR red. It's a good promotion candidate: it stands up no `Server`, just `create_engine` over a `tmp_path` SQLite file. Alphabetical slot is between `test_plugin_lifecycle_gaps.py` and `test_project_membership_service.py`. Pre-existing deferral rather than something this PR created, so saying "out of scope" is a fine answer.

### Not verified

Nothing was executed — no vitest, no linters, no app, no backend suite. Blocking findings 1 and 3 and the focus note are reasoned from emit order, store/selector divergence and `querySelector` semantics respectively; 1 I re-confirmed against `OptionRows.vue:66–78` directly. The backend parenthesisation conclusion is read off the construction plus SQLAlchemy's `self_group` precedence rule, corroborated by two sibling `or_()` clauses already under green tests; one `print(stmt.compile())` would settle it outright. No timing figures: every cost statement is read off query construction and the task-runner code, not measured.

`GbFilterPanel.vue` (1332 deleted lines) was read only for its root template and width declarations — **nobody has verified that every behaviour it had has a successor in the new menus.** One concrete difference did surface that way: its reset called `filterStore.resetFilters()` where the new `Clear all` walks the chip list, which is blocking finding 3. A deliberate old-to-new feature diff before merge would be worth it.

Also unassessed: visual rendering in either theme, the two-menu cascade at phone width, and whether focus returns to the funnel when the menu closes with focus inside an unmounting submenu.

---

frontend/src/components/panels/FilterMenu.vue
lindkvis: Fixed in 72ba559c, differently from the suggestion. Your microtask latch would make arrow-then-Enter turn the filter **off**, which is still the opposite of confirming. The design says removing the chip is how you get back to Any, so I removed the click-again toggle altogether. The pick-one lists now bind only `update:modelValue`: arrows select, and Enter or a click on the chosen row confirms it. Clear or the chip's × removes it. The confidence thresholds follow the same rule. `FilterMenu.test.js` now runs arrow → Enter → Clear for all four kinds, one case per "off" sentinel (`"all"`, `null`, `false`), and reverting to the old pick-toggle turns 4 tests red.

frontend/src/components/panels/FilterMenu.vue
lindkvis: Fixed in 72ba559c. Both Score groups now use `utils/radioGroup.js`: `tabStopId` gives each group one tab stop and `arrowStep` on the group's keydown selects and moves focus. At most rows below the minimum are `disabled` options, so both helpers skip them. I kept the hand-rolled rows because the star rendering doesn't fit `OptionRows`' label. A new test asserts one tab stop and that two ArrowDowns select 1 then 2. Removing the arrow handler or the tabindex turns it red.

frontend/src/utils/filterChips.js
lindkvis: Fixed in 72ba559c, as you suggested, at the class level. `Toolbar.vue` now lights the funnel from `filterChips(store, { allPicturesView }).length`, the same test the strip uses. `isActive` had no other consumer, so it's gone, and so is `activeCount` (no consumer since the badge left). In your repro the leftover flag now does nothing visible in the set view: no chip and no lit funnel. Back in All Pictures its chip is there again, so it isn't silent and × clears it. New Toolbar tests cover both directions, and lighting the funnel from the flag again turns them red.

pixlstash/utils/query/predicate_filter.py
lindkvis: Decided rather than inherited, in 72ba559c: a range spanning every rating (`max_score=5` or `min_score=0`) plus `unscored` means every picture, the consistent reading of "range or unrated". Pinned by `test_unscored_with_a_range_spanning_every_rating_is_everything`, and the `unscored` description in `_listing.py` now says so. Thanks for the grouping check. Noted on `/pictures/export`: it now exports the union, which matches what the grid shows.

frontend/src/composables/useFilterCounts.js
lindkvis: Mostly fixed in 72ba559c. The cache is now module-level and keyed by the full query, so the strip and the menu header share one "of N" request. At most 6 requests are in flight (a queue behind them). A failure removes its key so the next render asks again. Under `sort=CHARACTER_LIKENESS` only the total is counted. New `useFilterCounts.test.js` covers the stale-response discard, the retry, the cap and the likeness skip, and each goes red when its line is removed.

One deliberate difference: the cache still empties when the menu **opens**. Counts should reflect scoring or tagging done while the menu was closed, and keying on the base query alone would show those stale. With the cap, a reopen costs at most six requests at a time. I didn't add AbortSignal: `getPictureCount` takes none today, and discarding the result plus the cap bound the cost.

frontend/src/components/panels/FilterMenu.vue
lindkvis: Fixed in 72ba559c. `openKind` now searches only the submenu body (`.tbm-section`): the roving radio stop first, then the first enabled input or button, so the header's Clear is never picked. I dropped the dead `activeElement` guard; the checklist and confidence menus still focus their search field from their own `onMounted`. A new test opens Stacks with a filter on and asserts focus is on a radio, and the old selector list turns it red.

lindkvis:
Review answered in 72ba559c. The three blocking findings and the inline notes are replied to on their threads. The summary-only points:

- **`OptionRows` rule.** The §5 "Fields and pick-one" paragraph did change in the first commit: it now reads "`OptionRows` for sort by and the filter menu's pick-one lists". The docstring in `OptionRows.vue` still had the old wording and is now fixed.
- **Panel widths.** Now named tokens, `--filter-menu-w` (320px) and `--filter-submenu-w` (280px), with a comment on why they are narrower than a lone toolbar panel: two sit side by side. You were right that they didn't match an existing panel.
- **Footer names the first option.** Fixed: it names the chosen option, and the first option only as an example when none is chosen. Tested.
- **Checklist highlight off-screen.** `move()` now scrolls the row into view (`block: "nearest"`), and the search field carries `aria-controls` / `aria-activedescendant`.
- **Arrow behaviour differs between submenus.** Every pick-one list now follows the radiogroup contract, confidence thresholds included (arrows select). A chosen option is removed with Clear or the chip's ×, everywhere.
- **`role="dialog"`.** Changed to `role="group"` on all panels. `aria-expanded` / `aria-controls` are kept.
- **`.fm-check--kbd` looked like a focus ring.** It is now the hover wash.
- **"of N" fetched twice.** The strip and the menu read one shared cache entry now.
- **Hidden chip scrollbar.** Now `scrollbar-width: thin`, so chips past the edge show there is more.
- **Smart score / Resolution are strip-only.** Deliberate; commented in `filterChips.js`.
- **Unused `activeCount`.** Removed along with `isActive`, and the test naming the old badge is gone.
- **`FilterStrip` race guard test.** The guard moved into the shared `useFilterCounts`, which now has tests for the stale-response discard, retry and in-flight cap. The strip itself no longer has any request logic.
- **`off` sentinel for one kind only.** Now parameterised over Media, Faces, Stacks and Sharing.
- **Promote `test_predicate_filter.py` to the gate.** Agreed it's a good candidate, but I've left it out of this PR: it means editing `ci.yml`, which goes past security review under our rules, and the deferral predates this PR.
- **Old-to-new feature diff.** Everything `GbFilterPanel` offered has a successor:
  - Shared only → Sharing
  - Unassigned only → Problems → No character
  - Media, Faces, Stacks → their menus
  - Min, max and unscored → Score
  - Impossible tags → Problems
  - Tags with the reject toggle → Tags with Has / Lacks
  - Confidence above/below → Tag confidence
  - Models and LoRAs → their menus
  - The header count and Clear → "N of M" and Clear all

  One deliberate loss, from the design: confidence thresholds are fixed 90/80/70/60% steps instead of a free number.

Still not verified by me: rendering in either theme, the cascade at phone width, and whether focus returns to the funnel when the menu closes.

Issue #1378, opened 2026-09-16
OCR: build the chosen design (text field, word boxes, search scoping, overlay tabs)


lindkvis:
Implements the design chosen for #1197 ("Narrow Results to Text"). Design mockup: https://claude.ai/artifact/6iMXygWz6BmisxAwReqxMp

## Storage and extraction

- New picture field for the text, separate from `description` and from tags, so the caption embedding stays about meaning and the tag board stays clean. Word-level boxes stored with it.
- OCR engine must return a box per word: the overlay's click-a-word interaction depends on it, so a captioner prompted to transcribe is not enough.
- Reading is limited to pictures with a high `text_score`; a `Missing*Finder` picks them up.

## Search

- Every query searches the text as well as descriptions and tags. No option in the search menu.
- Text matches get their own capped ranking weight, so text-heavy pictures do not drift up in unrelated searches.
- A regression set of queries that must **not** return known text-heavy pictures, asserted in tests.
- Search reports per hit whether it matched on text, so the result pill can count and filter without a second query.
- Result pill gains an `All · In text (n)` switch when text matched, beside the face-search controls. Resets when the search is cleared and is never remembered. The pill must not wrap.

## Overlay

- Description section gets two tabs, Description and Text; never both on screen. Tab appears only when text was found; a spinner while it is being read.
- Opening the Text tab draws a faint box around each word; closing it removes them. No toolbar toggle.
- Clicking a word selects its box on the picture (olive selected ring); clicking a box selects the word. Shift-click adds; Copy and Clear underneath.
- Opened from a search that matched text: lands on Text, matched words underlined and dash-outlined, first one selected. Dashes mean matched, olive means selected.
- Text is read-only. Re-run with *Read again* or the model menu.

## Suggested split

1. Backend: field, boxes, migration, finder/task, engine.
2. Backend: search weighting, match source, regression query set.
3. Frontend: overlay tabs plus word boxes.
4. Frontend: result pill switch.

Issue #708, opened 2026-08-04
Three routes disclose project ids that narrow_project_fields hides


lindkvis:
Three routes disclose project ids that `narrow_project_fields` deliberately hides

Found by the independent CSO sign-off on #651's inline-count change (`31fbf954`). **All three are pre-existing and unrelated to that change** — they were confirmed textually outside its diff and reproduce identically with and without the new `include_counts` param. Filing separately so they are not buried in a perf PR.

The common root cause: `visible_project_ids` / `narrow_project_fields` exist so a non-project-scoped token never learns which projects exist or what is in them (#125 / R1b). Three routes skip that narrowing on the way out. The authz coverage matrix cannot catch this class, because it records *which policy a route has*, not *whether the route narrows the project ids inside its payload* — that second axis has no machine check.

## F1 — `POST /projects/membership` returns raw project ids. MEDIUM

`pixlstash/routes/projects.py:296-332`. The handler correctly scope-filters the *picture ids* via `fetch_scope_allowed_picture_ids`, then returns `project_assignments` keyed by the **raw** `project_id`, with no narrowing.

Reproduced with a picture_set-scoped read token whose set holds 3 pictures:

```
GET  /picture_sets        -> project_ids: [[]]        (correctly narrowed)
GET  /projects/1/summary  -> 403
GET  /projects/2/summary  -> 403
POST /projects/membership -> 200 {'project_assignments': {'1': [1, 2], '2': [2]},
                                  'unassigned_picture_ids': [3]}
```

The token learns projects 1 and 2 exist and which of its pictures belong to each, having just been told 403 for both by name.

Fix: intersect the returned keys with `visible_project_ids(server, request)` before serialising — the empty set for character/picture_set/picture tokens. The registry declaration is fine; the missing piece is the narrowing call.

## F2 — `GET /picture_sets?project_id=N` gives a set-scoped token a per-project breakdown. LOW-MEDIUM

`pixlstash/routes/picture_sets.py:466-479`: `project_id` is forced only for *project* tokens. A picture_set token may pass it freely, and it feeds both `set_project_id_filter` (row presence becomes a membership oracle) and `project_filter` (narrows `picture_count`). Reproduced on a 5-picture set:

```
baseline           picture_count = 5, project_ids = []
?project_id=1  ->  [(3, 2)]
?project_id=2  ->  [(3, 1)]
?project_id=3  ->  []            (set not in project 3)
/projects/{1,2,3}/summary -> 403 for all
```

[...]

Issue #651, opened 2026-07-30
API reads stall behind writer-queue commits and WorkPlanner cache eviction


lindkvis:
Split out from #646. That issue is being closed by a frontend stale-while-revalidate cache for the Person/Set context-menu lists, which removes a redundant request storm. It does **not** address why those requests are slow in the first place: API reads stall behind background task work. This issue tracks the root cause so the frontend fix does not bury it.

Reported symptom (#646): on a >500k-picture library, UI actions feel unresponsive while background tasks run.

### 1. Auth on the request path shares the single writer thread with task commits

Every authenticated request's user lookup goes through the serialized DB writer queue:

- `pixlstash/auth.py:485-489` — `get_user()` → `run_task(..., priority=DBPriority.IMMEDIATE)`
- `pixlstash/auth.py:1033-1041` — `get_user_for_request()` → same
- `pixlstash/auth.py:828-830`, `:845-847` — token candidate fetch and `update_last_used` on token auth
- `pixlstash/auth.py:1772-1774` — guest-session cookie lookup
- called from the middleware at `auth.py:1699`

`DBPriority.IMMEDIATE` only wins queue *ordering*. The worker loop (`pixlstash/database.py:842-886`) still runs the currently executing task's session to completion first, so a long batch commit is dead time added to every API request.

Those commits are amplified by the metadata-hash hook attached to every writer-thread session (`database.py:877-878` → `_attach_session_hooks`, `database.py:208-217`). On each flush, `_after_flush_hash_updater` (`database.py:179-205`) calls `_compute_picture_metadata_hash` (`database.py:64-152`) once per dirty picture — 5 queries each plus a Core UPDATE. A tag batch dirties one picture per `Tag` row, so a 64-picture tagger batch adds roughly 320 extra SELECTs + 64 UPDATEs *inside* the write transaction, holding the SQLite write lock and the GIL.

### 2. The WorkPlanner re-runs ~21 finder queries every 50 ms while work exists

`work_planner.py:16` `MIN_INTERVAL_S = 0.05`; `work_planner.py:249-257` resets to `MIN_INTERVAL_S` whenever anything was submitted; `_run_finders_once` (`work_planner.py:264-346`) iterates every finder each pass.

[...]

PR #1388, opened 2026-09-16
Text in pictures: read, search, narrow, find on the picture (#1197)

### LINKED ISSUE #1197: OCR: capture text in images and make it searchable
## Goal

Text inside a picture should be captured and searchable, integrated smoothly
into the search people already use rather than bolted on as a separate mode.
Currently the only AI capability Immich has that we do not.

## What already exists

Pieces that a design should probably use rather than rebuild:

- **Candidate selection.** `Picture.text_score` (indexed, `db_models/picture.py:263`)
  already scores how document-like a picture is, with `TextScoreTask` and
  `MissingTextScoreFinder` behind it. OCR need not run over a whole library.
- **Extraction.** Both JoyCaption (`description_prompt`, an exposed textarea
  parameter) and the published `moondream2_captioner` (`query` mode) can be
  prompted to transcribe today with no code. Neither is necessarily the right
  engine to ship on: a VLM paraphrases and can hallucinate text, and verbatim
  tokens are what an index needs.
- **The plugin interface.** `generate_descriptions` returns `{path: str}`, so a
  dedicated OCR model plugs in the same way whichever engine wins.
- **Lexical search.** `_levenshtein_internal` (`database.py:565`) already does
  word-level normalised edit distance, not whole-string, so it tolerates OCR
  character errors for free. It currently runs over concatenated `Tag` rows
  only, inside `Picture.semantic_search` (`db_models/picture.py:652`).

## Known hazards

- Storing transcripts in `description` puts them into the SBERT vector via
  `to_embedding_dict`, where a wall of text drowns the caption's meaning, and
  means a picture cannot have both a caption and a transcript.
- Feeding long transcripts into the existing lexical branch unweighted inflates
  `coverage` and depresses `mean_query`, so documents drift up in unrelated
  searches. Search quality degrading silently is the main regression risk here,
  and wants a set of queries that must *not* return known documents.
- OCR output is not tag-shaped. Routing it into `Tag` rows would flood the tag
  health board and the review queue.

## Out of scope for this issue

How to assemble it: storage, weighting, engine choice and UI are open. Not on
the website comparison table until it ships.

lindkvis:
Closes #1197. Builds the chosen design, "Narrow Results to Text" (#1378).

## What changed
- **Reading:** Florence-2, which the app already ships, reads pictures with `text_score >= 0.25` using `<OCR_WITH_REGION>`. The result goes into new `Picture.ocr_text` / `ocr_words` columns (migration 0120), never into `description` or tags.
  - Florence boxes whole lines, so each word gets the share of its line's box its characters take up.
  - Boxes are fractions of the picture as displayed, with EXIF orientation applied.
  - The work runs through a new `MissingOcrFinder` / `OcrTask`. A rotate clears the text so the picture is read again, and the open overlay re-reads it at once so the old boxes never linger.
- **Search:** a text match adds a capped 0.35 to the score, and only when every query word appears in the text. Short words must match exactly; longer ones tolerate one or two misread letters ("C0FFEE" matches "coffee").
  - Each search row carries `text_match`.
  - `tests/test_ocr_text.py` holds queries that must not match known receipt, menu and screenshot text.
- **Routes:** `GET /pictures/{id}/text` returns the words, their boxes and which ones matched. `POST /pictures/{id}/text/read` is Read again. Both are `PICTURE_SCOPED`, and the coverage matrix and route table are updated.
- **Overlay:** Description and Text are now tabs. Word boxes show while Text is open. Clicking a word selects it on the picture and clicking a box selects the word; shift-click adds words, with Copy and Clear underneath. Opening a picture from a search that matched its text lands on Text, with the first match selected.
- **Result pill:** an *All · In text N* switch narrows the results client-side. It resets when the query changes and gives way to the face-search controls.
- The OCR columns are left out of `metadata_fields()`, the metadata hash and the restore diff, so metadata and search payloads don't carry kilobytes of words.

## Where this departs from the design or the issue
- **No model menu on the Text tab.** There is one reading engine, so a one-item menu wasn't built. Read again is there.
- **Florence-2 instead of a word-level detector.** The design asks for an engine that returns word boxes. Florence returns line boxes, and splitting them by character count gives word boxes that are close but not exact on proportional fonts (marked with a `ponytail:` comment). This avoids a new dependency and a new model download.
- **Reading follows captioning.** It is idle while captioning is switched off, because it loads the same model.

## Known limits
- **Panning over words:** word boxes stop pointer events, so while Text is open a drag that starts on a word does not pan. This keeps a click on a word from turning into a pan.
- **Box accuracy:** boxes come from splitting each line's box, so they drift on proportional fonts and are wrong for text set at an angle.
- **Small text:** pictures are read at 1024 px, so small body text on a large screenshot may come back empty.


pixlstash/tasks/ocr_task.py
```diff
@@ -0,0 +1,196 @@
...
+                path_to_id[path] = pic.id
+
+        lines_by_path: dict = {}
+        if path_to_id and not self._stop_event.is_set():
+            lines_by_path = self._engine.read_text(list(path_to_id)) or {}
+            if not lines_by_path:
+                # Nothing read at all is the model failing (not loaded, out of
+                # memory), not every picture being unreadable. Store nothing, so
+                # the pictures stay unread and a later sweep tries again.
+                raise RuntimeError(
```
lindkvis, line 118: **Blocking — a permanently broken reader makes the planner submit a doomed task roughly every 0.05 s, forever.**

Storing nothing on a whole-batch failure is the right call for a *transient* one. The problem is that nothing distinguishes transient from permanent, and the retry path has no damping:

- `Florence2Service.read_text` returns `{}` on **any** exception — a missing model, a processor/transformers mismatch, a CUDA error whose CPU reload also fails. Those don't heal.
- This raise means `_persist_text` never runs, so `ocr_text` stays `NULL`.
- `BaseTaskFinder.on_task_complete` releases the claimed picture ids unconditionally, so the same pictures are immediately selectable again.
- `MissingOcrFinder._guard` only checks that an `active_description_plugin` is configured. It doesn't know the reader just failed.
- `WorkPlanner`'s adaptive backoff doesn't apply: at `work_planner.py:443–449` the interval resets to `MIN_INTERVAL_S` (0.05 s) whenever anything **was submitted**, and this task is submitted every cycle. The backoff only damps cycles that find *no* work.

Steady state: load the model, fail, raise, release, re-select, repeat at 20 Hz on the GPU queue, with an `ERROR` log line per attempt, for the life of the process. `ensure_captioning_ready()` raising instead of returning gives the same loop.

This repo already has the pattern. `MissingCheckpointHashFinder` keeps a `self._deferred: set[int]` for the life of the process (`missing_checkpoint_hash_finder.py:67`, `:163`, `:166`), fed both from the task's result and from `on_task_complete`'s error — and its docstring gives the reason in almost these words: *"or one broken path would make the planner submit a doomed task every cycle"*.

`tests/test_ocr_text.py:399` asserts the picture stays `NULL` after the reader returns nothing — i.e. it pins the half of the behaviour that creates the loop, and nothing covers the other half.

**Suggested fix:** give `MissingOcrFinder` an `on_task_complete` that counts consecutive whole-batch failures and stops handing out work after a small number, or defers the failing ids for the process lifetime, mirroring `MissingCheckpointHashFinder`. `POST /pictures/{id}/text/read` bypasses the finder, so an operator still has a way back in without a restart. A test driving two consecutive failures and asserting the third planning cycle submits nothing would guard it.

---

pixlstash/routes/pictures/_search.py
```diff
@@ -460,6 +463,15 @@ def log_semantic_results(label: str, rows):
                     and getattr(result[0], "id", None) not in hidden_ids
                 ]
         rows = [Picture.serialize_with_likeness(r) for r in results]
+        # Per row, whether the picture's own text matched: the result pill counts
+        # and narrows on this without a second search (#1197).
+        text_matched_ids = (
```
lindkvis, line 468: **Blocking — every search scans the whole read library through a Python SQLite UDF, to set a flag on at most one page of rows.**

`Picture.ids_matching_text` takes no ids and has no `LIMIT`:

```python
select(cls.id).where(
    cls.ocr_text.is_not(None),
    cls.ocr_text != "",
    func.ocr_text_match(cls.ocr_text, query) > 0,
)
```

`ocr_text_match` is a Python function registered with `create_function` (`database.py:846`), so SQLite calls back into Python once per surviving row. Per row it does `ocr_text.split()` and then, for a row that does **not** match, walks every word in the picture for the first query term before `all()` can short-circuit — and `ocr_word_matches` re-runs `re.findall(r"\w+", word.lower())` on each word for each term (`database.py:726–729`).

So the cost is `O(pictures_read × words_per_picture × query_terms)` Python-level regex plus Levenshtein calls **per search request**, and the non-matching rows — the majority — are the expensive ones. Dense pages are hundreds of words each. This runs on every keystroke-debounced search, and `query` is mandatory (`:114`), so no path skips it.

All of that work is discarded except for the ids that happen to be in `rows`, which is bounded by the page limit. `docs/integration_architecture.md` §25 says the *All · In text* switch "filters on `text_match` client-side; nothing is refetched" — so the per-row flag is only ever needed for rows already in hand.

It's also unscoped: it returns ids of soft-deleted pictures and of pictures outside the caller's token scope. That isn't a leak — the ids are only used as a set-membership test against rows that already passed scope and hidden-id filtering, and the security lane confirmed that separately — but it means a scoped token pays for a full-library scan.

**Suggested fix:** pass the ids in and filter on them.

```python
@classmethod
def ids_matching_text(cls, session, query, picture_ids):
    if not picture_ids:
        return set()
    return set(session.exec(
        select(cls.id).where(
            cls.id.in_(picture_ids),
            cls.ocr_text.is_not(None),
            cls.ocr_text != "",
            func.ocr_text_match(cls.ocr_text, query) > 0,
        )
    ).all())
```

That turns it into one indexed lookup over at most a page of rows. While you're there, `ocr_text_match` can hoist the per-word tokenisation out of the term loop — a picture with N words and a T-term query currently does N×T `re.findall` calls over the same N strings. (The `score_cutoff=budget` / `<= budget` pairing is correct for rapidfuzz, for the record: `distance` returns `score_cutoff + 1` when the budget is exceeded.)

Alternatively, `semantic_search`'s `text_match_weight` term has already computed this — carrying the value out of there would retire the second pass entirely.

---

pixlstash/tasks/ocr_task.py
```diff
@@ -0,0 +1,196 @@
...
+            pic.ocr_words = json.dumps(lines) if lines else None
+            session.add(pic)
+            changed.append((Picture, picture_id, "ocr_text", bool(lines)))
+        session.commit()
+        return changed
+
+    @staticmethod
+    def _unread_query(query):
+        return (
+            query.where(Picture.ocr_text.is_(None))
```
lindkvis, line 166: **Blocking on the letter of the rule, though I'd accept a reasoned "no" — the finder's probe predicate has no index leading with the nullable work column.**

`docs/backend_architecture.md` §7, *Probe cost (#651)*, states two rules a new finder must obey. Rule 1 (never select the whole ORM `Picture`) is obeyed — the `load_only(Picture.id, Picture.file_path)` is right and the docstring explaining why is good. Rule 2 is not:

> Give the predicate an index, and lead it with the nullable column. … a partial index only wins when it can claim *more* equality terms than the index it competes with, and `<col> IS NULL` counts as one.

The predicate is `ocr_text IS NULL AND text_score >= 0.25 AND deleted = 0 AND file_path IS NOT NULL`. The only index available is `ix_picture_text_score`, and `text_score >= 0.25` is a **range**, not an equality term.

The docstring's claim that "the `text_score` index narrows it to the few pictures that qualify" holds while few pictures qualify — but that set only grows, and once those pictures **have** been read, the probe still range-scans every one of them every cycle and rejects them on `ocr_text IS NULL` afterwards. On a screenshot- or document-heavy library that's the steady state, at the planner's 0.05 s cadence. §7 records the measurement the rule came from: 80 ms vs 0.003 ms per probe on a 200k-row table, for exactly this shape.

**Suggested fix:** add `(ocr_text, text_score, deleted, id) WHERE ocr_text IS NULL` in migration 0120. Leading with the nullable column is what makes the plan independent of whether `ANALYZE` has ever been run, per §7.

Marked blocking because §7 words it as mandatory for a new finder, not because a regression was measured here — the `text_score` index does mitigate it on a library that's mostly photographs. If you think the mitigation is enough, say so on this thread and leave it; that's a reasonable position, it just needs to be an argued one rather than inherited.

---

frontend/src/components/views/OverlayDescriptionPanel.vue
```diff
@@ -116,15 +196,74 @@
...
+        class="picture-text"
+        role="tabpanel"
+        :aria-labelledby="textTabId"
+      >
+        <div
+          v-for="(line, lineIdx) in wordLines"
+          :key="lineIdx"
+          class="picture-text-line"
+        >
+          <button
```
lindkvis, line 225: **Blocking — arrow keys inside the Text panel navigate to another picture instead of moving through the words.**

The overlay's global `handleKeydown` maps `ArrowLeft/Right/Up/Down` to `showPrevImage()` / `showNextImage()`, and only stands down for `isEditableElement(target)` — which is `INPUT` / `TEXTAREA` / `SELECT` / `contenteditable` / `role="textbox"`. A `<button>` is none of those, so with focus on a word:

- `ArrowUp` / `ArrowDown` — the natural way to scroll a panel that is `overflow: auto` with a 56px min-height — swap the picture instead, and the user loses both their place and their selection.
- `ArrowLeft` / `ArrowRight` do the same.

You already solved this exact collision one element away: `onTabKeydown` (`:421–432`) calls `preventDefault()` and `stopPropagation()` and moves between the two tabs itself. The word list needs the same treatment, which is why this reads as an oversight rather than a decision.

**Fix:** give `.picture-text-word` a `@keydown` that at minimum stops propagation for the four arrow keys. Roving focus — arrows move between words, only the active word carries `tabindex="0"` — would also fix the second half of the problem: a photo of a page of text currently produces one tab stop per word, with no way to skip past them.

---

frontend/src/components/views/ImageOverlay.css
```diff
@@ -804,6 +804,69 @@
...
+
+.picture-text-box-halo {
+  stroke-width: 3px;
+}
+
+.picture-text-box-line {
+  stroke-width: 1px;
+}
+
+.picture-text-box .picture-text-box-line {
```
lindkvis, line 841: **Blocking on the hover state — at rest and on hover a word box is a dark line with no halo, so it disappears over dark photographs.**

The block comment two lines above says "every mark pairs a line with a contrasting halo underneath". That holds for `--match` (light halo under a dark dashed line, 848–857) and for `--selected` (dark halo under the olive ring, 858–869). Both read correctly over any photograph, and both are good.

It doesn't hold for the two states that are neither. `.picture-text-box rect` sets `stroke: transparent`, and `--halo` only ever receives a stroke inside the `--match` and `--selected` rules. So:

- **rest:** `rgba(var(--v-theme-dark-surface), 0.13)` — a 1px stroke of `#242628` at 13%, no halo.
- **hover:** `rgba(…, 0.6)` stroke plus a 12% fill, still no halo.

Over a light photograph both are faint but present. Over a dark one — a night scene, a dark product shot, a black-background poster — both are effectively invisible. The repo's own rule is that a mark depending on what the photograph happens to look like is a defect; this is that rule with the polarity reversed.

Rest being quiet is arguably deliberate and I won't push on it. **Hover is the part I'd hold for:** it's the only confirmation that a 1px outline over arbitrary imagery is a hit target at all, and on a dark photo the user gets no feedback that the thing under the cursor is clickable. Clicking a box is one of the two headline interactions, so that matters.

**Fix:** give the halo a stroke in the two remaining states, mirroring `--match`:

```css
.picture-text-box:hover .picture-text-box-halo {
  stroke: rgba(var(--v-theme-on-dark-surface), 0.9);
}
```

and, if rest should be visible at all, a lower-alpha version of the same on the base rule.

This is a structural argument — no halo, dark-only ink over arbitrary imagery — rather than a measured contrast figure, since nothing was rendered.

---

pixlstash/db_models/picture.py
```diff
@@ -1195,12 +1240,16 @@ def find(
 
         return session.exec(query).all()
 
+    # Served by ``GET /pictures/{id}/text`` alone: a screenshot's words and
```
lindkvis: **Non-blocking — `OCR_FIELDS` doesn't do what its comment claims.**

From the security pass. `metadata_fields()` subtracts the OCR columns, but `GET /pictures/{id}/{field}` builds its allowlist from `Picture.scalar_fields()` — the raw column namespace — so `GET /pictures/{id}/ocr_text` and `.../ocr_words` return the raw text and the word-box JSON directly.

**This is not an authz hole.** That route carries the same `picture_scoped` `id_param="id"` policy as `/text`, so the principals are identical and nothing reaches a caller that `/text` wouldn't already serve. But both of the stated rationales are defeated: the text is not "served by `/text` alone", and the kilobyte argument for keeping it out of metadata payloads doesn't hold when the same bytes are one field fetch away.

Worth knowing that CI won't catch it either way: `test_no_new_relationship_becomes_servable` asserts `len(columns) >= 5`, a floor rather than a pin.

**Fix:** either exclude the two columns in `field_allowlist.py`, or reword the comment to say what's actually true. Both are defensible — the point is that the comment and the behaviour should agree.

Minor, same lines: `OCR_FIELDS` is a `ClassVar` declared between two `@classmethod`s. CLAUDE.md's class-member order puts class-level variables before `__init__`, and the file's only other `ClassVar` (`ExportType`, line 202) follows that. I can see why it's here — it reads well next to its only consumer — but it's now 2 of 2 on the convention going one way and one the other. Either move it up beside `ExportType` with a pointer in `metadata_fields`, or say on this thread that proximity wins, so the next person doesn't "fix" it.

---

lindkvis (commented):
## Review — text in pictures

Reviewed in three lanes — backend/ML, security/authz, frontend — merged here. **Six blocking findings: three backend, three frontend. The authorization work signs off clean.** Six inline notes plus the items below.

### Security sign-off

Taken separately and adversarially, because CLAUDE.md requires an independent pass on any authz change and forbids the author certifying their own.

**No blocking authorization defect.** The registry arithmetic was re-derived by AST-parsing `ROUTE_POLICIES` at both refs rather than read from the PR body: `develop` 328 entries / 41 `PICTURE_SCOPED` → branch 330 / 43. Exactly +2 picture-scoped, all nine other policy classes byte-identical, no duplicate keys, nothing retargeted. `_CURRENT_ROUTE_ALLOWLIST` is still `frozenset()`. The coverage matrix gains exactly the two matching rows. Five endpoints touched, no empty cells.

The adversarial negatives that did **not** land, recorded so the absence is auditable:

- **No existence oracle.** Scoped tokens get an identical 403 for out-of-scope and nonexistent; the handler's 404 is only reachable by principals the gate didn't narrow. That was #708's bug and it isn't repeated here.
- **No gate/handler parser differential.** This was the main bypass hunt, since the route annotates `id: str` and converts in-handler. Both the gate (`_check_one_id`) and the handler (`_picture_id`) call `int()` on the same raw string, so no value can authorize as one picture and serve another.
- **No inline scope ladders** in `_text.py` (grep count 0). The gate owns it, correctly.
- `text_match` is a bool, never content. The WS event carries the field *name* and `bool(lines)`, never the text. No OCR content in any log. No SQL injection — bound scalar function.

**And the negative tests would not pass for the wrong reason.** I asked specifically about this, because the READ-token middleware runs ahead of routing and a 403 from a dead path is indistinguishable from a real refusal. `test_text_routes_follow_the_token_scope` puts the out-of-scope 403 beside an in-scope **200 positive control on the same path template**, so a renamed path would fail the 200 first. The POST's route-resolution proof is a separate test POSTing the same template as owner and asserting 200 plus the queued task. It uses `TestClient(server.api)` with an explicit bearer header rather than the authenticated fixture, so the negative isn't passing on a missing credential either. The one residual: no positive control for a *scoped* POST — but no such token can exist, since resource-scoped implies READ, so 403 is the only correct outcome.

One forward-looking note: the POST's safety rests on a chain it isn't named in — resource-scoped ⇒ mintable `READ` only ⇒ middleware refuses non-GET unless the path is in `READ_SAFE_POST_PATHS`. Correct today, which is what makes the deliberate no-dedupe owner-only self-inflicted load. An explicit "do not add `/text/read` here without a claim" note at that frozenset would keep it that way.

### What checks out

Stated up front so the blocking list reads as a short list rather than a verdict.

- **Migration 0120 is correct.** Both `op.add_column` calls go through the `inspector` + `existing_cols` guard, the revision chain is strictly last (`0119_rescan_pictures_for_a1111_recipes` is the current head and nothing else claims it as `down_revision`), `__all__` is present, and the file is schema-only. No NULL-reset needed since the columns are born NULL. The extra table-existence guard and the mirrored `downgrade()` are both sound.
- **The word-box degenerate cases are handled and tested.** Division by a zero character count is unreachable — `not text.strip()` rejects empty and whitespace-only labels before `char_w` is computed. Zero-area boxes are rejected by `x2 <= x1 or y2 <= y1`. A one-word line works. Coordinates are clamped to the image edges. `tests/test_ocr_text.py:98` feeds a zero-area quad and a whitespace-only label in the same call and pins the character-share arithmetic to exact pixel values rather than to a shape.
- **EXIF orientation is consistent with what the browser gets.** For JPEG the browser applies the tag itself; for every other format the backend `exif_transpose`es before responding (`_serving.py:135–193`). So "the picture as displayed" is the EXIF-applied bitmap either way, and `ImageOps.exif_transpose` in `read_text` puts the boxes in the right frame.
- **Zoom, pan and window resize need no new handling on the frontend.** The `<svg>` mounts inside `mediaInnerRef`, the same layout space as the face and detection boxes, inside the CSS transform. `overlayDims` is driven by the existing `ResizeObserver` pair plus the window listener, and both `textBoxesStyle` and `textBoxRect` read it reactively. Correct by reuse.
- **Stale boxes on a picture swap can't happen** — two independent guards, either sufficient: the `watch(() => image.value?.id)` resets `overlayDims`, making `showTextBoxes` false until the new picture is measured; and `load(id, { landing: true })` sets `text.value = NO_TEXT` synchronously.
- **The "lands on Text with the first match" race doesn't exist.** `usePictureText` keys only off `image.value?.id`, which is set *before* `fetchOverlayMetadata`, so the text read neither waits on nor can be beaten by the metadata round trip. `load` is serialised by a monotonic `requestId` that `readAgain` also bumps.
- **No listener or observer leaks.** No new window listeners, no new observers, `textCopyTimer` cleared in `onUnmounted`.
- **Test gating and changelog are correct.** `tests/test_ocr_text.py` is in the `backend` job in alphabetical position between `test_model_unload_race.py` and `test_openapi_response_schemas.py`. Frontend tests run over a glob, so the six new/extended files are gated automatically. `changelog.d/1197.md` is user-visible and `CHANGELOG.md` is untouched. Docs updated in the same change.
- **The test corpus is invented, not real.** `RECEIPT`/`MENU`/`SCREENSHOT` read as synthetic, every path is the sanctioned `/home/me/` placeholder, and the sample words in the frontend tests are invented too. Nothing private committed — I asked both lanes to check this specifically, since OCR fixtures are exactly where real receipts and screenshots end up.

### One blocking finding that can't anchor inline

**A rotate leaves wrong — not merely stale — word boxes on the picture, and they stay clickable.** `ImageOverlay.vue:1549`, untouched context, inside `runRotate`.

`runRotate` already re-reads the two other box families after a turn, under its own comment:

```js
// The boxes are drawn in the file's own coordinate space, which the turn
// just redefined. Re-read rather than transform them here: ...
fetchFaceBboxes(imageId);
fetchDetections(imageId);
```

The word boxes are a third family in that same space and weren't added. The backend confirms the invariant applies to them: `_apply_orientation` NULLs both columns "so `MissingOcrFinder` reads the text again in the new orientation", and `GET /pictures/{id}/text` then returns `{state: "pending", lines: []}`.

So after a turn the overlay keeps rendering pre-turn boxes until the background OCR task reruns — GPU work behind a queue, so seconds to minutes. In that window the boxes aren't stale-but-harmless: after 90° they sit over unrelated parts of the picture, they still take pointer events, and clicking one selects an unrelated word in the sidebar. The rotate emits `fields: { pixels: true }`, which `pictureChangeAffectsView` deliberately ignores, so nothing else clears them.

**Fix is one line beside the two existing siblings:** `pictureText.refresh();` — it re-reads, gets `pending`, empties `lines`, and the Text tab drops to its spinner, which is the honest state.

**This contradicts declined item (a), so here's the argument rather than an assertion.** The declined framing was "the old boxes stay until you move to another picture", which reads as cosmetic lag. Three things make it more: the boxes are positively *wrong* rather than absent; they remain *interactive*, so the surface lies about which word is where; and the fix is one line in a function that already performs exactly this operation twice, under a comment stating the rule being broken.

### On the other two declined behaviours

**(b) Word boxes swallow drag-to-pan** — agreed, fair follow-up. Pan is available everywhere the words aren't, Text is opt-in, and the alternatives (a movement threshold, or forwarding the pointerdown to `onPanStart`) are real work with their own failure modes.

**(c) "Read again has no dedupe"** — this looks **already handled**, and may need no follow-up at all. Three guards stack: `usePictureText.js:159` checks `readAgainBusy` synchronously before the `await`, so a second call in flight is a no-op regardless of the DOM; `:disabled="textReadBusy"` covers the window before the POST resolves; and once it resolves `isTextTab` goes false and the button unmounts. The only way to queue a second read is to wait for the first to finish and press again — a deliberate repeat. Worth confirming on the real thing, but no accidental double-queue path turned up.

### Non-blocking, backend and ML

- **The 5-letter one-edit band admits ordinary English minimal pairs.** `budget = 0 if len(term) < 5 else 1 if len(term) < 9 else 2`. The 4/5 boundary is pinned deliberately and the four-letter cases in the corpus are exactly the dangerous ones, so the low end is well chosen. The question is the other side: at five letters one edit covers a lot of vocabulary — a query for `loose` matching a read `goose`, `moose` or `lose`. For multi-word queries the conjunction makes this harmless; for a **single** five-letter term it's the whole test, and 0.35 clears the app's 0.1 floor on its own. Consider starting the one-edit band at six, or requiring the first character to match (OCR confusions are overwhelmingly interior — `0`/`O`, `1`/`l`, `rn`/`m` — so that costs little recall). If the current band is the considered choice, a `MUST_NOT_MATCH` entry that's a five-letter minimal pair would document it.
- **A capped 0.35 is below the API's own default threshold.** The composition is sane and matches the comment. But `threshold` defaults to `Query(0.5)`, and `semantic_search` keeps a row only when `combined_score >= threshold` — so a picture whose *only* signal is a text match scores exactly 0.35 and is filtered out at the route's default. The app is fine (`pictures.js:249` sends 0.1), but any API consumer using the documented default gets a feature that silently does nothing, and §25 says "every search also matches the text" without the caveat. `tests/test_ocr_text.py:189` pins `threshold=0.01`, so no test covers the route default.
- **`_words_from_ocr_regions` silently drops words when `quad_boxes` and `labels` disagree in length** (`florence2.py:106`). A bare `zip` truncates to the shorter sequence; a truncated generation at the 1024-token cap is the realistic way there. Per the no-silent-failures rule, a `logger.warning` naming both counts and the path would make it visible. (`strict=True` would be worse — losing the whole picture over a partial parse.)
- **Leading/trailing whitespace in a label shifts and shrinks every box on that line** (`florence2.py:117–119`). `len(text)` includes the padding but Florence boxes the ink, so for a label `"  HELLO"` `char_w` is a seventh of the line width instead of a fifth, and the word is drawn at two-sevenths in at five-sevenths of the true width. Work from `text.strip()` and offset `match.start()` by the characters stripped from the left. The whitespace-only case is covered; a padded-real-words case isn't.
- **The `ponytail:` caveat is honest but under-scoped.** It covers proportional-font drift. Two other mechanisms produce the same symptom: the whitespace inflation above, and — more importantly — **rotated and skewed quads**. `<OCR_WITH_REGION>` returns a quad, and lines 110–116 reduce it to its axis-aligned bounding box. For text at an angle the box is much larger than the text, and splitting it *horizontally* by character index puts every word in the wrong place. That's the case where boxes aren't merely drifted but wrong, and a reader of the current caveat wouldn't expect it. One extra sentence is enough; not asking for the fix.
- **The batch isn't grouped by expected output length** (`ocr_task.py:182`, `.order_by(Picture.id)`). The vision side is moot — Florence's processor resizes to a fixed square, so `pixel_values` are uniform. But batched generation runs every sequence until the longest finishes, and with `num_beams=3` and `max_new_tokens=1024` one dense page makes its three neighbours pay its token count, three beams deep. Ordering by `Picture.id` mixes a receipt with three sparse screenshots by construction. `text_score` is already the signal for how much text a picture carries, already indexed, and already in the predicate — ordering the probe by it makes each batch roughly uniform at no cost to correctness. Worth measuring first, but it's a one-line change to a query you're already writing.
- **The VRAM estimate is knowingly wrong for this task.** The comments are candid: `BATCH_SIZE = 4` is kept small because reading "decodes with three beams and up to 1024 tokens, more than the captioning estimate the VRAM gate charges for", and `estimated_vram_mb` returns the captioning estimate anyway. Three beams triples the KV cache, so the gate is systematically under-charged — and the consequence isn't just this task's OOM but that the gate admits *other* GPU work alongside an under-priced read. Worse, `read_text` swallows the CUDA error and returns `{}`, so the OOM never surfaces to `is_vram_oom()`, the VRAM retry never engages, and the failure lands in the spin loop above instead. Cheapest honest fix: multiply the captioning estimate by the beam count so the gate sees the real number and `BATCH_SIZE` stops carrying two jobs. Also minor: the bare `except Exception` there logs at `debug`; a failed VRAM estimate is only ever read after something went wrong, so `warning` serves better.
- **Reading happens at a fixed downscale with no tiling** (`max_dim=1024`, then the processor's own resize). For a 4K screenshot or a photographed page, body text is below the effective resolution by the time it reaches the model, so the realistic output is headings and large text. Inherent to Florence-2 and consistent with the existing `detect_objects` path, so the right call — but it interacts with the `text_score >= 0.25` gate in a way worth a doc sentence: a picture can score well above 0.25 on MSER and still come back `ocr_text = ""` because none of its text survived the downscale. That's stored as "read, nothing found" and never retried.
- **`POST /pictures/{id}/text/read` doesn't check `deleted`**, so a scrapheap picture can be queued for a read — while `GET .../text` is careful about exactly this (`_text.py:64–70`). The two disagree. Also, the task is built from a `SimpleNamespace` and submitted directly, so `_claimed_picture_ids` never sees the id; if `MissingOcrFinder` has the same picture in flight, both write and last one wins. Harmless for derived data, but the existing `ponytail:` note only flags the no-dedupe half.
- **`PictureTextWord.box` accepts a list of any length** (`_text.py:26`). The `try/except` catches malformed JSON and missing keys and `tests/test_ocr_text.py:346` covers three shapes — but a valid `list[float]` of the wrong length passes both and reaches the client, where it becomes a rendering bug at a distance from its cause. `conlist(float, min_length=4, max_length=4)` closes it and adds a fourth parametrised case.

### Non-blocking, frontend

- **Pressing "Read the text again" drops keyboard focus to `<body>`** (`:60–66`). `:disabled="textReadBusy"` unfocuses the active element per spec; then once the POST resolves the whole `<span v-if="isTextTab">` unmounts with the button inside it. Either way a keyboard user who activates the primary new action restarts Tab from the top of the document. `focusOverlayCanvas()` is the existing convention.
- **Read again bounces the user off the Text pane and back, twice per read** (`usePictureText.js:166`). Setting `{state: "pending", lines: []}` makes `hasText` false, so `activeTab` falls back to Description while `chosenTab` stays on Text — landing the user on a live textarea where a stray keystroke can start an edit. Keeping the Text panel mounted while pending would read better.
- **Copy and Clear disappear if every selected word reads as empty** (`:231–232`). The strip is gated on `selectedText`, the joined string; `flattenWords` coerces a missing `text` to `""`, so a selection of only empty-text words yields `""` and the user has a visibly selected word and no Clear. Gate on `props.selectedWords.length`.
- **The `All | In text` segment omits the pressed-state hover from the recipe it cites** (`SearchResultBar.vue:662–666`). Reproduces `.seg__opt--on` faithfully but `Segmented.vue` also has `.seg__opt--on:not(:disabled):hover { background-image: var(--hover-shade); }`. Without it the pressed segment is the one control in the pill with no hover response. One rule; everything else in the block is clean.
- **`textSearchResults` is never released** (`ImageGrid.vue:1564`). `searchPictures` is called with `topN: 10000` and the full row array is held for the grid's lifetime, replaced only by the next text search. `textMatchCount` guards on the query, so nothing is wrong — just resident. Clearing it in the same `watch(searchQuery, …)` that resets `textMatchesOnly` would be tidy.
- **`readAgainBusy` is shared across pictures** (`usePictureText.js:85`) — press Read again, navigate before the POST resolves, and the button is disabled on the next picture until the first request finishes. Transient and self-correcting.
- **Question on a token.** The selected-word wash is `0.28` where `--active-wash` on dark chrome is `0.2` (`:849–852`). The `0.28` in `ImageOverlay.css:867` carries a comment saying the alphas are the approved design's and live only there — fair for a mark over a photograph. This one is in the sidebar, the surface `--active-wash` was measured for, so it reads as the SVG value copied inward. The ring beside it does match `--active-bar`, and keeping the label in ink rather than olive is correct.
- **Off-token values**, ranked: `rx="1.5"` (`ImageOverlay.vue:762`) is an ad-hoc radius where `--radius-sm` exists; `style="opacity: 0.6"` (`:88`) adds a 38th instance of the literal `visual-language.md` explicitly names as the one to stop copying, where `--opacity-text-secondary` is the value; `column-gap: 0.6ch` is off-grid but defensible in a monospace block. The `min-height: 56px / 28px` and `height: 24px` literals are house style (51 exist), though `--control-h-sm` would cover the 24px button.
- **No cap on word count anywhere in the chain** — a scanned page could produce thousands of words, each rendering two `<rect>`s plus a `<button>`, no virtualization, and `textBoxRect()` re-runs across every box on resize. Unmeasured and possibly a non-issue; if it matters, moving the fractional coordinates into the viewBox (`viewBox="0 0 1 1"` + `preserveAspectRatio="none"`, strokes already carrying `vector-effect: non-scaling-stroke`) makes the rects invariant under resize.

**One correction worth recording so a later reviewer doesn't "fix" it:** the raw `--v-theme-dark-surface-primary` at `OverlayDescriptionPanel.vue:790` should **not** become `--active-bar` / `--selected-ink`. Those aliases resolve to the deep olive `#567309` under the light theme, which would be wrong on this always-dark surface. The raw lifted olive is the established pattern here, 30 occurrences. The code is correct as written.

**Spec note:** the description says shift-click "adds words"; `nextWordSelection` implements **toggle**, not range — shift adds an unselected word and removes a selected one, with no start-to-end fill. That matches the description and the in-app hint, so it reads as correct; flagging only in case range selection was intended upstream.

### Test coverage

~490 new frontend test lines and a substantial backend file, with negative controls of the right shape. The state machine is well covered: tab resolution across `none`/`pending`/`read`-with-no-words, landing on Text with the first match, selection cleared per picture, `ocr_text` never counting as a view change (with a `tags` control), the re-cut without a second `searchPictures`.

Gaps, in risk order:

1. **The coordinate transform is untested.** `textBoxRect` / `textBoxesStyle` (`ImageOverlay.vue:4427–4447`) are the highest-risk code in the change and carry no assertions. Fractions in, pixels out against a known `overlayDims`, plus the `box.length !== 4` and null-box branches, would be cheap.
2. No test of `onTabKeydown`, **including the `stopPropagation()`** that keeps the overlay from swapping pictures. That guard is exactly what gets refactored away silently — and the blocking arrow-key finding is the same bug one element over.
3. No test that the boxes clear on a picture swap (they do, two ways — neither pinned), and none of `onTextBoxClick` → `revealWord`.
4. **`frontend/e2e/` is unchanged** — no end-to-end pass over search → open → land on Text.
5. No test drives a **partial** `read_text` result (some paths returned, others omitted), which is the "image could not be opened" branch at `ocr_task.py:127–134`.
6. Minor: `OverlayDescriptionPanel.test.js:141` writes the expected count as `String("BAKERY No.4\nC0FFEE".length)` rather than `"18"`, restating the implementation's join rule in the expectation.

On the OCR corpus specifically: it's honest but thinner than its length suggests. Several `MUST_NOT_MATCH` entries genuinely bite the 4/5-letter boundary ("cart" against a read "CARD", "cafe" against "Cake") and several exercise the conjunction over terms. But four entries — "dog", "sunset beach", "mountain lake", "white dress" — share no vocabulary with any corpus text and would pass against almost any implementation.

### Not verified

No tests were run, no model was loaded, nothing was rendered — the checkout is shared with other sessions, so all three lanes worked read-only. Consequences worth stating plainly:

- **Per the repo's mutation rule, nobody has shown a new assertion failing when the thing it guards is broken.** That should happen before merge. The cheapest one that proves the whole gate story: delete the GET `/text` registry line and expect three independent reds — `test_all_routes_declare_access_policy`, `test_coverage_matrix_document_matches_the_registry`, and the in-scope 200 flipping to a gate 403.
- Every performance claim (the search scan, the probe predicate, the batch grouping) is reasoned from the code and from the measurements already recorded in `backend_architecture.md` §7, not measured here.
- The `text_score >= 0.25` threshold and the `max_dim=1024` accuracy ceiling are judged from the code and the model's known properties, not validated empirically.
- `ruff` isn't installed in the review environment, so neither check was run.
- Contrast is a structural argument, not a computed figure. Nothing was rendered, so no judgement on whether the boxes actually land on the words for a real rotated picture, or whether read quality at `max_dim=1024` is good enough on a dense page.
- The rotate **undo** path wasn't traced — `restore/preview.py:989–993` puts both columns in the `_SKIP` set, which is right for a diff, but skipping the diff and restoring the value are different questions and only the first was confirmed.
- Registration order in `routes/pictures/__init__.py:46` puts `_text` before `_crud`, which is what keeps `GET /pictures/{id}/text` from being swallowed by the `/{id}/{field}` catch-all. That's the documented correct pattern and tests exercise it, but it's a registration-order invariant a future reorder would break silently, and no test names it as such.

---

pixlstash/tasks/ocr_task.py
lindkvis: Fixed in b56ea88b. `MissingOcrFinder` now keeps a session-lifetime `_deferred` set, the `MissingCheckpointHashFinder` pattern: `on_task_complete` releases the claims and, on any error except `TaskCancelledError`, defers that batch's picture ids with a warning; `_fetch_candidates` skips them. A permanently broken reader therefore burns each qualifying picture once and then the finder runs dry. Read again still reaches a deferred picture. Separately, `read_text` now re-raises an out-of-memory error instead of returning `{}`, so the runner's VRAM retry engages. Test: `test_a_failed_batch_is_deferred_but_a_cancelled_one_is_not` (red with the defer removed, and red with cancellation deferred too).

pixlstash/routes/pictures/_search.py
lindkvis: Fixed in b56ea88b. `Picture.ids_matching_text(session, query, picture_ids)` takes the returned rows' ids and filters `id IN (...)` in `chunked` slices, so the scan follows the result set, not the library. `ocr_text_match` also tokenises the text once into a set and takes an exact hit before any edit-distance work. Test: `test_text_match_flags_are_looked_up_only_for_the_ids_given` (red without the id filter).

pixlstash/tasks/ocr_task.py
lindkvis: Fixed in b56ea88b. Added `ix_picture_ocr_unread` on `(ocr_text, deleted, text_score) WHERE ocr_text IS NULL`, on the model and in migration 0120 (guarded by name). The range on `text_score` is last so the index also serves the new `ORDER BY text_score DESC` (most text first, which also groups similar-length pages into a batch, per the note in the summary). No trailing `id`: the order is by score. Test: `test_the_finder_probe_reads_the_partial_index` asserts EXPLAIN QUERY PLAN names the index and has no temp B-tree (red with the columns changed).

frontend/src/components/views/OverlayDescriptionPanel.vue
lindkvis: Fixed in b56ea88b. The word list is now one tab stop with roving focus: Left/Right move word to word, Up/Down to the same position on the neighbouring line; arrows `preventDefault` + `stopPropagation`, edges included. Tests cover the tabindex pattern, each direction, and that the key never reaches a `window` listener; `onTabKeydown` and its `stopPropagation` are now tested too.

frontend/src/components/views/ImageOverlay.css
lindkvis: Fixed in b56ea88b. Every state now pairs line and halo: at rest a light halo at 0.13 under the 0.13 line, on hover the light halo at 0.9 (shared with the match halo). The selected box keeps its dark halo on hover. Block comment updated.

pixlstash/db_models/picture.py
lindkvis: Took the reword, in b56ea88b. Excluding the columns from the generic reader means a change to `field_allowlist`, which I'd rather not widen this PR into for a non-authz point, so the comment now says what is true: left out of metadata and search rows for size, served shaped by `/text`, and still readable raw through `/{field}` under the same picture scope. `OCR_FIELDS` moved up beside `ExportType`, per the member order.

lindkvis:
Review items from the summary that have no thread. Everything below is in b56ea88b unless it says otherwise.

**Blocking: rotate leaves wrong word boxes.** Fixed. `runRotate` now calls `pictureText.refresh()` beside `fetchFaceBboxes`/`fetchDetections`, guarded on the rotated picture still being the one on screen. The re-read returns `pending`, so the boxes go and the tab shows its spinner. The new test in `ImageOverlayRotate.test.js` expects an extra `GET /pictures/{id}/text` after a rotate.

**(c) Read again dedupe.** Agreed, the busy guard covers accidental repeats. The PR description note overstated it.

**Backend and ML**
- **5-letter band:** a one-edit match now also needs the first letter to agree. `loose` no longer matches `goose` and `shank` no longer matches `THANK`; both are in the tests. The four `MUST_NOT_MATCH` entries that shared no vocabulary were replaced with near misses (`rake`, `moon`, `shank`, `sound effects`).
- **Route default threshold:** §25 now says the SPA sends 0.1, and that a text-only match (0.35) is filtered out at the route's 0.5 default. I left the default alone, since it applies to every search caller.
- **`quad_boxes`/`labels` length mismatch:** now logs a warning with both counts and keeps the pairs that match.
- **Padded labels:** the label is stripped before `char_w` is computed. A padded case is pinned in the word-box test.
- **`ponytail:` caveat:** now names angled text, where the axis-aligned bounds put every word in the wrong place.
- **Batch grouping:** the probe orders by `text_score DESC`.
- **VRAM estimate:** captioning's estimate times the three beams, via `FLORENCE_OCR_NUM_BEAMS`, which the generate call uses too. A failed estimate now logs at `warning`. `read_text` re-raises out-of-memory so the runner retries.
- **Downscale:** the §7 task row notes that small body text may not survive 1024 px, and that such a picture is stored as read with nothing found.
- **POST on a deleted picture:** already refused. `fetch_picture_file_path` filters `deleted = False`, so a scrapheap picture gets a 404; `test_read_again_keeps_the_text_and_queues_an_urgent_read` asserts it. The finder/route claim overlap is real and harmless, since both write the same derived text. Left as is.
- **`box` length:** words are validated through `PictureTextWord` in the handler, with `box` constrained to exactly 4. A 3-element box is a fourth parametrised case and reads as no text.
- **READ_SAFE note:** added at `READ_SAFE_POST_PATHS`. `/text/read` must not be added there without a claim.
- **Partial `read_text` result:** tested. The picture the reader skipped is stored `""`, the other keeps its text.
- **Mutation evidence:**
  - Every new assertion was broken and seen red.
  - Removing the GET `/text` registry line goes further than predicted: the server's startup assertion refuses to boot, so every server test in both files errors and `test_coverage_matrix_document_matches_the_registry` fails.

**Frontend**
- **Read again focus:** the button stays mounted and uses `aria-disabled` while busy, so focus stays on it.
- **Read again bounce:** the backend now keeps the old text until the new read succeeds. The Text tab stays open with "Reading the text again…" (`aria-busy`) until the `ocr_text` frame arrives.
- **Copy/Clear:** gated on `selectedWords.length`, with a test.
- **Pressed-segment hover:** rule added.
- **`textSearchResults`:** released on every query change.
- **`readAgainBusy`:** tracked per picture, with a test.
- **Tokens:**
  - Word-box radius is `var(--radius-sm)`. That is rounder than the 1.5 it replaces; say if the smaller radius should come back as a commented one-off.
  - The chevron uses `--opacity-text-secondary`.
  - The sidebar wash is 0.2, the dark `--active-wash`.
  - The Read again button uses `--control-h-sm`.
  - The raw dark-surface olive is untouched, as you noted.
- **Test gaps 1–3:**
  - The box maths is now pure `wordBoxRect`/`wordLayerStyle`, tested for pixels, a null box and a wrong-length box.
  - Overlay tests cover box clicks selecting and revealing a word, and boxes clearing on a picture swap.
- **Count literal:** now `"18"`.
- **Declined: `viewBox="0 0 1 1"`.** With `preserveAspectRatio="none"` the `rx` stretches with the picture, so corners go oval and grow on large pictures, and fixing that needs the pixel size per box again. The recompute is plain arithmetic that runs only when the picture resizes.
- **Not done: e2e (gap 4).** It needs a picture that has been read, i.e. a model in the e2e run. Left for a follow-up.

**Registration order (`_text` before `_crud`).** Covered in effect: every `/text` test would get the `{field}` catch-all's 400 if the order flipped. There is no test named for it.

lindkvis:
Added b2e48b7c, a desktop dev fix that rode along at the owner's request: in dev mode the backend and CLI now get the checkout the app was built from first on `PYTHONPATH`. Before, a venv shared between checkouts ran whichever branch it had pixlstash installed from, which is how this branch's library (migrated to 0120) was opened by develop's 0119 code and refused. The branch was also rebased onto develop in place of the two merge commits; the tree is unchanged.

PR #1389, opened 2026-09-17
Show the picture ID in the Metadata panel

### LINKED ISSUE #1379: Ensure the Picture's Metadata section has the ID in it

lindkvis:
Closes #1379.

Adds an **ID** row at the top of the lightbox Metadata panel's Picture/Video information grid, plus a small vitest (verified red with the guard inverted) and a changelog fragment.

Note: the grid is two columns, so the new first cell shifts later rows by one (ID|Size, Aspect|MB, …). Left as is; the ID leading the grid reads naturally.


frontend/src/components/views/OverlayMetadataPanel.vue
```diff
@@ -338,6 +338,9 @@ const infoHeaderLabel = computed(() => {
 const pictureInfoEntries = computed(() => {
   if (!props.image) return [];
   const entries = [];
+  if (props.image.id != null) {
+    entries.push({ label: "ID", value: String(props.image.id) });
```
lindkvis, line 342: nit: as the PR description notes, unshifting this into the grid moves every later two-column pairing by one (ID|Size, Aspect|MB, ...). Reasonable call for a single always-short numeric value, and not a design-token issue since no new CSS is added — flagging only because "what a control shows" changes technically falls under the repo's ui-ux-expert review guidance. Not a blocker.

lindkvis (commented):
Reviewed as senior-frontend-developer, per docs/frontend_architecture.md (Component Catalogue / Overlay side-panels) and docs/design/visual-language.md (spacing tokens, §5).

Matches the issue: adds an ID row to `OverlayMetadataPanel`'s Picture/Video information grid, closes #1379.

**Correctness.** `pictureInfoEntries` gains `{ label: "ID", value: String(props.image.id) }` gated on `props.image.id != null`, so `id === 0` still renders (correct — `!=` catches only null/undefined) and a missing id renders nothing. The entry follows the exact shape/convention of every sibling entry (Size, MB, Smart score, ...) and isn't marked `fullWidth`/`clickable`, which matches the pattern (those flags are reserved for long text like file paths and locked-by lists, not a short numeric value) — correct call.

**Test.** I mounted the PR's test file against develop's current (pre-PR) `OverlayMetadataPanel.vue` (no tracked files changed, just a throwaway copy, removed after): the "shows the picture id as the first info row" test fails red as expected (`Size/10×20` instead of `ID/4217`), confirming the assertion genuinely catches the feature being removed. The second test (no id → no ID row) is a reasonable companion guard even though it can't distinguish "feature absent" from "feature present but id null" on its own. Mocks (`apiClient`, `api/pictures`) match the component's real import paths.

**Design/tokens.** No new CSS is added — the row reuses the existing `.metadata-info-item`/`.metadata-info-label`/`.metadata-info-value` classes, which already use `--space-*`/`--text-*` tokens. Nothing to flag here.

**Changelog.** `changelog.d/1379.md` is a genuinely user-visible addition (new info in an existing panel), format and voice match `changelog.d/README.md` and sibling fragments.

No blockers. One non-blocking nit inline on the column-shift tradeoff the PR description already calls out.

PR #1390, opened 2026-09-17
Workflows & Recipes B1: workflow identity (card key, stack key, type, differs-by)

lindkvis:
Step B1 of the v1.12 Workflows & Recipes plan: a pure derivation module, no schema, nothing wired up yet.

`services/workflow_identity.py` reads the stored hub document, whose assets are references rather than filenames. It refuses a raw graph.
- `slots`: model slots labelled by the loader's WL label at topology tier + widget; picture inputs left out.
- `guess_mark`: speed-LoRA names → structural, anything else recipe; the caller freezes the guess.
- `workflow_key`: topology + non-LoRA slots + structural LoRA slots (the card).
- `core_hash`: topology with plumbing, post-processing and LoRA loaders stripped, and the edges re-wired through them (the automatic stack).
- `workflow_type` and the `differs_by` chips. "plumbing only" requires identical graphs once plumbing is stepped through, with the same assets on the same loaders. Mutation-checked.

`workflow_hash.graph_key` now uses an extracted `node_labels`, and every hash is unchanged on every fixture graph. `workflow_io` exposes `is_picture_loader`. The glossary line (variant vs saved recipe) is in the services table.

**Scope.** The plan is one PR per step; issues #1394–#1410 cover the rest.

**Where this differs from the plan, or stops short of it**
- Slot labels refine until a round splits nothing, rather than stopping at 4 rounds. A chain of 12 LoRA loaders shared labels at 4 rounds.
- The mark guess favours precision over recall: whole words only, Hyper-SD rather than bare "hyper", and step counts of one or two digits.
- `differs_by` can also return "other models", and "N nodes differ" when models are swapped between loaders.
- Not handled yet:
  - the PixlStash checkpoint loader's `checkpoint_id` (#1416, now a B2 dependency);
  - UI-format and A1111 reductions;
  - pixel-space hires fix.
- `ImageScale*` is stripped as post-processing, as the plan lists, even when it downscales.


pixlstash/services/workflow_identity.py
```diff
@@ -0,0 +1,443 @@
...
+
+STRUCTURAL = "structural"
+RECIPE = "recipe"
+
+# The filename guess for a LoRA slot's mark. Speed LoRAs change how a workflow
+# samples (steps, CFG), so they are part of the workflow; anything else is
+# taken to be the look. Anchored on non-letters so "hyperrealism" and
+# "turbocharged_style" stay recipe LoRAs.
+_STRUCTURAL_LORA_RE = re.compile(
+    r"(?<![a-z])(lightning|turbo|hyper|lcm)(?![a-z])|(?<![a-z0-9])\d+[-_ ]?steps?(?![a-z])"
```
lindkvis: **should-fix.** The guess is frozen on first sight, so its precision matters more than its recall. A false "structural" pulls a character LoRA into the card key, and every new character then splits the card. The regex currently gets these wrong:
- `alice_3000steps.safetensors` and `alice-v2-1500-steps` → structural. Trainers put step counts in file names all the time.
- `hyper_detailed_skin.safetensors` → structural. `hyper` followed by `_` passes the lookahead.
- `…lightx2v_cfg_step_distill…` (the Wan speed LoRA), `causvid`, `sdxl_flash`, `tcd` → recipe.

Suggestion: `\d{1,2}[-_ ]?steps?` (speed LoRAs are 1-16 steps), and `hyper[-_ ]?(sd|flux|sdxl)` instead of bare `hyper`. Consider adding `lightx2v|causvid|dmd2?|tcd|pcm|flash`. Add the two false positives to the parametrised test.

One more gap: removing the `(?<![a-z])` lookbehind keeps the suite green. A case like `superturbo_style` → recipe would pin it.

pixlstash/services/workflow_identity.py
```diff
@@ -0,0 +1,443 @@
...
+def slots(document: dict) -> list[Slot]:
+    """Every model slot a stored document names, pictures excluded.
+
+    Args:
+        document: A stored structural document, assets as references.
+
+    Raises:
+        WorkflowGraphError: The document holds no usable node.
+    """
+    nodes = reduce_api_graph(document)
```
lindkvis: **should-fix.** This reads only values starting with `asset:`. If a caller passes the raw API graph instead of the stored document, which is an easy mix-up since `workflow_type` and `core_hash` accept either, `slots()` returns `[]`. `workflow_key` then collapses every checkpoint into one card, and nothing complains. I checked: `workflow_key(topology_hash(g), slots(g), set())` is identical for `base` and `z` checkpoints.

A stored document reduces with every widget value either `None` or an `asset:` reference (the `*_sha256` fields), so a cheap guard is enough:
```python
if any(v is not None and not v.startswith(ASSET_REFERENCE_PREFIX)
       for n in nodes.values() for _, v in n.widgets):
    raise WorkflowGraphError("slots() needs a stored document, not a raw graph")
```

pixlstash/services/workflow_identity.py
```diff
@@ -0,0 +1,443 @@
...
+    topology = drop_widgets(nodes)
+    labels = node_labels(topology, rounds=len(topology))
```
lindkvis: **nit.** `rounds=len(topology)` runs a fixed n rounds. It never checks whether the refinement has stabilised, which the PR body and the docs row say it does. The cost is O(n^2) digests: I measured 0.29 s for a 307-node chain, which is fine today but grows fast for large graphs in a backfill. Stopping once the number of distinct labels stops growing is a few lines and gives the same separation.

More important for B2: a label at full refinement encodes the whole graph, so adding one PreviewImage changes **every** slot label (I checked). Labels, and therefore frozen marks and `(slot label, input)` overrides, only mean something within a single topology. `workflow_key` includes the topology hash, so that holds, but please say it in the `Slot` docstring so B2 keys the marks table accordingly.

lindkvis (commented):
Senior backend review of B1 (workflow identity). Posted as a comment, not an approval.

**Verdict: sound design and a clean, well-scoped PR. Three should-fix items before B2 builds on it, one of them a wrong "plumbing only".** Nothing blocking.

What I checked:
- Ran `tests/test_workflow_identity.py` (35 passed) plus `test_workflow_library.py` and `test_workflow_io.py` (203 passed), so extracting `node_labels` did not move any hash.
- ruff check and ruff format are clean on all three Python files.
- The new test file is in ci.yml in the right alphabetical place, and `test_every_test_file_is_classified` passes.
- The services-table row is hand-written, and re-running `scripts/render_backend_architecture.py` leaves the doc unchanged.
- No try/except and no silent pass anywhere: malformed input raises `WorkflowGraphError` (empty graph, or nothing left once stripped). A dangling link keys without error, which matches `workflow_hash`.
- **Mutation checks.** Each of these turns the suite red: dropping the wiring comparison in the plumbing-only check, dropping `not loras_differ`, dropping `not chips`, dropping `unclassified += plumbing`, putting slot labels back to the default 4 rounds, and ignoring structural marks in `workflow_key`. One mutation stays green: removing the `(?<![a-z])` lookbehind in the mark regex (see the inline comment).

Findings (details inline):
1. **should-fix**: `differs_by` compares assets as a multiset, not by position. Two checkpoints swapped between loaders (base and refiner) come back as `[]`, and with a PreviewImage added they come back as `["plumbing only"]`. That is the "never a guess" case.
2. **should-fix**: the mark guess over-matches. `\d+ steps` catches training-step counts (`alice_3000steps` gives structural), and bare `hyper` catches `hyper_detailed_skin`. Because the guess is frozen, a wrong first guess sticks, and it pulls a character LoRA into the card key.
3. **should-fix**: `slots()` handed a raw graph instead of a stored document quietly returns `[]`, so every checkpoint collapses into one card. A one-line guard fixes it.
4. **should-fix (a follow-up is fine)**: `PixlStashCheckpointLoader` addresses its model by `checkpoint_id`, which has no extension, so the structural tier nulls it. With the house loader, two different checkpoints therefore share a card.
5. nits: the slot-label rounds and the "until stable" wording, a duplicate `_digest`, dead plumbing class names, and `workflow_type` not seeing the house picture loader.

pixlstash/services/workflow_identity.py
```diff
@@ -0,0 +1,443 @@
...
+
+    label: str
+    node_id: str
+    class_type: str
+    widget: str
+    asset: str
+    is_lora: bool
+
+
+def _digest(payload: object) -> str:
```
lindkvis: **nit.** This duplicates `workflow_hash._digest` minus `default=str`. Import it, or make it public, instead of keeping two copies.

pixlstash/services/workflow_identity.py
```diff
@@ -0,0 +1,443 @@
...
+    plumbing = count(added, PLUMBING) + count(removed, PLUMBING)
+    unclassified += sum(
+        count(counter, group) for counter in (added, removed) for group in (LORA, None)
+    )
+    if (
+        plumbing
+        and not chips
+        and not unclassified
+        and not loras_differ
+        and _stripped_key(cover_nodes, {PLUMBING})
```
lindkvis: **should-fix.** The "plumbing only" guarantee rests on `_stripped_key`, which drops widgets, plus `_assets`, which is a **multiset** of `(widget, asset)`. Neither ties an asset to its loader. Reproduction: two `CheckpointLoaderSimple` nodes (base feeding sampler 1, refiner feeding sampler 2). Swap the checkpoints in the member:
- `differs_by(cover, member)` returns `[]`, even though the two workflow keys differ.
- Add a PreviewImage to the member and it returns `["plumbing only"]`.

That is exactly the hide-a-real-difference case the docstring rules out.

The smallest fix is to key the plumbing-stripped graph **with** its widgets, i.e. `graph_key` over nodes that keep `node.widgets`. A stored document's widgets are its asset references, so one equality then covers wiring, assets and LoRAs by position, and `loras_differ` is no longer needed on this branch. Also, don't return `[]` when the structural keys differ: fall through to a generic chip.

pixlstash/services/workflow_identity.py
```diff
@@ -0,0 +1,443 @@
...
+)
+
+# A widget naming the picture a run starts from. It is an asset in the
+# structural hash, but an input rather than part of the workflow, so it never
+# reaches the card key: an img2img workflow is one card whatever it was fed.
+# By name, because the stored document holds only a reference and cannot say
+# whether it was a picture or a model. A custom node naming its input
+# picture some other way still splits cards per picture.
+_PICTURE_WIDGET_RE = re.compile(r"(^|_)(image|images|video|mask)(_|$)")
+_CHECKPOINT_WIDGETS = frozenset({"ckpt_name", "unet_name"})
```
lindkvis, line 82: **should-fix, a follow-up is fine.** `PixlStashCheckpointLoader` (ComfyUI-PixlStash) names its model by `checkpoint_id`, a plain id with no file extension. `structural_widget_value` nulls it, so it never becomes a slot. I checked: two graphs using that loader with ids `12` and `99` get the **same** `workflow_key`, and `other checkpoint` can never fire for them. That breaks this PR's headline rule for the house loader. The VAE and CLIP loaders use `*_sha256`, so they are fine. The root fix is in `workflow_hash`, treating `checkpoint_id` as a topology asset, which would re-key structural hashes. At minimum, list it under "not handled yet".

pixlstash/services/workflow_identity.py
```diff
@@ -0,0 +1,443 @@
...
+
+PLUMBING = "plumbing"
+UPSCALE = "upscale"
+FACE_DETAILER = "face_detailer"
+LORA = "lora"
+
+_PLUMBING_CLASSES = frozenset(
+    {"PreviewImage", "Reroute", "Note", "MarkdownNote", "GetNode", "SetNode"}
```
lindkvis: **nit.** `Reroute`, `Note`, `MarkdownNote`, `GetNode` and `SetNode` never reach a stored document. API format has no such nodes, and `reduce_ui_graph` removes UI-only and passthrough classes before storing. So these entries are dead, and the `note` case in the stacking test covers a graph that cannot occur. They're harmless, but PreviewImage and `Primitive*` are the only live plumbing.

pixlstash/services/workflow_identity.py
```diff
@@ -0,0 +1,443 @@
...
+    and an img2img graph may still carry an empty latent. ``None`` when none
+    applies.
+    """
+    nodes = reduce_api_graph(document)
+    classes = {node.class_type for node in nodes.values()}
+    if "ImagePadForOutpaint" in classes:
+        return "outpaint"
+    if classes & {"VAEEncodeForInpaint", "InpaintModelConditioning"}:
+        return "inpaint"
+    has_picture_input = any(cls.startswith("LoadImage") for cls in classes)
```
lindkvis: **nit.** `LoadImage*` is the only picture source recognised here and in `_fed_by_picture`. A graph built on `PixlStashPictureLoader` therefore reads as txt2img, or `None`. Inpaint via `SetLatentNoiseMask` also reads as img2img. Fine for B1, but worth a known-gap line.

pixlstash/services/workflow_identity.py
lindkvis: Fixed in 414abf73. Step counts are now at most two digits, and bare `hyper` became `hyper[-_ ]?(sd|sdxl|flux)`. I added `lightx2v|causvid|dmd2?|tcd|pcm` and `sdxl[-_ ]?flash`, all as whole words. I left out bare `flash`, because it would catch style LoRAs like `flash_photography`. The parametrised test now has `alice_3000steps`, `alice-v2-1500-steps`, `hyper_detailed_skin` and `superturbo_style` as recipe, and the lightx2v/causvid/sdxl_flash names as structural. Removing the lookbehind, reverting to bare `hyper`, and going back to `\d+` each turn the suite red.

pixlstash/services/workflow_identity.py
lindkvis: Fixed in 414abf73. Every public function now reduces through a shared `_reduce`, which raises `WorkflowGraphError` when a widget names an asset by value. That covers `core_hash` and `workflow_type` too, so none of the four accepts a raw graph any more. Test: `test_a_raw_graph_is_refused_rather_than_read_as_having_no_models`, red with the guard removed.

pixlstash/services/workflow_identity.py
lindkvis: Fixed in 414abf73. `node_labels(..., rounds=None)` now stops at the first round that splits no class, capped at n. The `Slot` docstring says a label only means something within one topology, and why, so B2 keys the marks and overrides with the topology. Setting it back to 4 rounds turns the 12-loader chain test red.

pixlstash/services/workflow_identity.py
lindkvis: Fixed in 414abf73: it imports `workflow_hash._digest` and the local copy is gone.

pixlstash/services/workflow_identity.py
lindkvis: Fixed in 414abf73, as you suggested. `_reduce` puts each asset reference back onto its widget; `reduce_api_graph` nulls references under filename widgets because they have no extension. The plumbing-stripped graph is keyed **with** those widgets, so one equality covers wiring and assets by position, LoRAs included, and `loras_differ` is gone. When no chip fires but the graphs differ, it now falls through to "N nodes differ" and never returns `[]`. Your base/refiner swap is `test_models_swapped_between_loaders_are_a_difference`, both bare and with a PreviewImage. Keying without widgets turns it red, and so does dropping the fall-through.

pixlstash/services/workflow_identity.py
lindkvis: Agreed. It's a follow-up, filed as #1416. The root fix is in `workflow_hash`, and it re-keys structural hashes for graphs on that loader. I added #1416 to B2 (#1394) as a hard dependency, so the backfill doesn't key cards wrong first and then re-key them.

pixlstash/services/workflow_identity.py
lindkvis: Fixed in 414abf73. `_PLUMBING_CLASSES` is now just `PreviewImage` (plus `Primitive*`), with a comment on why. The `note` stacking case is gone.

pixlstash/services/workflow_identity.py
lindkvis: Fixed in 414abf73. `workflow_io` now exposes its picture-loader rule as `is_picture_loader`, and both modules use it, so `PixlStashPictureLoader` counts as a picture source. `SetLatentNoiseMask` reads as inpaint. Test: `test_the_pixlstash_picture_loader_and_a_noise_mask_are_picture_sources`.

PR #1391, opened 2026-09-17
Private-address guardrail: skip node_modules

### LINKED ISSUE #1383: Private-address guardrail walks node_modules, so it fails on any machine with the Electron deps installed
`tests/test_architecture_guardrails.py::test_no_unsanctioned_private_address_literal` walks `electron/` recursively (it is in `_PRIVATE_ADDRESS_ROOTS`) and nothing excludes `node_modules`. `@types/node`'s `.d.ts` files document `net`, `dgram` and `os` with RFC 1918 examples, so the moment `electron/node_modules` exists the guardrail reports ~20 offenders and the suite is red locally.

Reproduce on a clean `develop`, with no branch and no local edits:

```
cd electron && npm install
cd .. && python -m pytest tests/test_architecture_guardrails.py::test_no_unsanctioned_private_address_literal
```

Offending lines are all vendored, e.g. `electron/node_modules/@types/node/net.d.ts` ("blockList.addRange(...)", "Subnet: IPv4 ...") and `electron/node_modules/@types/node/os.d.ts` (`networkInterfaces()` sample output).

CI evidently does not install those deps in the job that runs this test, which is why it has stayed green there — it fails only for a developer who has built or run the desktop app.

The scan already knows about this class: `_PRIVATE_ADDRESS_FLAT_DIRS` carries the comment *"Files only. A subdirectory here would be walked, and the one sitting in frontend/ is node_modules"*, so the flat-dir case was handled and the recursive-root case was not.

Fix: skip any path with a `node_modules` component in `_private_address_offenders` (or add it to `_PRIVATE_ADDRESS_SKIP`, which is prefix-matched today and so would need `electron/node_modules` and `frontend/node_modules` spelled out). The guardrail's teeth test should gain a case proving a vendored path is skipped *and* a first-party one beside it is still caught, so the skip cannot quietly widen.

Not urgent for CI, but it costs every local full-suite run a false red, and a false red that everyone learns to ignore is how a true one gets ignored too.

lindkvis:
Fixes #1383.

The private-address scan walks `electron/` recursively, so after `npm install` it read the vendored `@types/node` .d.ts files and failed locally. It now skips any path with a `node_modules` component (a whole component, not a substring). The teeth test gains a vendored file that must be skipped, plus a sibling first-party file and a `node_modules_notes.md` that must still be caught. I checked it both ways: removing the skip, or matching a substring, turns the test red.

Not done: pruning the walk. `rglob` still lists everything under node_modules before skipping it. That cost about 0.2 s on a 35k-entry tree, which isn't worth an `os.walk` rewrite.


tests/test_architecture_guardrails.py
```diff
@@ -2448,11 +2451,14 @@ def _private_address_offenders(root: Path, repo_root: Path) -> list[str]:
     named = root.is_file()
     paths = [root] if named else sorted(root.rglob("*"))
     for path in paths:
+        rel = path.relative_to(repo_root)
+        # A component, not a substring: node_modules_notes.md is first-party.
+        if "node_modules" in rel.parts:
```
lindkvis, line 2456: Confirmed this is a component check (`in rel.parts`), not a substring test — verified with a real `electron/node_modules/@types/node/net.d.ts` (skipped) beside a sibling first-party `electron/*.js` file carrying the same literal (still caught), outside `tmp_path`. Matches the issue's exact reproduction path.

lindkvis (commented):
Reviewed as senior-backend-developer. Verdict: sound, no blockers.

Ran the guardrail on a throwaway worktree (`pr/1391`, not the shared `develop` checkout) with `.venv/bin/python -m pytest tests/test_architecture_guardrails.py -k private_address` — 3 passed.

Mutation checks (planted, confirmed red, removed, confirmed green again):
- **Real-tree node_modules skip**: created `electron/node_modules/@types/node/net.d.ts` with an unsanctioned RFC 1918-shaped literal alongside a sibling first-party `electron/*.js` file with the same literal. Vendored file was silently skipped; the first-party file was still reported. This reproduces the issue's exact trigger (`electron/ && npm install`) outside `tmp_path`, not just inside the teeth test.
- **Coverage not weakened**: planted an unsanctioned literal directly under `pixlstash/` and `frontend/e2e/` (two of the real `_PRIVATE_ADDRESS_ROOTS`, outside `tmp_path`) — `test_no_unsanctioned_private_address_literal` went red on both, confirming the skip is scoped to `node_modules` only.

`"node_modules" in rel.parts` is a path-component check, not a substring test, so it can't false-skip something like a hypothetical `node_modules_helper/` directory or catch `node_modules_notes.md` (the PR's own added case, which I reproduced independently). Sibling-guardrail sweep: no other scan in this file, or elsewhere under `scripts/`/`tests/`, recurses into `electron/` or `frontend/src`/`frontend/e2e`, so nothing else shares this bug.

Minor, non-blocking: `rglob("*")` still fully expands `node_modules` before the per-entry skip runs (the PR body already discloses this, ~0.2s on a 35k-entry tree). Fine to leave as-is — an `os.walk`-with-prune rewrite isn't worth it for a one-time local-dev cost, and it keeps the diff minimal.

Stand-in-values check: `_TEETH_OFFENDER` (used in the new vendored-file cases) is derived from `tests/network_vectors.py`'s `LAN_IPV4` via `.replace(...)`, not a hand-written literal, consistent with CLAUDE.md's guardrail-vocabulary rule. No real private IP or home path appears in the diff.

tests/test_architecture_guardrails.py
```diff
@@ -2538,12 +2544,25 @@ def test_private_address_guardrail_has_teeth(tmp_path):
         + "\n"
     )
     (tmp_path / "longer.md").write_text(f"the gateway is {LAN_IPV4}.7\n")
+    # Vendored code is skipped by path component, and only that: a first-party
+    # file beside it, or one merely named after it, is still read.
+    vendored = tmp_path / "app" / "node_modules" / "@types" / "node"
+    vendored.mkdir(parents=True)
+    (vendored / "net.d.ts").write_text(f"// Subnet: {_TEETH_OFFENDER}\n")
+    (tmp_path / "app" / "main.js").write_text(f"// host {_TEETH_OFFENDER}\n")
+    (tmp_path / "app" / "node_modules_notes.md").write_text(f"{_TEETH_OFFENDER}\n")
```
lindkvis, line 2553: Good case — `node_modules_notes.md` proves the skip can't be fooled by a substring match on the directory name. I independently re-ran this style of check against the real repo tree (not just `tmp_path`) with the same result.

tests/test_architecture_guardrails.py
```diff
@@ -2448,11 +2451,14 @@ def _private_address_offenders(root: Path, repo_root: Path) -> list[str]:
     named = root.is_file()
     paths = [root] if named else sorted(root.rglob("*"))
```
lindkvis, line 2452: Nit, non-blocking: the walk still lists everything under `node_modules` before the per-entry skip filters it (no `os.walk`-style prune). The PR body already discloses the ~0.2s cost and calls the rewrite not worth it — agreed, this keeps the diff minimal and the existing named-roots-plus-skip-list style consistent with the rest of the file.

PR #1392, opened 2026-09-17
Tooltips no longer take the click meant for what they cover

### LINKED ISSUE #1380: The new tooltips in develop often overlaps with where you want to click
Image

lindkvis:
Fixes #1380.

The shared tooltip was made hoverable (`pointer-events: auto` plus a 150 ms close grace). A tip usually sits on the neighbouring control, like the next filmstrip thumb or the row above. When you moved onto that control, the tip stayed open under the pointer and caught the click.

- `Tooltip.vue`: drop `pointer-events: auto` so Vuetify's own `none` applies again, and drop the close grace. A tip now closes when the pointer leaves its control, and clicks go through to whatever is underneath.
- `designDrift.test.js`: add a guard that fails if a tooltip rule turns pointer events back on. I checked that it goes red when that happens.
- Updated the comments and `frontend_architecture.md`.

**Trade-off:** tips no longer meet the WCAG 1.4.13 "hoverable" rule, because you can't move the pointer onto a tip to keep it open. It stays open while the pointer is on its control, and Escape and focus behave as before. A tip is never the only place its information appears, which is our rule, and no tip has anything to click inside it.

A review suggested keeping tips hoverable but closing them on press. That doesn't help here: the press lands on the tip, not on the control under it, so the click is still lost. Another option is to position tips per call site, but that fixes one spot at a time. If the design side wants hoverable tips back, the way is to place tips so they never cover a neighbour. That's a bigger change.

There's no behaviour test for the CSS, because jsdom doesn't compute styles. The source guard covers it instead.


frontend/src/components/widgets/Tooltip.vue
```diff
@@ -88,10 +81,14 @@ import { VTooltip } from "vuetify/components";
...
- * painted on the app's own surface, and given an open delay. Vuetify's
- * default delay is none, so a tip strobed open as the pointer crossed a bar.
+ * already opens on focus and on touch, painted on the app's own surface, and
+ * given an open delay. Vuetify's default delay is none, so a tip strobed open
+ * as the pointer crossed a bar.
+ *
+ * The surface is not hoverable: it keeps Vuetify's `pointer-events: none`, and
+ * closes as the pointer leaves the control. A tip lands on the neighbouring
+ * control (the next filmstrip thumb, the row above), and a hoverable one took
+ * the click meant for it (#1380).
```
lindkvis, line 91: **[Info / a11y trade-off, accepted]** Hoverable has no exception in SC 1.4.13, so this is a real AA gap, and the PR says so. What remains mitigates it: the tip stays open while the pointer is on its control, Escape closes it, focus opens it, and a tip is never the only home of its information. The people who lose out are screen-magnifier users. At high zoom they may need to move the pointer onto a long tip to pan it into view, and the ScrapheapSection bullet list is the longest one. I agree a pointer "bridge" would not help, because the overlap is the problem. If hoverable comes back, it should come with placement that keeps the tip off neighbouring controls.

frontend/src/styles/designDrift.test.js
```diff
@@ -61,6 +61,16 @@ describe("design drift", () => {
     ).toEqual([]);
   });
 
+  // A tip lands on the neighbouring control; if it takes the pointer, it takes
+  // that control's click (#1380). Vuetify's own rule is `pointer-events: none`.
+  it("leaves the tooltip surface click-through", () => {
+    expect(
+      hits(
+        /(?:app-tooltip|v-tooltip)[^{]*\{[^}]*pointer-events:\s*(?!none\b)\w[^;]*/g,
```
lindkvis, line 69: **[Nit]** I checked that the guard goes red when `pointer-events: auto` is put back. It only sees rules whose selector names `app-tooltip` or `v-tooltip`, though. Two things would get past it: a parent's `:deep(.v-overlay__content)` rule, or a style set from JS through `content-props` (the file header already says JS styles are not scanned). Neither exists today, and the regex is a fair size for the risk. Putting `close-delay` back is also unguarded, but on its own that only makes the tip linger for 150ms and does not take clicks. Fine to leave as is.

lindkvis (commented):
Reviewed with a frontend lens and a UX/accessibility lens.

**Verdict: good to merge. Nothing blocking.**

- **Root cause:** fixed where every tip goes through. `pointer-events: auto` and the close grace are both gone from `Tooltip.vue`, `HelpTip` is a preset of it, and the design-drift test still forbids a second `v-tooltip`.
- **Consumers:** I swept every `<Tooltip>` with default-slot content. There are four name `<span>`s in SideBar and one plain `<ul>` of text in ScrapheapSection. None has a link, button or anything else to click or select, so none relied on a hoverable tip. No `open-on-hover` menu or hover card shares the problem.
- **Mutation check:** with `pointer-events: auto` put back in `.app-tooltip`, `leaves the tooltip surface click-through` goes red. Restored, it passes. `designDrift.test.js` and `Tooltip.test.js` pass (17 tests).
- **The design-drift change** only adds a test and loosens nothing.

Findings are inline:
- [Info] the WCAG 1.4.13 trade-off
- [Nit] what the guard can't see

Pre-existing, not this PR: `docs/design/buttons.md`, which the architecture doc cites as the tooltip spec, is not tracked in the repo.

PR #1393, opened 2026-09-17
Lightbox rotate no longer snaps back to the old orientation

### LINKED ISSUE #1382: Still have issue with rotation in the ImageOverlay
Do a left or right rotation inside the ImageOverlay. Sometimes the image overlay updates briefly, but the end result is always the same: it reverts back to the original look and you have to exit the ImageOverlay and go back again to see the rotated picture

lindkvis:
Fixes #1382.

After a rotate, the grid refreshes its list, and the lightbox re-resolves the open picture from the snapshot it froze when it opened. That snapshot still had the old orientation, so the image URL lost its `?v=o<n>` cache-buster and the browser repainted the unrotated file. Moving to a neighbour and back hit the same stale entry.

The fix: when the lightbox's metadata read brings a new orientation, it also writes it into the frozen snapshot. Two tests in `ImageOverlayRotate.test.js` cover this: one for the grid refresh, one for going to a neighbour and back. Both fail without the fix.

Known gap, not addressed here: an undo or redo of a rotate while the lightbox is open is still not re-read, because the `pixels` socket event is deferred while the overlay is open and the overlay has no signal for it. That needs its own signal from App.vue down to the overlay, so it belongs in a separate change.


frontend/src/components/views/ImageOverlay.vue
```diff
@@ -3768,6 +3768,26 @@ async function fetchOverlayMetadata(imageId) {
       merged.metadata = { ...currentMeta, ...dataMeta };
     }
     image.value = merged;
+    // Carry the orientation into the snapshot frozen on open as well. Every
+    // re-resolve of the picture reads that snapshot - the grid's list refresh
+    // after a rotate, and moving to a neighbour and back - so a stale entry
+    // there puts the pre-rotate `?v=` back on the <img>.
+    const frozen = frozenAllImages.value;
```
lindkvis, line 3791: **Medium:** this patches one of the two sources `setOverlayImageById` resolves from. It tries `allImageById` (this snapshot) first, then `filmstripImageById`. When a stack is expanded in the filmstrip and its members came from `listStackPictures` (so they live in `overlayExpandedStackMembers`, not in `frozenAllImages`), `frozenIndex` is -1 here and nothing gets patched. The grid refresh after the rotate then re-resolves the stale member row, and the picture snaps back the same way. I reasoned this from the code; I couldn't drive filmstrip expansion in the stubbed test harness.

Smallest fix: in the `isSameImage` branch of `setOverlayImageById`, keep `orientation` (and `pixel_sha`) from `image.value`, the same way `score` is kept. That covers every source on the grid-refresh path. This snapshot patch is still what fixes neighbour-and-back.

frontend/src/components/views/ImageOverlay.vue
```diff
@@ -3768,6 +3768,26 @@ async function fetchOverlayMetadata(imageId) {
...
+      frozenIndex >= 0 &&
+      merged.orientation != null &&
+      frozen[frozenIndex].orientation !== merged.orientation
+    ) {
+      const next = frozen.slice();
+      next[frozenIndex] = {
+        ...frozen[frozenIndex],
+        orientation: merged.orientation,
+      };
+      frozenAllImages.value = next;
```
lindkvis, line 3805: **Low:** only `orientation` is carried into the snapshot. The filmstrip builds `thumbSrc` from the row's `thumbnail` URL, whose version token only the server recomputes, so the rotated picture's filmstrip tile stays unrotated until the lightbox closes, even though the grid has already pushed the fresh row. The freeze is there for membership, not content. A follow-up could merge live `allImages` rows into the snapshot by id on update, keeping membership frozen. That fixes the thumbnail and would make this orientation-only patch unnecessary for the grid-refresh case. Fine as a separate change.

lindkvis (commented):
Reviewed as senior frontend.

**Verdict:** this is the actual root cause for #1382, and the fix is minimal: no timers, no fallback. After a rotate, `overlay-change {pixels}` makes the grid push a new `allImages`. `applyAllImagesUpdate` then calls `setOverlayImageById(sameId)`, which spreads the row frozen on open over `image.value`. The same-image branch keeps description, smartScore, score and tags, but not `orientation`, so `?v=o<n>` drops and the `` repaints the bytes it already decoded. The earlier rotate fixes (#966, #1040) handled the cache-buster and URL pinning but never this re-resolve, which is why they did not hold.

**Verified:** all 16 tests in `ImageOverlayRotate.test.js` pass. With only the component hunk reverted, both new tests fail and the rest pass. A probe where the rotate finishes while you're on the neighbour and you then come back also shows `?v=o6` with the fix and the bare URL without it, so leaving mid-rotate is covered. No new watchers or timers to clean up.

**Findings**
1. **Medium:** stack members loaded via `listStackPictures` are not in the frozen snapshot, so they can still snap back (inline).
2. **Low:** the filmstrip tile of the rotated picture keeps its pre-rotate thumbnail for the lightbox's lifetime. It's the same stale-snapshot problem (inline).
3. **Low, follow-up:** undo/redo of a rotate while the lightbox is open gives exactly the #1382 symptom. The PR body names this gap; please file it as an issue so it isn't lost when #1382 closes.

frontend/src/components/views/ImageOverlay.vue
lindkvis: Fixed in 66c4a69a. `setOverlayImageById` now keeps `orientation` and `pixel_sha` from `image.value` when it re-resolves the same picture, next to `score`, so a row the snapshot patch never touched can no longer undo the turn. The snapshot patch stays in place for going to a neighbour and back.

I could not drive filmstrip stack expansion in this harness either, so the new test covers the same code path a different way: the open picture is missing from the frozen snapshot, and the grid then pushes a stale row for it. It fails without the fix.

frontend/src/components/views/ImageOverlay.vue
lindkvis: Agreed, but I am not doing it in this PR. Merging live rows into the snapshot also changes what the stack-signature logic in `applyAllImagesUpdate` reads, which can move the overlay to a new stack leader. That needs its own look. I am leaving this thread open to track the follow-up (the filmstrip tile stays unrotated until the lightbox closes).

PR #1412, opened 2026-09-17
Refuse importing JSON that is not a ComfyUI workflow

lindkvis:
Any JSON object dropped on the Workflows view (or put in the watched workflows folder) was stored as a workflow.

`check_comfy_workflow` now runs first in `_store_workflow`, so both the import route and the inbox go through it:

- **UI format:** a `links` list and a non-empty `nodes` list, every node with an `id` and a `type`. Requiring `links` is what shuts out other node-graph exports (React Flow, n8n).
- **API format** (bare or under `prompt`): at least one node, and every entry other than a `pixlstash_*` key has a `class_type` and an `inputs` object.

The route answers 400 with the reason, which the drop notice shows ("Could not add x.json: not a ComfyUI workflow."). The inbox logs the file and leaves it where it is, the same as a file that isn't valid JSON.

I checked it against every workflow JSON I could find: the fixtures, the built-ins and a few hundred real ones. None were refused.

Review notes:
- An empty canvas (`nodes: []`) and an API graph with a stray non-node top-level key are now refused. That is on purpose, since neither can run.
- `{"prompt": graph, ...}` still allows sibling keys, because ComfyUI's own `/prompt` payload has them (`client_id`, `extra_data`).
- A non-workflow file left in the inbox is warned about again on every reconcile. That matches how a broken JSON file there already behaves.


pixlstash/utils/comfyui_utilities.py
```diff
@@ -544,6 +544,72 @@ def is_comfy_workflow(value: Any) -> bool:
...
+    Raises:
+        NotAWorkflowError: *value* is not a workflow; the message says why.
+    """
+    if not isinstance(value, dict):
+        raise NotAWorkflowError("not a ComfyUI workflow: not a JSON object")
+    if "nodes" in value:
+        nodes = value["nodes"]
+        if not isinstance(nodes, list) or not nodes:
+            raise NotAWorkflowError("not a ComfyUI workflow: it has no nodes")
+        if not isinstance(value.get("links"), list):
```
lindkvis: **Low:** in the frontend's schema-version-1 serialiser (`LGraph.asSerialisable`), `links` is `this._links.size ? [...] : undefined`, so JSON drops the key when the root graph has no links. That covers a single-node graph, or a pipeline collapsed into one subgraph node whose links all live in `definitions.subgraphs`. The normal Save/Export path (`serialize()`, version 0.4) always writes `links: []`, so this is a narrow case today; it only bites a v1 document produced by a tool or a future default.

Suggestion: treat a missing `links` as fine when the document also carries a litegraph marker that React Flow/n8n exports lack, e.g. `"links" in value or isinstance(value.get("state"), dict) or "last_node_id" in value`. Both existing React Flow/n8n reject cases stay red under that.

pixlstash/utils/comfyui_utilities.py
```diff
@@ -544,6 +544,72 @@ def is_comfy_workflow(value: Any) -> bool:
...
+            isinstance(node, dict)
+            and _is_name(node.get("class_type"))
+            and isinstance(node.get("inputs"), dict)
+        )
+    ]
+    if not entries or len(strays) == len(entries):
+        raise NotAWorkflowError("not a ComfyUI workflow")
+    if strays:
+        raise NotAWorkflowError(
+            f"not a ComfyUI workflow: {strays[0]!r} is not a node with a "
```
lindkvis: **Nit:** `strays[0]!r` is an arbitrary client key, echoed without a length limit into the 400 detail, the drop notice and the warning log. Something like `{strays[0][:60]!r}` keeps the message useful without letting a pathological key flood the notice or the log.

pixlstash/utils/comfyui_utilities.py
```diff
@@ -544,6 +544,72 @@ def is_comfy_workflow(value: Any) -> bool:
...
+        if not isinstance(value.get("links"), list):
+            raise NotAWorkflowError("not a ComfyUI workflow: it has no links list")
+        for index, node in enumerate(nodes, start=1):
+            if not (
+                isinstance(node, dict)
+                and node.get("id") is not None
+                and _is_name(node.get("type"))
+            ):
+                raise NotAWorkflowError(
+                    f"not a ComfyUI workflow: node {index} has no id or type"
```
lindkvis: **Nit:** `index` is the node's 1-based position, but in a ComfyUI graph "node 3" reads as node id 3. That points the user at the wrong node when ids are not dense, which is the normal case after edits. Consider `the {index}th node in the list` or similar, or naming the id when there is one.

tests/test_workflow_io.py
```diff
@@ -737,12 +742,49 @@ def test_an_unfileable_or_too_deep_import_is_handled(import_route):
...
+def test_a_document_that_is_not_a_workflow_is_refused(import_route, document):
+    call, user_dir, _built_in, _hub = import_route
+    with pytest.raises(HTTPException) as refused:
+        call(name="unrelated", workflow=document)
+    assert refused.value.status_code == 400
+    assert "not a ComfyUI workflow" in refused.value.detail
+    assert not (user_dir / "unrelated.json").exists()
+
+
+def test_both_formats_are_accepted_with_pixlstash_keys_or_a_prompt_wrapper():
```
lindkvis, line 782: **Nit:** the accept side covers one UI fixture and one synthetic API graph. A parametrize over every file in `BUILT_IN` and `UI_FIXTURES` would cost milliseconds, and it would turn red the day a tightening of this check refuses a workflow we ship. Right now that promise rests only on the manual sweep described in the PR body.

lindkvis (commented):
**Verdict: good to merge. No blockers**; one low finding worth a look, plus nits.

Reviewed as backend + ComfyUI format review.

**What holds**
- Validation sits where every import path goes through. `_store_workflow` has two callers: the import route, and `Server._store_inbox_workflow` (the inbox). Nothing else writes to the user workflow folder. The Workflows view drop only calls the import route, and it only takes `.json`, so a workflow embedded in a PNG is not an import path.
- `NotAWorkflowError` subclasses `ValueError`, so `reconcile`'s existing `except (OSError, ValueError, RecursionError)` logs the file and leaves it. No silent swallow.
- The 400 detail reaches the drop notice through `errorMessage`. The check isn't recursive, so the deep-nesting 400 still fires; the updated test proves that by wrapping the deep doc in a node.
- No route added or changed, so `ROUTE_POLICIES` needs nothing.
- The prompt wrapper is handled the same way here as in `workflow_io` and `workflow_bindings`.
- The formats match the ComfyUI frontend (checked against 1.47.x). `graphToPrompt` saves through `LGraph.serialize()` (version 0.4), which always writes a `links` array, and every node carries `id` and `type`. Subgraphs (`definitions.subgraphs`) and groups are ignored, which is right.
- The accept side matches reality. I ran `check_comfy_workflow` over every local JSON file that `is_comfy_workflow` recognises, about 5k ComfyUI workflows (templates, custom-node examples, saved graphs). All were accepted except 3: two were not workflows at all, and one was a Civitai generation-metadata graph (see note).

**Verification**
- `tests/test_workflow_io.py`: 82 passed. Ruff check and format are clean.
- Mutation checks, each restored afterwards:
  - Removing the `check_comfy_workflow` call from `_store_workflow`: 12 failures.
  - Dropping the `links` requirement: the React Flow and n8n cases fail.
  - Dropping the `pixlstash_*` exemption: the accept test fails.

**Note (no change asked):** Civitai's copied generation data is an API graph with extra keys beside the nodes (`extra`, `extraMetadata`, `resource-stack*`). `is_comfy_workflow` accepts it for display; this check refuses it at import. That matches the PR's rule "refuse what cannot run", since ComfyUI's `/prompt` rejects an entry without `class_type`. Just flagging that the two checks now disagree on this shape on purpose.

pixlstash/utils/comfyui_utilities.py
lindkvis: Fixed in 337e1247, the way you suggested. A missing `links` is accepted when the document carries `state` (dict) or `last_node_id` (int), so a schema-v1 file whose root graph has no links goes through. A `links` key that is present still has to be a list. New cases: v1 and v0.4 files without `links` are accepted; `links: {}`, and `links: null` beside `state`, are refused. The React Flow and n8n cases stay red.

pixlstash/utils/comfyui_utilities.py
lindkvis: Fixed in 337e1247. A new `_shown()` quotes the value cut to 60 characters, and both the stray-key message and the node-id message use it. The test puts a 5000-character key in the graph and asserts the message stays under 200 characters.

pixlstash/utils/comfyui_utilities.py
lindkvis: Fixed in 337e1247. The two failures are separate now. A node with a type problem is named by its id ("node '7' has no type"). An entry with no id at all says "entry 3 of its nodes has no id", since there is no id to name. Tested with a graph whose second node is id 7.

tests/test_workflow_io.py
lindkvis: Added in 337e1247: `test_every_shipped_and_fixture_workflow_is_accepted` covers every file in `BUILT_IN` and `UI_FIXTURES`, 12 cases.

Issue #1401, opened 2026-09-17
Workflows & Recipes F0: WorkflowCard, ChipRow, InfoPopover


lindkvis:
Part of **v1.12 Workflows & Recipes**. The design is round 2, alternative C, "Stack Opens with Grid or List" (the design bundle `design/1.12-workflows-recipes/` in the business repo). Build what the design decides and use shipped components for the chrome. Proceed on the plan's recommended answers to the open owner decisions unless the owner has recorded otherwise in that bundle's `DECISIONS.md`.

Rules for every step: one PR against `develop`, about 600 lines of diff at most, shippable alone. Until F1b swaps the route, the shipped Workflows shelf keeps working. In code and docs, `workflow_recipe` / `structural_hash` is a **variant** and the user's Recipe is a **saved recipe**. The hub tables are append-only and are never renamed.

## Depends on

| Dependency | Hard? | Check it is in `origin/develop` |
|---|---|---|
| #1390 (B1 workflow identity): every step comes after it | yes | `git cat-file -e origin/develop:pixlstash/services/workflow_identity.py` |

## Before you start (mandatory)

1. **Start from the latest `origin/develop`:** `git fetch origin && git checkout -b <branch> origin/develop`. If you were handed a worktree that already exists, `git fetch origin` and rebase it onto `origin/develop` before touching anything. Do not build on a stale checkout.
2. **Check that every dependency below is actually in `origin/develop`,** not just closed or badged MERGED. Merges to develop do not auto-close issues. For each dependency, find its PR (`gh pr list --state merged --search "<issue or PR number>"`), confirm `git merge-base --is-ancestor <merge commit> origin/develop`, and run the check listed.
3. **If any hard dependency is missing, stop.** Comment on this issue saying which one, and do not start. Never base a PR on another PR's branch.

## What

- `widgets/WorkflowCard.vue`, the uniform card:
  - Cover: one large and two small pictures (2fr/1fr).
  - Four single-line rows in priority order: name, checkpoint, LoRAs (solid chip = workflow LoRA, dashed = recipe LoRA slot), special facts (differs by, type, imported).
  - Overflow becomes "+N". ⓘ sits bottom-right.
  - A stack card adds ▸ before the name and a layered count on the cover. A card with no pictures keeps the cover size and shows the type, one line and *Run it…*.
- `widgets/ChipRow.vue`: measures, clips and emits the "+N" count.
[...]

PR #1414, opened 2026-09-17
Workflows & Recipes F0: WorkflowCard, ChipRow, InfoPopover

lindkvis:
Part of #1401 (v1.12 Workflows & Recipes, step F0). This adds three widgets. Nothing mounts them yet; F1a does that.

- `WorkflowCard.vue`: the uniform card. The cover is one large and two small pictures (2fr/1fr). Four single-line rows follow: name, checkpoint, LoRAs (solid = workflow, dashed = recipe slot), and facts. It has ⓘ bottom-right, and ▸ plus a layered count on a stack. A card with no pictures shows the type, one line and *Run it…*. It is exactly `--wf-card-h` (252) tall.
- `ChipRow.vue`: measures the chips, clips to one line and emits the "+N" count.
- `InfoPopover.vue`: a `v-menu` holding `.tbm` and `.tbm-caret`, grouped into Models, Differs by, Defaults and the saved-recipe count.
- New shared token `--chip-h: 20px`. Everything else maps to existing tokens, as in the issue's table.
- "+N" is not a control, so the card's `aria-label` lists every chip, with "workflow LoRA" or "recipe LoRA slot", and gives ratings as "N of 5". ▸, ⓘ and *Run it…* are buttons at `tabindex=-1`, for the roving cursor F1a adds.

**Tests.** jsdom can't lay out, so the height check resolves the stylesheet against the token sheet and asserts the sum. It also asserts row 4's padding keeps it clear of ⓘ. The mounted tests cover structure, a11y and edge cases. I measured the 0-LoRA and 5-long-LoRA cases once in headless Chromium, at three widths: 252px every time, and row 4 always stopped before ⓘ.

**Size.** The diff is about 1,250 lines: roughly 800 of source (mostly scoped CSS) and 450 of tests. That is over the ~600 guide. I kept it as one PR because the issue puts all three pieces in F0 and none of them is useful alone.

**Pre-push review, answered here:**
- `:offset="8"` and `v-icon size="14"` follow the existing `.tbm` menu and icon usage. They are left as is.
- The caret stays on top if Vuetify flips the menu upward. Every existing `.tbm` menu does the same, so it is left for a shared fix.
- A single workflow with no type, not imported and no differences has an empty row 4. That case can't happen once B1 classifies every workflow's type.


lindkvis (commented):
Review (senior frontend and lead designer). The PR is solid work. The height arithmetic holds (1 + 132 + 8 + 4×24 + 3×2 + 8 + 1 = 252), `fitChipCount` is correct, and the pieces stay close to the shipped patterns (`AppButton`, `.tbm`, scrim badge, `--opacity-text-secondary`). Nothing blocks merging an unmounted F0. The two things to settle before F1a builds on it are the popover's accessibility and the card-data shape.

**Verification.** In a scratch worktree, the 30 vitest tests in the three new files pass. Mutation checks: ignoring the "+N" width in `fitChipCount` fails 2 tests, as it should. These mutations still pass: removing `aria-hidden` from the LoRA ChipRow, removing the reduced-motion block, removing `role="dialog"` from the popover, and making `.chip-row` `overflow: visible`. So the a11y and motion claims in the PR body aren't pinned by any test yet.

**Reuse check.** There's no existing shared popover (Tooltip and HelpTip are hover tips, not panels). There's no generic chip component either: `SettingsChip`, `DedupWhyPills` and `StackBadge` are each tied to their own domain. ChipRow and InfoPopover are justified, and the plan also names a second ChipRow consumer (the List view's Differs-by column). `WorkflowInspector`'s `.wfins-chip` is the same kind of "model chip", though, and it looks different (see the design-tokens comment).

**Dependency.** #1401 lists #1390 (B1) as a hard dependency and says to stop if it isn't in `develop`. It isn't merged. The plan's lane table says F0 only needs the contract, and this PR adds no API calls, so I don't see a merge risk. Still, please say in the issue that F0 went ahead on the plan's rule, so a later step doesn't copy that as a precedent. Checking the frontend's assumptions against #1390 turned up two gaps (inline on `utils/workflowCard.js` and `InfoPopover.vue`).

**Design system.** I checked the tokens against `docs/design/visual-language.md`, the shipped token sheet and the plan's token table. The design-sync tool wasn't available in this session, so I have **not** checked the chip against the Claude Design project's workflow-card spec. The lead designer should still confirm `--chip-h` and the chip type size there. `--chip-h` is on the 4px grid, and `docs/design/design-tokens.css` is a symlink to the frontend sheet, so the two sheets can't drift. Contrast: dashed chip and "+N" text use `--opacity-text-secondary`, which the token test already holds at 4:1 or better on the theme surfaces.

Severity: **Major** (fix before F1a mounts it), **Medium** (settle before B3/F1a), **Minor**, **Nit**.

frontend/src/components/widgets/InfoPopover.vue
```diff
@@ -0,0 +1,210 @@
...
+          ...menuProps,
+          'aria-haspopup': 'dialog',
+          'aria-expanded': String(open),
+        }"
+      />
+    </template>
+
+    <div
+      class="tbm info-popover"
+      role="dialog"
```
lindkvis, line 23: **Major (a11y).** `role="dialog"` on a panel that never gets focus. In Vuetify 3.6's `VMenu`, `onFocusIn` only moves focus to `focusableChildren(contentEl)[0]`, and this panel has no focusable children. Focus stays on ⓘ, and the content is teleported to the overlay container at the end of `<body>`. A screen-reader user who presses ⓘ hears "expanded" and then nothing. The dialog is never announced and is hard to reach with the virtual cursor. `VMenu` also adds `aria-owns` pointing at the menu to the activator, which makes this panel a child of a `<button>` in the accessibility tree, and a button's children are presentational.

Suggested fix: add `tabindex="-1"` to the panel and focus it when `open` becomes true (so Escape and click-outside still return focus through `VOverlay`), keeping `role="dialog"` + `aria-label`. Or drop `role="dialog"` and link the content to ⓘ with `aria-describedby`. Please add a test that fails without the fix. With the always-open `VMenu` stub in `WorkflowCard.test.js`, nothing here is exercised now: removing `role="dialog"` still passes.

Also a nit: `aria-expanded` on L15 repeats what `VMenu`'s activator props already provide. Only the `aria-haspopup` override is needed.

frontend/src/utils/workflowCard.js
```diff
@@ -0,0 +1,93 @@
+// Pure helpers behind WorkflowCard and ChipRow, kept out of the components so
+// the fitting arithmetic and the accessible name are testable without layout.
+//
+// The card's input is one workflow card from the Workflows & Recipes contract:
+//   { key, name, checkpoint, loras: [{ name, recipe }], type, imported,
```
lindkvis: **Medium (contract).** This shape isn't defined anywhere on the backend yet. #1390 is a pure derivation module with no serializer, and plan §6 doesn't freeze a card response. Two things will clash with B1/B3:

1. **Casing.** The app's API is snake_case (`node_id`, `forgotten_models` in `WorkflowInspector`). `differsBy`, `stackSize`, `pictureCount` and `savedRecipeCount` mean either a mapping layer or a rename in F1a. Pick one now and write it here.
2. **`loras[].recipe: boolean`.** B1's vocabulary is `mark: "structural" | "recipe"` (`STRUCTURAL`/`RECIPE`, frozen per slot label). Using the same field avoids a translation step that could invert the dashed/solid meaning.

The semantics match: structural LoRA = part of the card key = solid, recipe = slot = dashed. And a stack's row being the union of its members' `differs_by` (which B1 computes per member against the cover) agrees with plan §5.4.

frontend/src/components/widgets/InfoPopover.vue
```diff
@@ -0,0 +1,210 @@
...
+
+const subtitle = computed(() => {
+  const n = props.card.pictureCount ?? 0;
+  const pictures = n === 1 ? "found in 1 picture" : `found in ${n} pictures`;
+  return isStack(props.card)
+    ? `Stack of ${props.card.stackSize} workflows · ${pictures}`
+    : pictures.charAt(0).toUpperCase() + pictures.slice(1);
+});
+
+const models = computed(() => [
```
lindkvis, line 110: **Medium (contract).** Models lists only `checkpoint` + LoRAs, but #1390's `workflow_key` covers *every* non-LoRA slot (checkpoint/`unet_name`, VAE, CLIP loaders), and `differs_by` can emit "other models" when one of those differs. A stack chip that says "other models" then has no model in ⓘ to explain it. A single `checkpoint` string also can't hold a two-checkpoint graph (base + refiner). Suggest a card field `models: [{ name, kind, mark? }]` covering all slots. The card row can still show only the checkpoint.

frontend/src/components/widgets/WorkflowCard.vue
```diff
@@ -0,0 +1,347 @@
...
+
+    <span
+      v-if="stack"
+      class="wf-card__badge wf-card__badge--start"
+      aria-hidden="true"
+    >
+      <v-icon size="12">mdi-layers</v-icon>{{ card.stackSize }}
+    </span>
+
+    <!-- The visible rows are hidden from assistive tech: the card's own label
```
lindkvis, line 53: **Minor (a11y/tests).** Hiding the visible rows because the label reads them is the right call, but nothing tests it. Removing `aria-hidden="true"` from the LoRA `ChipRow` keeps all 30 tests green. One assertion that every `.chip-row` and `.wf-card__name` carries `aria-hidden` would pin it. Related: the empty cover's type (L29) and "No pictures yet" (L30) aren't hidden, so a pictureless card reads the type and picture count twice (once in the label, once in the content).

frontend/src/components/widgets/ChipRow.vue
```diff
@@ -0,0 +1,185 @@
...
+  align-items: center;
+  gap: var(--space-2);
+  height: var(--chip-h);
+  max-width: 100%;
+  padding: 0 var(--space-3);
+  border: 1px solid rgb(var(--v-theme-border));
+  border-radius: var(--radius-sm);
+  background: rgb(var(--v-theme-input-background));
+  color: rgb(var(--v-theme-on-surface));
+  font-size: var(--text-2xs);
```
lindkvis, line 128: **Minor (design).** `visual-language.md` reserves `--text-2xs` (11px) for "uppercase section labels, badge counts". Model names are mixed-case content. The sibling chips that show the same kind of thing use `--text-xs`: `.wfins-chip` in `WorkflowInspector` and `.tag-chip`. At `--chip-h` 20px with 1px borders, `--text-xs` at `--leading-snug` still fits. If the mock really draws 11px, that's a ramp decision for the lead designer to record, not a local choice.

frontend/src/styles/design-tokens.css
```diff
@@ -242,6 +242,11 @@
   --badge-size: 16px;      /* min height / min-width of a count pill (on-grid; was an ad-hoc 15px) */
   --badge-size-dot: 8px;   /* the no-count activity dot (on-grid; was an ad-hoc 7px) */
 
+  /* A model chip: the checkpoint and LoRA chips on a workflow card, in the
+     inspector and in the overlay's Recipe tab. One height so a row of them
+     sits inside a --control-h-sm row with room for its 1px border. */
+  --chip-h: 20px;
```
lindkvis: **Minor (design).** The comment says this token is shared by the inspector and the overlay's Recipe tab, but `.wfins-chip` in `WorkflowInspector.vue` is still `padding: 2px var(--space-3)` with no height, a `divider` border, a 0.04 wash and `--text-xs`. That's a different chip from ChipRow's (`border` colour, `input-background`, `--text-2xs`). Either move the inspector onto `--chip-h` and one chip treatment in the step that touches it, or word the comment as intent ("will be shared by") so it doesn't describe something that isn't true yet. Please also confirm the token against the design system project's card spec.

frontend/src/components/widgets/ChipRow.vue
```diff
@@ -0,0 +1,185 @@
...
+watch(
+  () => props.items,
+  () => nextTick(measure),
+  { deep: true },
+);
+
+let observer = null;
+onMounted(() => {
+  measure();
+  observer = new ResizeObserver(measure);
```
lindkvis, line 92: **Minor (perf, for F1a).** Every ChipRow creates its own `ResizeObserver` and a hidden copy of every chip. At three rows per card, a Workflows grid of N cards has 3N observers and doubles the chip DOM. That's fine for dozens of cards. If F1a expects hundreds, one shared observer (a WeakMap from element to measure) is a small change. Also, a ChipRow whose width depends on its content (the List column) could hit a ResizeObserver loop warning, because `measure` changes what's shown. Here that can't happen because the row is `flex: 1`.

frontend/src/components/widgets/ChipRow.vue
```diff
@@ -0,0 +1,185 @@
...
+  const gap = parseFloat(getComputedStyle(row).columnGap) || 0;
+  visibleCount.value = fitChipCount(
+    chips.map((c) => c.getBoundingClientRect().width),
+    row.clientWidth,
+    gap,
+    more.getBoundingClientRect().width,
+  );
+}
+
+watch(hiddenCount, (n) => emit("overflow", n));
```
lindkvis, line 79: **Nit (YAGNI).** Nothing listens to `overflow` yet: `WorkflowCard` hides "+N" from AT and doesn't use the count. It's in the issue, so keep it if the List column will use it. Otherwise add it when there's a listener. `defineExpose({ measure })` only exists for tests.

frontend/src/components/widgets/InfoPopover.vue
```diff
@@ -0,0 +1,210 @@
...
+
+<style scoped>
+.info-popover {
+  width: var(--stats-panel-w);
+  padding: var(--space-4);
+}
+
+/* Centred on the 24px ⓘ it opens from, which sits at the panel's end edge. */
+.info-popover__caret {
+  right: calc(var(--control-h-sm) / 2 - 5.5px);
```
lindkvis: **Nit (tokens).** This repeats `.tbm-caret--icon-center-end` (`App.css`, `calc(16px - 5.5px)`) for a 24px activator. A shared `.tbm-caret--icon-sm-end` modifier next to it would keep the 5.5px caret half-width in one file instead of two.

frontend/src/components/widgets/InfoPopover.vue
```diff
@@ -0,0 +1,210 @@
...
+    </template>
+
+    <div
+      class="tbm info-popover"
+      role="dialog"
+      :aria-label="`About ${card.name}`"
+      data-testid="workflow-info-popover"
+    >
+      <span class="tbm-caret info-popover__caret" aria-hidden="true"></span>
+      <h4 class="info-popover__title">{{ card.name }}</h4>
```
lindkvis: **Nit.** `h4`/`h5` inside a teleported overlay add headings with no level context to the page outline. The shipped `.tbm` panels use `.tbm-label` on non-heading elements. With `role="dialog"` + `aria-label` the panel is already named.

lindkvis (commented):
Follow-up: design system check (lead designer), against the DS Tag, Menu and spacing specs.

- **No workflow-card spec exists in the DS.** `ui_kits/app` only has folder-browser and unified-shell, so the card follows the plan's token table.
- **The dashed "recipe LoRA" chip isn't in the DS.** It's a new design decision and needs the owner's sign-off.
- **Matches the DS:** the card's `--radius-md`, and InfoPopover as a `.tbm` panel (`--radius-lg`, panel fill, 1px border, `--elevation-3`). Menu.prompt.md defines a surface that holds sections as a panel, not a menu.
- **Correction to my earlier review:** `--text-2xs` isn't reserved for uppercase labels. The DS Tag xs size uses it, so drop that Minor. The real mismatch is the chip itself (inline).
- **Size call for a 252px card:** use **Tag xs** (18px, `--space-2` padding, `--text-2xs`). Meta rows are 24px with a 2px gap. A Tag sm (24px) would fill each row edge to edge and leave only 2px between the chips in rows 2 and 3. xs leaves 3px either side, keeps 252 unchanged, and is the size the DS uses for dense overlay and filmstrip metadata.

frontend/src/styles/design-tokens.css
lindkvis: **Medium (design).** The DS Tag has no 20px size: xs is 18px, sm is `--control-h-sm` (24px), md is `--control-h` (28px). A new shared `--chip-h: 20px` would create a fourth chip height. Drop the token and build to Tag xs (18px). See the review body for why xs suits the 252px card.

frontend/src/components/widgets/ChipRow.vue
```diff
@@ -0,0 +1,185 @@
...
+
+.chip-row__chip {
+  display: inline-flex;
+  flex-shrink: 0;
+  align-items: center;
+  gap: var(--space-2);
+  height: var(--chip-h);
+  max-width: 100%;
+  padding: 0 var(--space-3);
+  border: 1px solid rgb(var(--v-theme-border));
```
lindkvis: **Medium (design).** This chip matches no DS Tag size or fill. What changes, per the DS Tag xs size and resting state:
- **Height:** 20px → 18px.
- **Padding:** `--space-3` → `--space-2`.
- **Border:** the 1px border goes. A resting DS Tag has none.
- **Fill:** `input-background` → a 10% ink wash, `rgba(var(--v-theme-on-surface), 0.1)`.
- **Label:** `on-surface` → muted text, `rgba(var(--v-theme-on-surface), var(--opacity-text-secondary))`, at regular weight.

Knock-on effects:
- `--fact` (transparent) has no DS equivalent. With no border, a transparent chip is just bare text, so either render facts as plain Tags or get the variant approved.
- Dashed recipe chip (L153): once the resting chip has no border, the dashed outline is the only border in the row. That's part of the owner's sign-off.
- The 12px icon is off the `--gutter-glyph` (16px) box, and a 16px glyph won't fit an 18px chip with padding. Decide whether chips carry icons at all.

frontend/src/components/widgets/InfoPopover.vue
```diff
@@ -0,0 +1,210 @@
...
+.info-popover__chips {
+  display: flex;
+  flex-wrap: wrap;
+  gap: var(--space-2);
+}
+
+.info-popover__chip {
+  display: inline-flex;
+  align-items: center;
+  height: var(--chip-h);
```
lindkvis: **Minor (design).** The Differs-by chips in the popover have the same mismatch as ChipRow (20px, border, `--space-3` padding). Inside a panel there's room for a Tag sm (24px, `--space-3`, `--text-xs`, 10% wash, no border), or use xs to match the card. Whichever you choose, use a DS size.

frontend/src/components/widgets/WorkflowCard.vue
```diff
@@ -0,0 +1,347 @@
...
+  background: rgb(var(--v-theme-input-background));
+}
+
+.wf-card__type {
+  display: inline-flex;
+  align-items: center;
+  height: var(--chip-h);
+  padding: 0 var(--space-3);
+  border: 1px solid rgb(var(--v-theme-border));
+  border-radius: var(--radius-pill);
```
lindkvis: **Minor (design).** The workflow type (`outpaint`, `txt2img`) is a category, not a status, so per the radius tiers it's a chip (`--radius-sm`), not a pill (`--radius-pill`). Build it as the same Tag xs as the card rows: no border, 10% wash, muted label.

frontend/src/components/widgets/InfoPopover.vue
lindkvis: Fixed in bc3b5f58. The panel now carries \`tabindex="-1"\` and focuses itself when \`open\` turns true, keeping \`role="dialog"\` + \`aria-label\`; `VOverlay` still returns focus to ⓘ on Escape and on a click outside. The `aria-expanded` override is gone — only the `aria-haspopup: dialog` override is passed to the activator now.

New file `InfoPopover.a11y.test.js` mounts the **real** `VMenu` (the stub elsewhere is always open and focuses nothing), clicks the activator, and asserts `role`, `aria-label`, `document.activeElement === panel`, and that Vuetify supplies `aria-expanded` false→true. Both mutations you named now go red: dropping the focus call fails 1 test, removing `role="dialog"` fails 1.

Not addressed: you are right that `VMenu` also sets `aria-owns` on the activator, so the panel is a child of a `<button>` in the AX tree. That is true of every `.tbm` menu in the app, so it wants one shared fix rather than a local one here.

frontend/src/utils/workflowCard.js
lindkvis: Fixed in bc3b5f58, both points, and the shape is now written down at the top of `utils/workflowCard.js`.

1. **snake_case**, as with every other response this app reads: `picture_count`, `stack_size`, `differs_by`, `saved_recipe_count`, `covers`, `defaults`.
2. **`mark: "structural" | "recipe"`** per slot, B1's own vocabulary, so nothing translates and nothing can invert the dashed/solid meaning. `loraChips` keys off `mark === "recipe"`; a mutation that inverts it fails 2 tests.

Fixtures and tests moved with it, and `docs/frontend_architecture.md` states the contract and why it is snake_case.

frontend/src/components/widgets/InfoPopover.vue
lindkvis: Fixed in bc3b5f58. The card now takes `models: [{ name, kind, mark? }]` covering every non-LoRA slot, and `checkpointModel()` picks the `checkpoint` slot, falling back to the first model for a unet graph. ⓘ lists **all** of them with `kind` as the note, so a stack reading "other models" has the models behind it; the card row still names the checkpoint only, as drawn. Two new tests: ⓘ lists a unet + vae pair, and the checkpoint is found when it is not first in the array (that one fails if the lookup degrades to `models[0]`).

frontend/src/components/widgets/WorkflowCard.vue
lindkvis: Fixed in bc3b5f58. A new test collects every `.chip-row`, `.wf-card__name` and `.wf-card__none` in a crowded card and asserts each carries `aria-hidden="true"`; removing it from the LoRA ChipRow now fails. The empty cover's type and "No pictures yet" are hidden too, with their own test, so a pictureless card no longer reads its type and count twice.

frontend/src/components/widgets/ChipRow.vue
lindkvis: Settled by your own follow-up: `--text-2xs` is the DS Tag xs size, and the chip is now Tag xs, so the size stays. The mismatch you pointed at is fixed in bc3b5f58 (see the ChipRow thread).

frontend/src/styles/design-tokens.css
lindkvis: Fixed in bc3b5f58 by dropping `--chip-h` for `--tag-h-xs: 18px`, named and commented as the DS Tag xs height, and the comment no longer claims consumers that do not exist. `.wfins-chip` is untouched: moving the inspector onto one chip treatment belongs to F3, the step that reshapes it. Confirmed against the DS: `components/core/Tag.jsx` gives xs 18px / sm `--control-h-sm` / md `--control-h`, so this token is one of the DS' three heights rather than a fourth. There is no workflow-card spec in the DS to check the card against, as your follow-up found.

frontend/src/components/widgets/ChipRow.vue
lindkvis: Acknowledged, no change. One `ResizeObserver` and one hidden chip copy per row is what makes a row measurable without layout thrash, and F1a expects tens to low hundreds of cards (§4.2 of the plan defers virtualisation above ~1000). I left a `ponytail:` comment in `ChipRow.vue` naming the ceiling and the upgrade (one shared observer keyed by element), so F1a finds it if the number grows. Your note about a content-width ChipRow in the List column is on the record for F2; there the row is `flex: 1`, so it cannot loop.

frontend/src/components/widgets/ChipRow.vue
lindkvis: Keeping `overflow`: the plan names F2's List "Differs by" column as the second consumer, and the issue asks ChipRow to emit the count. It is two lines and it stays. `defineExpose({ measure })` is indeed only for tests — it is how the suite can lay the row out in jsdom at all, which is the only way the "+N" arithmetic is checked anywhere.

frontend/src/components/widgets/InfoPopover.vue
lindkvis: Fixed in bc3b5f58. `.tbm-caret--icon-sm-end` now lives in `App.css` beside its 32px sibling, with the 5.5px caret half-width in one file, and `InfoPopover` uses it instead of a local rule.

frontend/src/components/widgets/InfoPopover.vue
lindkvis: Fixed in bc3b5f58. The title is a `<p>` and the three group labels are `<span class="tbm-label">`, so nothing adds a heading level to the page outline. The panel is still named by `role="dialog"` + `aria-label`.

frontend/src/styles/design-tokens.css
lindkvis: Fixed in bc3b5f58: `--chip-h` is gone. `--tag-h-xs: 18px` replaces it, commented as the DS Tag xs height with the other two named, so the app has three chip heights, not four. I kept it as a token rather than an 18px literal because three components need it (card rows, the type chip, and the overlay Recipe tab next); if you would rather see the literal, say so and it is a one-line change.

frontend/src/components/widgets/ChipRow.vue
lindkvis: Fixed in bc3b5f58, all five: 18px, `--space-2` padding, no resting border, a 10% ink wash (`color-mix(in srgb, rgb(var(--v-theme-on-surface)) 10%, transparent)`), and a muted label at regular weight.

Knock-ons, as you framed them:
- **`--fact` is gone.** Facts render as ordinary Tags, since a transparent chip without a border is just text.
- **Icons are gone from chips.** The DS Tag has none, a 16px glyph does not fit an 18px chip, and 12px was off the `--gutter-glyph` box. The mock draws them; reverting is one line if you want them back.
- **The dashed recipe chip stays** (1px dashed at 0.35 on-surface, transparent), and is now the only border in the row. Flagged in the PR body and in `docs/frontend_architecture.md` as a new design decision awaiting your sign-off.

Measured in Chromium after the change: chips 18px, rows 24px, card still exactly 252px at three widths.

frontend/src/components/widgets/InfoPopover.vue
lindkvis: Fixed in bc3b5f58. The popover chips are now Tag sm: `--control-h-sm`, `--space-3`, `--text-xs`, a 10% wash on `on-panel`, no border. Tag sm rather than xs because there is room inside the panel and the chips wrap there rather than competing with a 24px row.

frontend/src/components/widgets/WorkflowCard.vue
lindkvis: Fixed in bc3b5f58. The type chip is now the same Tag xs as the rows: `--radius-sm`, no border, 10% wash, muted label.

lindkvis:
Review answered in bc3b5f58 — every thread has a reply, and the twelve I acted on are resolved. Two are left open on purpose: the per-row `ResizeObserver` (kept, with a `ponytail:` comment naming the ceiling and the upgrade) and the `overflow` emit (kept, because F2's List column is its consumer). Both are your call rather than mine.

What changed:

- **Chips are the DS Tag xs** (18px, `--space-2`, `--text-2xs`, 10% ink wash, muted label, no border, no icon), per `components/core/Tag.jsx`. `--chip-h: 20px` is gone; `--tag-h-xs: 18px` names the DS' own xs height, so there is no fourth chip height. Popover chips are Tag sm, the type chip is Tag xs at `--radius-sm`.
- **The card contract is snake_case with `mark: "structural" | "recipe"`** and a `models: [{ name, kind, mark? }]` list covering every non-LoRA slot, so B3 needs no mapping layer and ⓘ can always explain an "other models" difference.
- **The popover takes focus on open** (`tabindex="-1"` + focus call), and `InfoPopover.a11y.test.js` mounts the real `VMenu` to hold that. Headings became non-heading elements; the caret moved to a shared `.tbm-caret--icon-sm-end` in `App.css`; the duplicate `aria-expanded` is gone.
- **The four surviving mutations you found are now red**: the focus call, `role="dialog"`, `aria-hidden` on the rows, and `.chip-row`'s clip. Same for the reduced-motion block, the `mark` inversion and the checkpoint lookup.

Still yours to decide, and both now stated in `docs/frontend_architecture.md`:

1. **The dashed "recipe LoRA" chip** is not in the DS and needs your sign-off. With the resting chip unbordered it is the only border in a row, which is what makes a slot read as a slot.
2. **Icons left the chips** (the DS Tag has none, 16px will not fit an 18px chip, and 12px was off the glyph box). The mock draws them; putting them back is one line.

Re-measured in Chromium after the change, at three card widths: card 252px, rows 24px, chips 18px, row 4 always clear of ⓘ. Full frontend suite: 4,333 tests pass.

PR #1417, opened 2026-09-17
Add a LoRA loader to a workflow that has none

### LINKED ISSUE #1376: Insert a LoRA loader into a workflow that has none

lindkvis:
Closes #1376.

A shelf LoRA can now go into a workflow with no LoRA loader: PixlStash adds one for that run.

- **Where:** straight after the model source. Every input reading its MODEL (and CLIP) output is rewired to the loader, so model patches, several samplers and a separate CLIP loader all keep working. Output types come from ComfyUI's `object_info`, not input names, so no reader gets missed.
- **Shown first:** new owner-only `GET /comfyui/workflows/{name}/lora-insertion`, plus `lora_insertion` on the recipe read. The run panel, Generate variants and Edit with ComfyUI show "A LoRA loader is added after #4 …, feeding …" before the run.
- **Opt-in:** the run adds a loader only with `insert_lora_loader: true`. A bare `adapter_sha256` is still a 400.
- **Which loader:** `LoraLoader` (or `LoraLoaderModelOnly` when nothing reads a CLIP) when that ComfyUI lists the file. Otherwise `PixlStashAdapterLoader` by digest, but not in `run_recipe`: its variant would carry a PixlStash node, and Generate variants refuses to replay those.
- **Refused, with the reason:** no model source; several models or text encoders; a linked node this ComfyUI lacks; a graph that already loads a LoRA with no swappable slot (a wired `lora_name`, rgthree's stacker), since adding one would silently stack two.
- The stored workflow file is never changed.

Not done, from review:
- Stacking several adapters is left out.
- A model loader that core `LoraLoader` cannot patch (Nunchaku, TensorRT) still gets a loader.
- A `*`-typed passthrough between the model loader and its readers reads as a second model and is refused.
- One unrelated missing node blocks insertion.
- The run recomputes the plan instead of checking it against the one previewed.
- Each loader-less workflow picked asks ComfyUI for `object_info`, uncached.

Documented in the three architecture docs. Authz matrix: +1 `owner_only`.


lindkvis (commented):
## Review of `8def930`: four blockers, plus an authz sign-off

Reviewed against `origin/develop` in four independently-briefed lanes — graph semantics, backend API/tests, adversarial authz, frontend. Every finding below was re-verified against the PR head before posting. Two pairs of lanes converged on the same defect without seeing each other's work (G1 and G4); that convergence is the strongest signal here.

**The caveat that qualifies everything:** no backend test was executed. The review environment has no `pytest` and no `fastapi`. Only the frontend suite really ran (115 files / 1778 tests, plus 16 applied mutations). The backend mutation evidence came from transcribing assertions into a standalone harness, and four authz guardrail tests were simulated by re-implementing their logic over the head sources. Before merge, the five touched test files should be run for real.

---

## Merge gate

### G1 — HIGH. A LoRA applied by any unrecognised means gets a second one stacked, silently
`pixlstash/services/comfyui_recipe_service.py:801-820`

`_loads_a_lora_unseen` recognises a pre-existing LoRA by exactly two field spellings — `^lora_name(_\d+)?$` (wired) and `^lora_\d+$` (dict). Verified non-matches:

| field | filename regex | stacker regex |
|---|---|---|
| `lora_name` | match | — |
| `lora_name_2` | match | — |
| `lora_1` | — | match (dict) |
| `lora_1_name` | **no match** | **no match** |
| `text` | **no match** | **no match** |

So a prompt-tag loader (`text: "<lora:styleA:0.8>"`, the LoraTagLoader / ImpactWildcardEncode shape) or a stacker spelling its slots `lora_1_name` (the easy-loraStack shape) falls through the guard and gets a loader spliced in ahead of it — **two LoRAs live, with nothing said**. This is the outcome the description says is refused. Reproduced by running the service against hand-built graphs.

A class-level rule closes it for free: by the time `plan_lora_insertion` runs, a swappable slot has already been ruled out, so *any* remaining LoRA-ish node (its `class_type` mentions lora, or it declares a LORA-ish type in the `object_info` already in hand) is by definition an unswappable LoRA. Refusing on that, rather than on two widget spellings, also folds in the `lora_name: null` case.

The new tests parametrise exactly the two shapes the code implements (`tests/test_comfyui_recipe_preflight.py:884-902`), so the "will not add a second one" claim is currently asserted only where it is true by construction.

### G2 — MEDIUM. A second model chain the type walk cannot see bypasses the >1-model refusal
`pixlstash/services/comfyui_recipe_service.py:746` is `if kind in ("MODEL", "CLIP")`.

A source whose declared output is `WANVIDEOMODEL` — or whose `object_info["output"]` is shorter than its real outputs, yielding `None` — never enters `links`, so the ">1 model" refusal at `:768` cannot see it. Reproduced: a `WanVideoModelLoader` → `WanVideoSampler` chain beside a `CheckpointLoaderSimple` → `KSampler` inserts on the checkpoint, and the video chain runs with no LoRA, silently. A graph where *no* chain types as MODEL is correctly refused, so the defect is specifically the mixed graph — the very risk the refusal exists to prevent, reached by a route it cannot see.

### G3 — MEDIUM. `run_recipe`'s replayability guard is asserted nowhere
`pixlstash/routes/comfyui.py:2892` passes `digest_loader=False` so an inserted `PixlStashAdapterLoader` cannot make the output unreplayable by Generate variants. But `tests/test_comfyui_recipe_route.py:528` injects a `LoraLoaderModelOnly` whose combo list contains the shelf file, so `_inserted_loader` returns on its **first** branch (`comfyui_recipe_service.py:871`) — and `if not digest_loader: raise` sits after it, at `:872`. The test asserts `class_type == "LoraLoaderModelOnly"`, which proves it took branch one, and `PixlStashAdapterLoader` appears nowhere in that file's `OBJECT_INFO`.

**Deleting `digest_loader=False` from the route leaves all three test files green.** The unit test proves `insert_adapter` honours the flag; nothing proves the route passes it.

Suggested fix: one recipe case with `PixlStashAdapterLoader` in `object_info` and a shelf LoRA the core loader does not list, asserting 400 and `submitted == []`.

### G4 — LOW/MEDIUM. The new authz negative cannot fail for the right reason
`tests/test_workflows_api.py:2230` — found independently by the backend lane and the authz lane.

`test_the_insertion_preview_is_owner_only` asserts 403 for a character-scoped token, but its liveness control GETs `/comfyui/workflows` — a **different** path from the `…/plain.json/lora-insertion` under test. `pixlstash/auth.py:2718-2725` refuses any GET under `READ_BLOCKED_GET_PREFIXES` for any scoped token *before routing*, and `/api/v1/comfyui/workflows/` is such a prefix.

Concrete mutation: flip the registry entry from `_OWNER` to `_ANY` and this test — named after that very declaration — **stays green**. It would also pass against `…/lora_insertion` or a route that was never mounted.

This file already names the hazard and fixes it ten times over. `tests/test_workflows_api.py:418-444` is the identical probe, and its own docstring says:

> `assert_real_route` is load-bearing: the middleware answers before routing, so a renamed route would 403 identically and the assertion would dissolve into a test of nothing.

`:887-888` (also `:1111`, `:1420`) additionally empties both belts "so the refusal is measured at the gate rather than at the middleware in front of it". The new test does neither — a two-line fix with patterns already in the file.

Separately, the new route is absent from `_WORKFLOW_ROUTES` (`:64`), so the test asserting every entry declares `OWNER_ONLY` does not cover it. Completeness only: a loosened declaration is still caught behaviourally, and `test_read_blocked_get_prefixes_cover_only_owner_class_gets` keeps the class machine-covered.

---

## Frontend

### F1 — HIGH. The feature is untested on 2 of the 3 surfaces (mutation-proved)
Deleting the feature outright from `frontend/src/components/io/RemixDialog.vue:791` or `frontend/src/components/views/ImageOverlay.vue:1727` leaves the **entire** frontend suite green — 115 files / 1778 tests. Deleting the disclosure markup from either is green. So is deleting `...comfyLoraBody()` from `ImageOverlay.vue:1895`, which also means the overlay's shipped #1310 swap has never had a test.

`RemixDialog.test.js` gained exactly one line here: a `getLoraInsertion` mock hard-wired to `{plan: null, reason: null}`, which keeps the old suite compiling and asserts nothing new.

Failure it allows: a refactor drops the second argument in Remix, the user is told nothing can be added, the feature is silently gone and CI is green. Or the user picks a LoRA in the overlay, reads "A LoRA loader is added after #4 …", clicks Run, and gets an image with no LoRA and no error.

### F2 — MEDIUM. The fixed sentence contradicts the reason printed beneath it
`frontend/src/components/panels/WorkflowRunPanel.vue:109`, `frontend/src/components/io/RemixDialog.vue:242-243`, `frontend/src/components/views/ImageOverlay.vue:242` each hardcode "This workflow has no LoRA loader." and then append the backend `reason` verbatim. Two named reasons make the pair self-contradictory:

- rgthree: *"This workflow has no LoRA loader. Node 7 (Power Lora Loader (rgthree)) already loads a LoRA in a way PixlStash cannot swap…"*
- UI-format: *"This template has no LoRA loader. This workflow is saved in ComfyUI's UI format, so PixlStash cannot add a LoRA loader to it."*

The backend deliberately declines to make that first claim — `pixlstash/routes/comfyui.py:535-536`:

```python
# Said as what it is: "no LoRA loader" would be false about a UI-format
# file that may well have one.
```

The frontend asserts it anyway. One string, three places — a class fix. Reachable on 2 of 3 for the UI-format case: Remix and the overlay filter on `valid`, which `_describe_workflow` sets `true` for UI-format; only the panel filters on `runnable !== false`.

### F3 — MEDIUM. The disclosure omits which loader is spliced in, and that choice can make the output unreplayable
`frontend/src/composables/useLoraSwap.js:40-56` states the source node, the rewired inputs and the model-only caveat — but not the loader class. The backend picks between `LoraLoader`, `LoraLoaderModelOnly` and `PixlStashAdapterLoader`, and the last makes the picture unreplayable by `run_recipe`, which refuses any graph carrying a `PixlStash*` node. The panel and overlay, unlike `run_recipe`, do not pass `digest_loader=False`, so they are the two surfaces that can produce it. The plan payload carries no loader class, so this wants an API field rather than prose in three components.

### F5 — LOW. A TDZ landmine held off by one short-circuit
`frontend/src/components/panels/WorkflowRunPanel.vue:263` reads `chosenName`, declared at `:309`, and the `useLoraSwap` watcher is `immediate: true`, so the computed runs during setup. Only `inputs.value === null` being tested first prevents `ReferenceError: Cannot access 'chosenName' before initialization`. Reordering those two lines turns 20/20 panel tests red with exactly that error. The code comments the hazard at `:258-260` rather than removing it; moving the declaration above the computed costs nothing.

---

## Authz: sign off

Independent adversarial lane, on arithmetic re-derived separately:

```
registry keys ADDED   = [GET /api/v1/comfyui/workflows/{workflow_name}/lora-insertion]
registry keys REMOVED = []
comfyui.py HTTP route decorators    = 15
registry keys under /api/v1/comfyui = 15
```

15/15 module completeness, exactly one key added, and **no existing declaration changed** across 257 changed lines — so the stale-key failure mode (an orphaned key, or a renamed route inheriting a policy written for another endpoint) does not land. Coverage matrix 331/331, no drift, no orphans; `owner_only` 151 → 152. The four derived belt guardrails hold, so the route stays closed under the documented `AUTHZ_GATE_ENFORCING = False` rollback with no new belt line. No ad-hoc per-handler scope check was added. The #708 existence-oracle precedent is not repeated: workflow names are not secret (`GET /comfyui/workflows` is `any_token`) and `workflow_name` passes through `basename` + forced `.json` + `resolve_path_within`.

Cleared after being candidate findings: the new `lora_insertion` field on the picture-scoped recipe route adds no disclosure class (every plan element and error string derives from the picture's own embedded graph, already served on that response, and `_describe_lora_insertion` never calls `_inserted_loader`, so no shelf filename or digest can reach it); `_apply_lora_swap`'s rewrite fails closed; the frontend does not call the owner-only route under a share link; the on-disk workflow file is never written; every reader of the model source *is* rewired, `*`-typed passthroughs included; node-id allocation cannot collide; `insert_adapter` validates the whole plan before mutating, so a divergence refuses whole rather than half-writing.

Also clean: the sibling sweep (3 `useLoraSwap` callers, 3 touched, no fourth surface — `runTextToImage` has no component caller); design tokens (the PR adds no CSS and no inline styles, so zero hardcoded hex, off-grid spacing, ad-hoc radius, raw rgba shadow or off-ramp font-size); CI gating (all three touched test files already in the blocking `backend` job list); `ruff check` and `ruff format --check` on the touched Python.

---

## Follow-ups, not blockers

- No acyclicity check (`comfyui_recipe_service.py:942-953`): a CLIP root that also reads MODEL produces loader→node→loader. Defect confirmed; the triggering node signature is plausible but unnamed.
- The inserted loader's own input names `model`/`clip` are hardcoded and never checked against `object_info` (`:942-944`) while its *outputs* are validated (`:918-922`). A `LoraLoader` declaring `base_model`/`text_encoder` yields an opaque 400 from `POST /prompt` after the run is queued — the failure this pre-flight exists to prevent. `find_input_spec` is already available.
- Insertion is refused by any missing node pack anywhere in the graph, however unrelated (`:734-739`). Fails closed, so usability only.
- `has_lora_loader: false` for a UI-format workflow that may well have one (`pixlstash/routes/comfyui.py:1910-1922`) — the same false claim `_resolve_lora_swap` avoids. Should be `None`/unknown on that branch.
- Two silently-ignored requests: `insert_lora_loader: true` with no `adapter_sha256` returns a LoRA-less 200 with no complaint; the same flag on a graph that *does* have a slot silently overwrites that slot.
- Unlogged `LookupError` catches (`pixlstash/routes/comfyui.py:689`, `comfyui_recipe_service.py:866`): the reason reaches the user, never the log.
- The preview validates the plan but not the adapter, so a confident disclosure can still precede a 400.
- `frontend/src/api/comfyui.js` JSDoc omits `lora_insertion` and `insert_lora_loader`.
- `rewires.sort` is lexicographic, so "#10" precedes "#3" in the owner-facing sentence.

## UX and design, raised rather than adjudicated

- The graph-mutation sentence renders at `--text-xs` / 60-70% opacity — the same class as incidental hints — on all three surfaces, while `RemixDialog` already ships `.remix-disclosure` for exactly this job and the new note sits beside it at caption weight.
- Shown, never acknowledged: `allow_unchecked` in the same dialog demands a tick, whereas "we will add a node and rewire N inputs" gets a caption. Remix opens focused on Generate, so LoRA + Enter is a two-keystroke path to a rewired graph.
- The sentence does not scale: all rewires joined into one sentence, CLIP before MODEL, so a Flux graph leads with text encoders in a long comma list at 12px in a dropdown-width panel.
- The loading branch — "This workflow has no LoRA loader. Checking whether PixlStash can add one…" — is two sentences where the second retracts the first, and is untested.

## Not verified by anyone

Behaviour against a real ComfyUI `object_info`: every graph probe used hand-written specs, and there are **no API-format workflow fixtures in the repo** — all `tests/comfyui_workflows/*.json` are UI format — so neither this review nor the PR's own tests exercise the graph logic against a real document. Also unverified: any backend test execution; runtime gate behaviour on the new route (argued from the code path, not observed); whether a real node pack has the MODEL-in/CLIP-out signature the acyclicity finding needs; how the disclosure reads at 12px in both themes; an end-to-end run against a live ComfyUI.

---

lindkvis:
Addressed in `a690928b`. Every merge-gate item is fixed; the backend tests the review could not run were run here, and each new assertion was checked by breaking what it guards.

**G1 — stacking. Fixed, at the class level.** `_already_loads_a_lora` replaces the two widget spellings: a node counts as an unswappable loader when its `class_type` mentions a LoRA, when `object_info` declares a LoRA-ish type on any of its inputs or outputs, or when a value carries a `<lora:…>` tag. A swappable slot has already been ruled out by the time planning runs, so any such node is one that cannot be swapped. Parametrised over five shapes now, including `lora_1_name`, the prompt tag and `lora_name: null`, plus a case where only the spec says LoRA.

**G2 — the invisible second model chain. Fixed.** Links typed `*MODEL*` but not `MODEL` (`WANVIDEOMODEL`) are now collected and refuse the graph whole. A source whose spec has no `output` list, or one shorter than the link, is refused too rather than read as "not a model" — that was the other way the chain stayed invisible. Test: a WanVideo chain beside a checkpoint, with the control that removing it splices again.

**G3 — the replay guard was untested. Fixed.** `test_a_replay_never_adds_the_pixlstash_loader`: `PixlStashAdapterLoader` present, the core loader not listing the shelf file — 400, nothing submitted, and the control that listing the file makes the same replay run. Deleting `digest_loader=False` from the route now fails it.

**G4 — the authz negative. Fixed both ways.** The GET belts are emptied, `assert_real_route` runs on the path under test, the liveness control is `/pictures`, and the owner's 200 on the same path is the positive control. Flipping the registry entry to `_ANY` now reds this test and `test_every_workflow_route_is_declared_owner_only`; the route is in `_WORKFLOW_ROUTES`.

**F1 — two untested surfaces. Fixed.** New `ImageOverlayLoraSwap.test.js` mounts the real overlay with the ComfyUI menu rendered and covers the shipped #1310 swap as well as insertion; Remix gets three tests (recipe-read plan, template fetch, reason shown). Checked by mutation: dropping `...comfyLoraBody()`, or the insertion source from either `useLoraSwap` call, goes red.

**F2 — the contradictory sentence. Fixed, one string.** `noLoaderNote` in the composable: the backend's reason replaces the sentence rather than following it, and all three surfaces call it.

**F3 — which loader. Fixed with an API field.** The plan carries `pixlstash_loader`, whether the digest loader could be the one inserted; the summary then warns that those pictures cannot be re-run by "Generate variants". The recipe read reports it `false`, since that route never inserts it.

**F5 — TDZ. Fixed.** `chosenName` moved above the computed; the short-circuit is no longer load-bearing.

Also fixed from the follow-ups: the inserted loader's `model`/`clip` **inputs** are now validated against `object_info`, not just its outputs; a CLIP source that reads the model is refused rather than wired into a cycle; `insert_lora_loader` without `adapter_sha256` is a 400 instead of a silent LoRA-less run; `has_lora_loader` is `null` for a UI-format file; `rewires` sorts numerically, so #3 precedes #10; both `LookupError` catches log; the three architecture docs and the `comfyui.js` JSDoc now match.

Left, deliberately, and recorded in `backend_architecture.md`:

- A model loader whose MODEL core `LoraLoader` cannot patch (Nunchaku, TensorRT) still gets a loader. Refusing it needs a list of loader-compatible model types, which is the class allowlist `detect_lora_targets` exists to avoid.
- Insertion is refused by any missing node pack anywhere in the graph. Fails closed.
- The run recomputes the plan rather than checking the one previewed. Named in the doc as not being proof the owner saw that splice.
- The preview validates the plan, not the adapter, so a 400 can still follow a confident disclosure.
- The UX points (disclosure weight, an acknowledgement rather than a caption, the sentence at Flux length) are design decisions rather than defects, and want the lead designer and UI/UX rather than this PR.

Verified here: 378 backend tests across the touched files plus the architecture guardrails and shard classification, and the whole frontend suite, 4356 tests in 252 files. Ruff and the generated route table are clean.

PR #1420, opened 2026-09-17
Vault: the database filename is a parameter

lindkvis:
The first import needs a vault under a temporary name; every other caller
keeps vault.db.


tests/process_teardown.py
```diff
@@ -0,0 +1,154 @@
...
+        # No session ran (``--collect-only``, a usage error). Nothing loaded
+        # the native libraries this exists for, so take the normal path.
+        return
+    reason = fast_exit_reason(
+        sys.platform if platform is None else platform,
+        os.environ if environ is None else environ,
+    )
+    if reason is None:
+        return
+    _end_process(config, int(_SESSION.exitstatus), reason)
```
lindkvis, line 125: **An exception raised in `pytest_sessionfinish` becomes a green Windows job.**

Storing the session by reference is the right call and it works: `_enforce_no_leaked_threads` sets `session.exitstatus = 1` rather than raising, and reading through the reference here picks that up.

The gap is the other shape. If anything in that hook body *raises* after `remember_session`, the status is never updated. In pytest's `wrap_session` only `exit.Exception` is caught around the `pytest_sessionfinish` call; any other exception escapes the `finally`, skipping that block's `config._ensure_unconfigure()`, and reaches `main()`'s own `finally: config._ensure_unconfigure()` — which runs `pytest_unconfigure`, which lands here and calls `os._exit(0)`.

Reproduced with this module unchanged, a conftest wired exactly as `tests/conftest.py` wires it, and a guardrail that raises after `remember_session`:

| arm | exit code | output |
|---|---|---|
| `PIXLSTASH_PYTEST_FAST_EXIT=1` | **0** | traceback never printed |
| `PIXLSTASH_PYTEST_FAST_EXIT=0` | 1 | `AssertionError` + traceback |

That is the failure mode `tests/test_process_teardown.py` names in its own docstring — "a harness that exits 0 on a red run is worse than the flake it replaces" — and the new step check in `ci.yml` does not catch it, because the fast exit *does* fire and the status *is* 0.

Scope, stated fairly: not reachable through a known path today. Both guardrails in that hook were read and neither raises — `_enforce_no_leaked_threads` sets the status, `_enforce_test_time_budget` documents that it deliberately never touches it — the three model-release blocks are `try/except Exception`, and `_stop_tqdm_monitors` guards internally. `trylast=True` also confines the window to this hook's own body, since a plugin raising earlier leaves `_SESSION is None` and takes the normal path. It is still worth closing: both guardrails do non-trivial work (walking thread stacks, parsing the shard spec, reaching into `terminalreporter`, writing the step summary) on the one platform where the process is in an unusual state by design.

`sys.exc_info()[0]` is populated inside `pytest_unconfigure` while an exception propagates and `None` otherwise — probed both ways — so the guard is cheap:

```python
# in _end_process, before os._exit
if status == 0 and sys.exc_info()[0] is not None:
    status = 1
```

Worth a case in the same subprocess style as the existing tests, since this is the file's stated contract.

---

.github/workflows/windows-segfault-soak.yml
```diff
@@ -0,0 +1,215 @@
...
+#
+# Push-triggered on purpose: `workflow_dispatch` requires the workflow to be on
+# the default branch (main), and this one deliberately never goes there. To run
+# it again without a commit, use "Re-run all jobs" on the run page. To change
+# the size, edit SOAK_REPEATS below and push.
+
+on:
+  push:
+    branches:
+      - 'soak/**'
```
lindkvis: **"TEMPORARY. Delete this file, and the branch it lives on" does not survive merging this PR.**

Merging into `develop` makes the file permanent there, and this trigger then fires for any future `soak/**` branch anyone pushes — deleting *this* branch no longer removes it. By the file's own accounting that is ~270 runner-minutes per push, and Windows bills at 2x, so ~540 billable minutes each time someone happens to use that branch prefix.

The soak is pre-merge evidence-gathering; it does not need to be on `develop` to do its job, and it still runs from this branch because the trigger is the branch pattern rather than the workflow's location on the default branch.

Two ways out:

- drop the soak file from this PR and leave it on the branch, so the teardown fix merges on its own merits, or
- split it into its own PR with an explicit removal step.

Either way, worth saying in the PR what removes it, since "delete the branch" stops being sufficient the moment this lands.

---

.github/workflows/windows-segfault-soak.yml
```diff
@@ -0,0 +1,215 @@
...
+            echo "::endgroup::"
+
+            if grep -q 'without interpreter finalization' soak.log; then
+              fired=yes
+            else
+              fired=no
+            fi
+            case "$status" in
+              0) ;;
+              139) crashes=$((crashes + 1)) ;;
```
lindkvis: Minor, but it bites exactly the measurement this job exists to take: a Windows access violation does not always reach bash as `139`. When the runner surfaces the native status it appears as `3221225477` (`0xC0000005`), which falls through to `other` — so the tally reads "0/3 exited 139" for a leg that crashed three times, with only the `::warning::` at the end to contradict it.

Since the summary line is what gets read as the result, counting both into the crash bucket would keep a real crash from being reported as a clean arm:

```bash
case "$status" in
  0) ;;
  139|3221225477) crashes=$((crashes + 1)) ;;
  *) other=$((other + 1)) ;;
esac
```

---

lindkvis (commented):
Reviewed `0d76922` against `develop` (merge-base `a9b8176`). The base retarget to `develop` changed no review scope — the diff against the merge-base is identical to what the PR reports, 3 commits / 5 files / +668−38.

The mechanism is sound and unusually well evidenced; the docstrings that say what the runs did and did *not* establish are the best part of this change. Three comments inline: one latent correctness hole worth closing, one scope question about the soak workflow landing on `develop`, and one crash-tally nit.

**One thing not line-anchorable: the title and body describe a different change.** Title reads "Vault: the database filename is a parameter" and the body talks about temporary vault names and `vault.db`, but the diff is entirely Windows process teardown plus the soak harness — nothing vault-related in any of the five files. The three commit subjects are accurate, so it looks like PR metadata picked up from a sibling branch. Worth fixing before merge, since the PR record is what a future reader of this change gets.

### Checked and correct — recording these so nobody re-derives them

- **Soak/job parity.** `workflow["jobs"]["backend_windows"]` is the real key (the comments' `backend-windows` is prose only). Ran the extraction snippet against this PR's `ci.yml`: 24 files, including the newly added `tests/test_process_teardown.py`. `PYTEST_FLAGS` is string-identical between the two workflows, so the soak measures the real job and not a lookalike.
- **Shell correctness.** Actions runs `shell: bash` with `-eo pipefail`, so `set +e` around the pipeline plus `${PIPESTATUS[0]}` is the right way to keep pytest's own status past the `tee`. The trailing `if` returns 0 when its condition is false, so a fired fast exit does not fail the step. The heredoc-in-`$(...)` dedents correctly under the YAML block scalar.
- **Gate classification.** `tests/test_process_teardown.py` is in the `backend` job's list in correct alphabetical position (`proc` < `proj`). Duration coverage is ~93% (157 of 169 paths timed) against the 0.9 `MINIMUM_GATE_COVERAGE` floor, so the untimed new file does not red the gate.
- **Decision matrix.** Exercised all 12 platform/env combinations against the real `fast_exit_reason`; behaviour matches the docstring exactly, including that the xdist guard outranks `PIXLSTASH_PYTEST_FAST_EXIT=1`. `exit_if_enabled` with no recorded session returns normally.
- **Capture handling is necessary, not cargo-culted.** `stop_global_capturing` is registered via `early_config.add_cleanup` in pytest's capture plugin, and `Config._ensure_unconfigure` drains `_cleanup_stack` *after* the last `pytest_unconfigure` — so without the explicit stop the notice really would land in a buffer nothing reads.
- **Lint.** `ruff check` and `ruff format --check` both clean on the two new Python files.
- **Workflow hygiene.** Third-party actions pinned by commit SHA, `permissions: contents: read`, and `HF_TOKEN` on a same-repo `push` trigger rather than `pull_request_target`.
- **No changelog fragment**, correctly — CI and test-only changes get none.

### Two smaller notes

- `os._exit` would skip pytest-cov's data write if `--cov` is ever added to the Windows job. Not used there today; worth a line in the module docstring's list of what the fast exit skips, next to the `DLL_PROCESS_DETACH` caveat.
- `pytest-windows.log` is neither cleaned up nor uploaded as an artifact. Harmless, since it is only read back in-step.

Per CLAUDE.md, the workflow changes still owe a `chief-security-officer` pass before push. Nothing here looks risky on that axis — read-only permissions, pinned SHAs, no new secret exposure — but it is a process gate rather than a judgement call.

---

tests/process_teardown.py
lindkvis: Fixed in 61f1436, and reproduced your table first: with the guard removed, the fast-exit arm exits **0** with the traceback never printed, while the control arm exits 1 and prints it.

`_end_process` now checks `sys.exc_info()` after global capture is stopped, and does two things rather than one:

```python
if sys.exc_info()[0] is not None:
    traceback.print_exc()
    if status == 0:
        status = 1
```

The print matters as much as the status here — `os._exit` would otherwise take the traceback with it, so the Windows arm would have gone red with nothing saying why. Stopping capture first keeps it out of a buffer nobody reads.

Covered by `test_a_session_finish_hook_that_raises_is_never_a_green_job`, parametrised over both arms, in the same subprocess style: the conftest body after `remember_session` is now a parameter (`_conftest(after_remember=...)`), so the test raises from exactly where `tests/conftest.py`'s guardrails sit. Both arms must end non-zero *and* print the reason. Mutation check: dropping the four lines above turns the `fast_exit=True` case red with `assert 0 != 0`, exactly the failure mode you described.

---

.github/workflows/windows-segfault-soak.yml
lindkvis: Taken, in a3bc628: both statuses count as a crash, and the two tally lines now read "crashed (139 or 0xC0000005)" so the summary cannot claim a clean arm for a leg that faulted.

```bash
case "$status" in
  0) ;;
  139|3221225477) crashes=$((crashes + 1)) ;;
  *) other=$((other + 1)) ;;
esac
```

The commit is kept ahead of b49234d, which removes the file (see the other thread), so the version anyone restores from history is the one that counts correctly.

---

.github/workflows/windows-segfault-soak.yml
lindkvis: Took your first option: b49234d deletes the file, so merging this PR leaves no `soak/**` trigger on `develop` and the teardown fix lands on its own merits.

Nothing is lost by it:

- the soak has already run once (run [35258233638](https://github.com/Pikselkroken/pixlstash/actions/runs/35258233638), 8 legs, all completed), and that run can still be re-run from its own page, because its head commit stays in this branch's history;
- to run a new one, `git show a3bc628:.github/workflows/windows-segfault-soak.yml` restores it, with the exit-code counting fix from the other thread already in it.

Worth stating plainly, since the workflow's own header says so: that run found **0 crashes in both arms**, which by its own rule means the sample was too small to see the base rate, not that the fix is proven. The fix's own evidence is the mechanism plus `tests/test_process_teardown.py`; the soak would need more repeats to say anything about the rate, and that is now a deliberate act rather than something every push to this branch pays for.

---

frontend/src/styles/designDrift.test.js
```diff
@@ -0,0 +1,117 @@
...
+  });
+}
+
+const files = sources()
+  .filter((path) => !path.endsWith(join("styles", "design-tokens.css")))
+  .map((path) => ({
+    name: relative(SRC, path),
+    text: readFileSync(path, "utf8")
+      .replace(/\/\*[\s\S]*?\*\//g, "")
+      .replace(/<!--[\s\S]*?-->/g, ""),
```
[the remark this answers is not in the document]
lindkvis: Fixed in 17f9e62. The file is byte-identical to `develop` here (`git diff origin/develop...HEAD -- frontend/src/styles/designDrift.test.js` is empty), so this alert is not introduced by this PR — it is fixed here because this is the run that reported it.

The finding is real for the stated reason: one pass of `/<!--[\s\S]*?-->/g` over `<!-- <!-- x --> -->` removes the inner comment and leaves a bare `-->` behind, and `<!-- unclosed` leaves the opener. Comment stripping is now looped to a fixed point, with the bare delimiters in the pattern:

```js
const COMMENT = /\/\*[\s\S]*?\*\/|<!--[\s\S]*?-->|<!--|-->|\/\*|\*\//g;
```

Looping is what makes that safe rather than a new hole: removing a delimiter can spell a new one (`<!<!---->` → `<!--`), and the loop only stops when a pass changes nothing, so the result provably contains no `<!--`, `-->`, `/*` or `*/`. Checked against the nested, unclosed and re-spelling cases; `npx vitest run src/styles/designDrift.test.js` is 7 passed, and `npx eslint` is clean.

---

pixlstash/services/workflow_inbox.py
```diff
@@ -0,0 +1,241 @@
...
+            or trashed. The stored workflow must then be kept.
+    """
+    digest = content_hash(workflow)
+    os.makedirs(folder, exist_ok=True)
+    stem = _split(name)[0]
+    written, counter = f"{stem}.{digest}.json", 2
+    while True:
+        try:
+            # "x": never over a file already there.
+            with open(os.path.join(folder, written), "x", encoding="utf-8") as handle:
```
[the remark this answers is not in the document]
lindkvis: Fixed in 17f9e62. This file is also byte-identical to `develop`, so the alert is not this PR's; fixed here because this run reported it.

`trash_workflow`'s `name` reaches it from the delete request, and the copy was written with a plain `os.path.join` into the inbox. It now goes through the sanitizer the rest of the codebase uses, which `pixlstash/utils/path_utils.py` documents as being placed at every path-construction site precisely so the taint analysis sees it:

```python
path = resolve_path_within(folder, written)
```

Not exploitable today — `delete_comfyui_workflow` already resolves the same name against the user workflow directory and answers 400 if it escapes — so this is the second guard on the *inbox* base, which had none. `ValueError` is now documented on the function and added to the caller's `except`, so a name that escapes is a handled 500 rather than an unhandled traceback.

Exercised against the names the function actually builds (`flow.<hash>.json`, the `flow (2).<hash>.json` collision form, one with spaces and `&`): all resolve inside the folder unchanged, while `../escape.<hash>.json`, `../../etc/x.json` and an absolute path are refused.

---

pixlstash/routes/comfyui.py
```diff
@@ -702,27 +1443,766 @@
...
+                    "min": p.minimum,
+                    "max": p.maximum,
+                    "step": p.step,
+                    "options": list(p.options) if p.options is not None else None,
+                    "multiline": p.multiline,
+                    "pinned": p.key in pinned,
+                }
+                for p in parameters
+            ],
+        }
```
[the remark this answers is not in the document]
lindkvis: Fixed in 17f9e62. `pixlstash/routes/comfyui.py` is byte-identical to `develop` too, so this alert is not introduced here either.

The flow was `comfyui_error = str(exc)` at the `fetch_object_info` call, straight into the response body. The route now composes the field itself, from the URL it already holds:

```python
comfyui_error = f"Could not read ComfyUI's node types at {comfyui_url}"
```

The exception still goes to the log at INFO with everything it carries. What the client loses is the distinction between unreachable / non-2xx / unparseable JSON, which is a distinction it cannot act on: the contract is that the values below are the file's own, and `ComfyUIWorkflowParametersResponse` already documents the field as "set only when ComfyUI could not be reached". `test_an_unreachable_comfyui_still_shows_the_recorded_values` now raises with a message carrying a token-shaped secret and asserts the response says the node types could not be read *and* does not contain it.

**Scope, stated fairly:** this is the class, not the instance, only for this endpoint. `str(exc)` reaching a response body is a repo-wide pattern — 111 occurrences under `pixlstash/routes/`, 9 in this file alone, including the LoRA-swap 502 at line ~584 and the `LookupError` detail below it, both with tests asserting the exception text. Sweeping those is its own change with its own review; CodeQL flagged this one site, and this PR fixes that one site.

---

PR #1422, opened 2026-09-17
Workflows toolbar carries Settings and the sidebar toggle

### LINKED ISSUE #1415: Workflows screen: right sidebar can't be opened (toolbar lacks Settings/Stats)
On the Workflows screen the right sidebar can't be opened. `WorkflowShelf.vue`'s toolbar has no `TbGlobalActions`, the block with Settings and the stats-sidebar toggle. The Models screen ends its toolbar with it. `WorkflowInspector` still renders on `sidebarStore.statsOpen`, so the rail is there, but nothing on this screen can open it.

## Fix

- End the Workflows toolbar with the documented tail, as `ModelShelf.vue` does: a separator, then `<TbGlobalActions @open-settings=…>` last in the bar (see `docs/design/toolbar-responsive-decisions.md`).
- Copy ModelShelf's tests into `WorkflowShelf.test.js`:
  - pressing the stats button opens the rail and a second press closes it;
  - `TbGlobalActions` is last in the bar.

## Notes

- Independent of the v1.12 Workflows & Recipes steps, so it can land whenever. It is a small PR against `develop` and needs no backend.
- **F1a (#1402) replaces this toolbar and must keep the same tail and test.** Otherwise the gap comes back when F1b (#1404) swaps the route.

## Check by hand

Open **Workflows** and press the stats button at the right end of the toolbar. The right sidebar should open with the selected workflow; press again and it should close. Settings should open from the same place.

lindkvis:
Fixes #1415.

The Workflows view replaces the grid and with it the grid's toolbar, but its
own bar never ended with `TbGlobalActions`. So nothing on that screen opened
Settings or the right rail, and `WorkflowInspector`'s content is only shown
while the rail is open — the inspector was unreachable.

- `WorkflowShelf.vue` ends its bar with `[separator] [TbGlobalActions]`,
  emitting `open-settings` up to `App.vue`, which now listens. No undo: nothing
  here writes to the operation log, the model shelf's documented exception.
- The pair sits in a `--space-3` cluster. This bar spaces its own controls at
  `--space-4`, and `TbGlobalActions` is multi-root, so without the cluster the
  tail would render 4px wider here than everywhere else — which is the one
  thing mounting the same component in every host is meant to prevent.
- `.wfshelf-toolbar` joins the bar list in `Toolbar.test.js`, the guard on the
  36px recipe and the shared right inset, now that it is a host of the tail.
- Tests: Settings emits and App.vue binds it, the stats button opens then
  closes the rail, the tail is last in the bar behind its separator, and no
  undo control is mounted. Each was shown failing before it was kept.

An adversarial pass raised three things not fixed here, deliberately:

- The bar has no `@container` collapse ladder, and the tail makes it ~80px
  wider. True, and the same is true of the Ghosts button that shipped before
  it; a ladder for this bar is its own measured piece of work, not a rider on
  restoring a missing control.
- The stats button's accessible name says "stats sidebar" while the rail here
  is the Inspector. Real, but it is `TbGlobalActions`' name in every view, so
  renaming it is an app-wide UX decision rather than a fix to this screen.
- `.wfshelf-toolbar` carries the `toolbar` class without declaring
  `container-type`. Pre-existing on `develop` and unrelated to this change;
  the tail ships no collapse rule, so nothing depends on it today.

`OverlayMetadataPanel.test.js` fails two tests on this branch and on its base
alike — from #1389, not from this change. Everything else in the frontend suite
passes.


lindkvis (commented):
Review of the tail. Four findings are on files this PR does not touch, so they have no line to hang on:

**1. Ctrl+Z still fires on this screen** — `useGlobalKeydown.js:66` declines the chord with `document.querySelector(".shelf")`, and the Workflows root is `.wfshelf`, a different class token. So on /workflows Ctrl+Z runs `operationStore.undo()` and mutates the library, with no `UndoControl` (this PR deliberately mounts none) and no `ActionReceipt` (mounted only in ImageGrid, DuplicateQueue and ImageOverlay, all unmounted here). The model shelf's undo exception was carried over for the toolbar but not for the keyboard.

**2. The same bug is left on two sibling screens** — `grep -rn toggleStats frontend/src` has exactly one UI caller, `TbGlobalActions.vue:20`. App.vue mounts five grid-replacing destinations; after this PR three carry the tail (DuplicateQueue, ModelShelf, WorkflowShelf) and two do not: `.mv-toolbar` (MovesReview) and `.ins-toolbar` (LibraryInsights). On /moves and Insights with `statsOpen` false the rail is still unreachable and Settings still has no button — #1415 unchanged. Population is three, one was fixed.

**3. The newly reachable rail can't show a run** — `App.vue:842`'s `v-if="isWorkflowsView"` shadows `WorkflowRunPanel`, while `useWorkflowRunStore.js:35` force-opens that rail when a run starts. Start a ComfyUI run and go to /workflows: the rail opens onto "Pick a workflow to see what it is made of." instead of run progress. Previously unreachable, now reachable by hand too.

**4. The new button announces the wrong surface** — `TbGlobalActions.vue:52` computes "Show/Hide stats sidebar" unconditionally, and `AppBarButton` promotes an icon-only button's tooltip to its `aria-label`. On this screen the rail is `AppInspector label="Inspector"`, and this PR's own `#wf-help` text calls it "the right sidebar's toggle, which shows the selected workflow's details". A screen-reader user hears "Show stats sidebar" for a control that opens the Inspector.

frontend/src/components/views/WorkflowShelf.vue
```diff
@@ -177,6 +178,23 @@
...
+           opened Settings or the right rail - and the rail is where
+           WorkflowInspector's content is shown, so the inspector was
+           unreachable (#1415). No undo: nothing here writes to the operation
+           log.
+
+           The cluster is not decoration: TbGlobalActions is multi-root, so its
+           buttons would otherwise sit at this bar's wider gap and the pair
+           would be 4px wider here than in every other view, which is the one
+           thing mounting the SAME component is meant to prevent. -->
+      <span class="wfshelf-bar-tail">
```
lindkvis, line 193: The tail has no shrink chain, so at narrow widths it reproduces #1415 rather than fixing it.

Every member of this bar is `white-space: nowrap` (`.bar-btn`, App.css:1109) at the flex default `min-width: auto`; `.wfshelf-title`/`.wfshelf-sub` set no `min-width: 0` or `flex-shrink`, and `.wfshelf-bar-tail` sets no `flex: 0 0 auto`. Min-content for the title + subtitle + Group/Sort/Show + Ghosts + the ~80px tail + 7 × `--space-4` gaps is ~1020px. A 1536px window with the left sidebar docked and the Inspector rail open leaves the main column ~976px, so the surplus leaves through the right edge and is clipped by the `overflow: hidden` wrapper at App.vue:752 — pressing the new toggle narrows the column and hides the button needed to close it again.

`DuplicateQueue.vue:3077` documents this exact defect verbatim and fixes it with `.dq-tb-left { min-width: 0; flex: 0 1 auto }` + `.dq-tb-right { flex: 0 0 auto }`; `Toolbar.vue:1140` uses `flex-shrink: 0`. Neither is present here.

frontend/src/components/views/WorkflowShelf.vue
```diff
@@ -177,6 +178,23 @@
...
+        <span class="bar-separator" aria-hidden="true"></span>
```
lindkvis: `TbGlobalActions` already ships this: `separator: { type: Boolean, default: false }` (TbGlobalActions.vue:41) renders a byte-identical `<span v-if="separator" class="bar-separator" aria-hidden="true">`. No host passes it, so the prop currently has zero call sites.

`<TbGlobalActions separator @open-settings="emit('open-settings')" />` is the same render with one element fewer, and the ordering test's `separator.nextElementSibling` anchor still holds. As written this is the third hand-rolled copy, and the three already disagree (the grid bar's is a bare `<div class="bar-separator">` with no `aria-hidden`).

frontend/src/components/views/WorkflowShelf.vue
```diff
@@ -177,6 +178,23 @@
...
+        <TbGlobalActions @open-settings="emit('open-settings')" />
```
lindkvis: Naming nit that the a11y tree sees: this mounts the stats toggle, whose `aria-label` resolves to "Show stats sidebar", but the rail on this screen is `AppInspector label="Inspector"`. See the review body.

frontend/src/components/views/WorkflowShelf.vue
```diff
@@ -1193,6 +1215,15 @@ onMounted(() => {
   flex: 1;
 }
 
+/* The app-wide tail's own gap, on `.shelf-bar-cluster`'s reason: this bar
+   spaces its controls at --space-4, and the tail must land at the --space-3
+   every other host lays it out at. */
+.wfshelf-bar-tail {
```
lindkvis, line 1249: `.wfshelf-toolbar` never declares `container-type`/`container-name`, so the `shelfbar toolbar` tokens added at line 39 are inert class names, not container names — they match no CSS rule anywhere in `frontend/src`.

`ModelShelf.vue:4179` declares `container-type: inline-size; container-name: shelfbar toolbar;`, and `Toolbar.vue:1115` / `DuplicateQueue.vue:3063` do the same for their bars. `docs/frontend_architecture.md:4763` states the contract: the bars "share the `toolbar` container name … so the shared chrome degrades identically at every width". Any `@container toolbar` rule in the shared chrome (`UndoControl.vue:469`/`475` already has two) silently no-ops in this host, because once the grid is unmounted no ancestor of this bar is a named `toolbar` container. The doc paragraph this PR adds at frontend_architecture.md:4226 asserts the tail is identical in each host; the CSS doesn't back that yet.

frontend/src/components/panels/Toolbar.test.js
```diff
@@ -227,8 +227,13 @@ describe("Toolbar - the shell band's one box recipe", () => {
       "src/components/views/LibraryInsights.vue",
       ".ins-toolbar",
     );
+    // The fifth bar, and the fourth host of the app-wide tail (#1415).
+    const workflows = blockOf(
```
lindkvis, line 231: `.mv-toolbar` (MovesReview.vue:427) satisfies the same recipe today — `height: 36px`, `box-sizing: border-box`, `padding: 0 var(--space-3) 0 var(--space-5)`, one `background: rgb(var(--v-theme-toolbar))` — and its own comment claims "the same 36px recipe as Insights/Duplicates/the model shelf". It would pass every assertion in the loop, so the "fifth bar" comment is only true because Moves was never added. Left out, the exact 41px/48px drift this test was written for ships unguarded there.

frontend/src/components/panels/Toolbar.test.js
```diff
@@ -261,6 +266,7 @@ describe("Toolbar - the shell band's one box recipe", () => {
     expect(rightInset(dq)).toBe(rightInset(grid));
     expect(rightInset(shelf)).toBe(rightInset(grid));
     expect(rightInset(insights)).toBe(rightInset(grid));
+    expect(rightInset(workflows)).toBe(rightInset(grid));
 
     // Two of the three bars LEAD with an identity (the queue's count, the
```
lindkvis, line 276: Stale after the loop above grew to five: four of the five bars now lead with an identity (`.qtitle`, `.shelf-title`, `.ins-title`, `.wfshelf-title`), and the block below still checks three pairs — so `.wfshelf-title`/`.wfshelf-sub` are unguarded for size and ink with nothing saying so.

frontend/src/components/views/WorkflowShelf.vue
lindkvis: Fixed in 4eb6b78b. `.wfshelf-title` gives first and from the right (`min-width: 0; flex-shrink: 6; overflow: hidden; text-overflow: ellipsis`), the subtitle after it, and `.wfshelf-bar-tail` is `flex: 0 0 auto` so the chrome is never what the edge eats. Guarded in `Toolbar.test.js`: the identity of each bar in the loop must carry `min-width: 0` and a non-zero `flex-shrink`, and the tail must not shrink. Shown failing by removing each declaration in turn.

frontend/src/components/views/WorkflowShelf.vue
lindkvis: Right, and it is now the prop: `<TbGlobalActions separator rail-name="inspector" …>` in 4eb6b78b. The two bars this PR also gives the tail to (Moves, Insights) take the prop as well, so the hand-rolled count went from three to two rather than four. The ordering test still anchors on `separator.nextElementSibling`.

frontend/src/components/views/WorkflowShelf.vue
lindkvis: Fixed in 65331009. `TbGlobalActions` takes a `railName` prop (default `"stats sidebar"`, so every other host is unchanged) and this screen passes `rail-name="inspector"`. Asserted on the aria-label in both states — "Show inspector" / "Hide inspector" — and shown failing with the prop removed.

frontend/src/components/views/WorkflowShelf.vue
lindkvis: Fixed in 4eb6b78b: `.wfshelf-toolbar` declares `container-type: inline-size; container-name: shelfbar toolbar`. You are right that nothing queries it today (the tail ships no collapse rule) — the point is that the tokens on line 39 now name a container that exists, so the next shared control mounted here degrades by the same rules as in the other hosts.

frontend/src/components/panels/Toolbar.test.js
lindkvis: Fixed in 4eb6b78b — `.mv-toolbar` is in the loop and in the right-inset comparison. It passes as written; you are right that the comment was only true because Moves had never been added. Moves also gains the tail itself in this PR, for the same #1415 reason.

frontend/src/components/panels/Toolbar.test.js
lindkvis: Fixed in 4eb6b78b. `.wfshelf-title`/`.wfshelf-sub` and `.mv-title`/`.mv-sub` are now checked against `.qtitle`/`.qsub` for size, weight and ink. The workflow pair did not pass: it shipped at `--text-lg` over `--text-sm` in the theme-on-surface ink, so this PR brings it onto the shared recipe (`--text-md` / `--text-xs`, toolbar ink) — a visible change to that bar, and the same drift the test was written for.

lindkvis:
Answering the four findings in the review body — they had no line to hang on, so they are here. Everything is on the PR head (65331009, 4eb6b78b); the six inline threads are answered and resolved on their own comments.

**1. Ctrl+Z still fires here.** Right, and on two more screens besides. `useGlobalKeydown` guarded `.shelf` alone; the invariant it was written for — the destination replaces the grid, mounts no `ActionReceipt` and no `UndoControl`, so the chord reverts a library action with nothing on screen to say it happened — is equally true of `/workflows`, `/moves` and `/insights`. The check is now `UNDO_BLIND_ROOTS = ".shelf, .wfshelf, .mv, .ins"` and the notice text is no longer shelf-specific. Tested per screen, and shown failing by narrowing the list back to `.shelf`.

**2. The same bug on two sibling screens.** Fixed with it: `.mv-toolbar` and `.ins-toolbar` end with the tail too, taking its own `separator` prop at the end of their existing right-hand group. Moves emits `open-settings` to App.vue; Insights emits `{ kind: "settings" }` on the `act` channel App.vue already routes. Both have the Settings/rail tests, including the App.vue-binding one, and `.mv-toolbar` joins the bar-recipe loop in `Toolbar.test.js`.

**3. The rail can't show a run.** Fixed: `WorkflowRunPanel` is now the `v-if` and `WorkflowInspector` the `v-else-if`, so a run outranks the inspector in the rail it force-opened. Guarded by a source assertion in `WorkflowShelf.test.js` — App.vue has no mount harness in this suite — which fails on the old ordering.

**4. The button announces the wrong surface.** Fixed: `TbGlobalActions` takes `railName` (default `"stats sidebar"`, so no other host changes) and this screen passes `rail-name="inspector"`. The `#wf-help` sentence no longer explains the discrepancy.

One thing to flag rather than fix: `frontend/src/components/views/OverlayMetadataPanel.test.js`, `LibrarySwitchOverlay.a11y.test.js`, `useDedupStore.test.js` and `useUpdatesSocket.test.js` fail 6 tests in my checkout — on this branch and on the untouched base commit alike, so not from this change. Everything else passes (4349).

PR #1423, opened 2026-09-17
Undo/redo of a rotate turns the open lightbox (#1419)

### LINKED ISSUE #1419: Undo/redo of a rotate doesn't update the open lightbox
Rotate a photo in the lightbox, then undo it (Ctrl+Z or the receipt's Undo) without closing the lightbox. The picture stays rotated until the lightbox is closed and reopened. Redo does the same, and so does a rotate made from another client while the lightbox shows that picture.

**Why:** the server's `pixels` change event is deferred while the overlay is open (`deferWhileOverlayOpen` in `useGridRealtimeSync.js`), and the overlay has no signal of its own for pixel changes. Its tag, description, smart-score, detection and text signals never re-read `orientation`, so nothing refreshes its metadata and the image URL keeps the old `?v=o<n>`.

**Likely fix:** pass a pixel-change signal from App.vue down to the overlay, the way `wsDetectionUpdate` is passed, and have the overlay re-read metadata, face boxes, detections and text when the change names the open picture. `runRotate` already does all four after a local rotate.

**Check:** open a JPEG, press `]`, then press Ctrl+Z. The picture should turn back without closing the lightbox. Then redo; it should turn again.

Found while fixing #1382 (PR #1393).

lindkvis:
Fixes #1419. Rotating in the lightbox and then undoing left the picture rotated until the lightbox was closed and reopened; so did redo, and so did a rotate from another client.

Two defects, one symptom.

**Nothing told the overlay.** `POST /pictures/rotate` broadcast a bare `updated`; its own undo path already stamped `fields: ["pixels"]`. It does now too, a new `wsPixelsUpdate` signal carries that field from the socket to the lightbox, and the overlay re-reads through `refreshAfterPixelChange` — `runRotate`'s own post-rotate block, extracted so a turn from any origin costs the same reads.

**Telling it was not enough.** `setOverlayImageById` re-seeds the open card from the sequence frozen at open, and preserved description, score and tags but not `orientation`. `applyRotatedCards` replaces the `allImages` array right after a rotate, so the undo turned the picture and the grid's own repaint turned it straight back — and because the re-seed reads the snapshot rather than the array it is handed, a correctly-rotated record did not help either. `orientation` and `pixel_sha` now survive it, on the same reasoning as `fetchOverlayMetadata`'s existing exception to its local-wins merge.

Also here:

- The `pixels` branch runs `applyRotatedCards` under an open overlay instead of deferring. The applier is fields-only, so there is nothing to keep off the frozen filmstrip; deferring it queued a whole-grid refetch for every lightbox close after a rotate.
- Five producers stamp `pixels` and only two are turns (the thumbnail sweep and layout moves do not). The overlay re-reads metadata always, and `/faces`, `/detections` and `/text` only when the picture actually turned — `pictureText.refresh()` clears the viewer's word selection, and background work should not.
- `rotateMetadataInFlight…` is a count rather than a set: the local rotate and its own echo can both be reading one picture.

Docs: `integration_architecture.md` gains §8.3 for the `pixels` contract; `frontend_architecture.md` §9.1 and `backend_architecture.md` §21.5 updated.

## On the adversarial review

An adversarial pass found the re-seed defect above, which I had missed and which would have shipped a fix that did not work — it is fixed here with a test that reproduces it. It also showed that deleting either the socket emitter or the `ImageGrid` prop binding left the whole frontend suite green, because most of the new test file asserted against copies of the source declared inside it. That layer is gone; there is a real socket test and a guard that every overlay-facing `ws*Update` is wired store → grid → prop, so the next signal added cannot go unwired. Both mutations now go red, as do the two halves of the turn gate and the orientation preservation.

Two of its findings I did not take. It read the un-deferral's justification as wrong because the repainted card is behind the lightbox and invisible — correct, and the comment and doc said so; the argument is the close-time refetch, and both now say that instead. And it read the id gate as contradicting the coalescing rationale the same file documents twice; it was right, so the gate is gone and the signal fires on every distinct key, which the turn check makes cheap.


pixlstash/routes/pictures/_crud.py
```diff
@@ -1833,12 +1833,25 @@ def rotate_pics(session, ids, direction):
...
+            # This was the one producer of a byte rewrite that did not name it;
+            # the restore above, ``ThumbnailGenerationTask`` and the layout move
+            # (route and task) all did.
             server.vault.notify(
                 EventType.CHANGED_PICTURES,
                 {
                     "picture_ids": rotated_ids,
                     "origin_client_id": origin_client_id,
                     "change_kind": "updated",
+                    "fields": ["pixels"],
```
lindkvis: Stamping the forward rotate with `fields: ["pixels"]` makes the issuing tab stop suppressing its own echo, so `applyRotatedCards` now runs twice for every rotate.

Before this line the forward-rotate event carried no `fields`, so `handleOwnOrigin` returned `SUPPRESSED, "own-origin-echo"` in the tab that issued it. With `fields: ["pixels"]` the event is caught by the card-content branch (useGridRealtimeSync.js:646), which runs **before** the origin dispatch and is origin-blind.

Rotate 20 pictures from the grid toolbar: `ImageGrid.runRotate` awaits `applyRotatedCards(rotated)` (ImageGrid.vue:4351) and the WS echo fires `applyRotatedCards(pictureIds)` again — 2× `POST /pictures/thumbnails`, 2×20 forced `fetchImageInfo` reads, 40 off-screen bitmap decodes.

Worse than the cost: both runs take `allGridImages.value.slice()` before their awaits and assign after, so any grid write landing in between (a score edit, a `removed` event, a slick insert) is silently reverted by the later write.

frontend/src/composables/useGridRealtimeSync.js
```diff
@@ -624,6 +624,28 @@ export function useGridRealtimeSync(deps) {
       if (pictureIds.length > MAX_TARGETED_UPDATE) {
         return reloadOrDefer("card-content-refresh-too-large");
```
lindkvis: A bulk rotate of 51–200 pictures now triggers a full grid refetch **in the tab that issued it**.

`ROTATE_MAX_IDS = 200` (_crud.py:91) against `MAX_TARGETED_UPDATE = 50`. Select 80 pictures and rotate: the echo now carries `fields: ["pixels"]`, so `fieldsAreCardContentOnly` is true, `pictureIds.length > 50`, and this returns `reloadOrDefer("card-content-refresh-too-large")` → `fullGridReload()` → `gridStore.refreshGridVersion()` — a whole-library refetch in the tab that already applied the turn optimistically. Before this PR that echo was `SUPPRESSED`.

With an overlay open it instead sets `pendingOverlayGridRefresh`, so the lightbox close pays the refetch this PR's own description says it is removing.

frontend/src/composables/useGridRealtimeSync.js
```diff
@@ -624,6 +624,28 @@ export function useGridRealtimeSync(deps) {
...
+      // the grid - so there is no restructuring to keep off the frozen
+      // filmstrip.
+      //
+      // What deferring it actually cost is the close: marking a deferred
+      // refresh queues a WHOLE-GRID refetch for `closeOverlay`, so every
+      // lightbox session containing a rotate paid one, to repair a card a
+      // fields-only write had already repaired. The card itself is behind the
+      // overlay and invisible until then - that is not the argument, and the
+      // filmstrip reads the frozen snapshot either way.
+      if (fields.includes("pixels")) {
```
lindkvis, line 636: The un-deferral is applied to the whole `pixels` class rather than to turns, so background thumbnail regeneration now mutates the grid under an open lightbox.

The justification given ("deferring queued a whole-grid refetch for every lightbox close after a rotate") is about turns, but this branch keys on `fields.includes("pixels")`, which three non-turn producers also stamp. During an import, every ≤50-id `ThumbnailGenerationTask` batch runs `applyRotatedCards` under the open overlay: `POST /pictures/thumbnails`, one forced `fetchImageInfo` per id, one decode per id blocking up to `ROTATE_BITMAP_WAIT_MS` = 5 s each, then a replacement of `allGridImages` — which fires the overlay's `allImages` watcher → `applyAllImagesUpdate` → `setOverlayImageById` re-seed and `invalidateThumbnailIndex` per patched row.

Cheaper shape: un-defer only when the event is a turn (or mark a targeted repair instead of `markOverlayDeferredRefresh`'s whole-grid one), keeping the deferral for the byte-rewrite case.

frontend/src/composables/useGridRealtimeSync.js
```diff
@@ -624,6 +624,28 @@ export function useGridRealtimeSync(deps) {
...
+        void grid.applyRotatedCards?.(pictureIds);
```
lindkvis, line 673: Un-deferring removes the only marker that the grid changed during the overlay session, so a snapshot captured before the turn silently discards the applier's write at close.

A streaming/id-list fetch that started before the rotate stores its result in `pendingGridImages` while the overlay is open (§9.1). The applier writes the turned cards straight into the live `allGridImages`. `closeOverlay` (ImageGrid.vue:6529) takes the `pendingGridImages !== null` branch first, assigns the pre-rotate snapshot over the applier's write, and clears `pendingTagFilterRefresh`/`pendingOverlayGridRefresh` — the flag the old deferring branch set, which would have forced the repairing refetch.

Net: the card behind the lightbox shows the pre-rotate bitmap and shape after close, with nothing queued to fix it.

frontend/src/components/views/ImageOverlay.vue
```diff
@@ -1530,6 +1553,73 @@ function rotateTitle(side, shortcut) {
...
+  const imageIdKey = String(imageId);
+  const before = knownOrientation(
+    String(image.value?.id) === imageIdKey ? image.value?.orientation : null,
+  );
+  rotateMetadataInFlightByImageId.set(
+    imageIdKey,
+    (rotateMetadataInFlightByImageId.get(imageIdKey) || 0) + 1,
+  );
+  try {
+    await fetchOverlayMetadata(imageId);
```
lindkvis, line 1601: Turn detection depends on a metadata read that the shared `metadataRequestId` counter can silently discard, and there is no retry.

`fetchOverlayMetadata` bumps a single module-scope counter (line 3820) and drops its own response when superseded. `handleDescriptionUpdate` bumps it unconditionally (line 4618), as do the tag / description / smart-score watchers and the `image.value?.id` watcher.

Undo a rotate in another tab while a description save is in flight here: `refreshAfterPixelChange` issues read N, the save bumps the counter, read N returns and is discarded, `image.value.orientation` never moves, so `knownOrientation(...) !== before` is false, `turned` is false, and `/faces`, `/detections` and `/text` are never re-read — while `fullImageSrc` keeps the pre-rotate `?v=o<n>`. Nothing is queued to retry, so the lightbox stays wrong until it is closed and reopened. That is #1419 again.

frontend/src/components/views/ImageOverlay.vue
```diff
@@ -1530,6 +1553,73 @@ function rotateTitle(side, shortcut) {
...
+    else rotateMetadataInFlightByImageId.delete(imageIdKey);
+  }
+  if (String(image.value?.id) !== imageIdKey) return;
+  // Everything below is invalidated by a TURN, not by a byte rewrite as such.
+  // Three of the five producers of a `pixels` event never turn anything - a
+  // background thumbnail regeneration, a layout move and its migration - and
+  // re-reading for those would cost three requests per batch AND take the
+  // viewer's word selection away with them (`pictureText.refresh` clears it),
+  // for a change that did not move a box or a word. The same gate absorbs the
+  // echo of a rotate this overlay has already applied: the second read finds
```
lindkvis: `before` is captured from `image.value` before the await, but only the **id** is checked after it, so a navigate-away-and-back during the read compares against the wrong record's orientation.

Open card 7 (orientation 6), a `pixels` frame fires the watcher, `before = 6`. While the metadata read is out the user presses right then left, back to 7. The id guard passes (`image.value.id` is 7 again), but `image.value` was rebuilt by `setOverlayImageById` from the frozen snapshot in between, so `knownOrientation(image.value?.orientation)` is whatever that snapshot carried — a spurious inequality re-reads boxes and text (clearing the word selection), a spurious equality skips a real turn.

Capturing the id-plus-orientation pair, or re-deriving `before` from the record the read actually merged, closes it.

frontend/src/components/views/ImageOverlay.vue
```diff
@@ -1530,6 +1553,73 @@ function rotateTitle(side, shortcut) {
...
+  // the orientation where the first left it.
+  const turned =
+    knownTurn ||
+    before === null ||
```
lindkvis: `before === null` classifies every `pixels` frame as a turn when the orientation is unknown, so a background batch naming other pictures does the exact harm the gate is documented to prevent.

`orientation` is NULL for videos (image_utils.py:931 leaves it None when `Image.open` fails) and for rows `MissingOrientationFinder` has not backfilled. Open the lightbox on a video while `ThumbnailGenerationTask` sweeps: each batch emits `fields: ["pixels"]` with up to 64 unrelated ids, the ungated watcher calls `refreshAfterPixelChange`, `before` is `null`, `turned` is true, and `fetchFaceBboxes` / `fetchDetections` / `pictureText.refresh()` all fire.

`refresh()` → `load(id)` sets `selectedWords.value = []` (usePictureText.js:183) and clears `rereadingId`, so the user's word selection vanishes and a running "Read again" loses its busy state — repeatedly, for a change that turned nothing.

frontend/src/components/views/ImageOverlay.vue
```diff
@@ -4439,6 +4517,39 @@ watch(
...
+    // smart-score and detection watchers above give: two signals written before
+    // this watcher flushes coalesce to the later value, dropping an
+    // intermediate one that named the open card. `pixels` is the most batched
+    // of the four - a rotate takes up to 200 ids, the thumbnail sweep announces
+    // 64 at a time - so the frame that names this picture is the likeliest of
+    // all of them to be the one that gets dropped, and dropping it is #1419
+    // again. Firing always is cheap because `refreshAfterPixelChange` is: one
+    // requestId-deduped metadata read, and the boxes and text only if that read
+    // shows the picture actually turned.
+    void refreshAfterPixelChange(image.value.id);
```
lindkvis: The coalescing rationale for firing on every frame doesn't hold for a socket-driven ref, so the ungating buys nothing and costs a metadata GET per unrelated frame.

`wsPixelsUpdate` is written at most once per `ws.onmessage`, and each WebSocket message is its own macrotask; Vue's pre-flush watcher queue is drained by `nextTick`, a microtask, between macrotasks. Two frames therefore cannot coalesce into one flush — which is why `wsTextUpdate`, whose frames are likewise one per message, **is** id-gated at ~line 4500.

With the gate removed, a `ThumbnailGenerationTask` sweep over a 5k-picture import issues one `GET /pictures/{id}/metadata?smart_score=true` on the open card per batch, none of which name it. `metadataRequestId` dedupes the *responses*, not the requests.

frontend/src/components/views/ImageOverlay.vue
```diff
@@ -1413,6 +1417,14 @@ function setOverlayImageById(nextId) {
...
       // orientation puts the pre-rotate `?v=` back on the <img>.
+      //
+      // `target` is usually the sequence FROZEN when the overlay opened, and
+      // this runs on every replacement of the `allImages` array - which is what
+      // `ImageGrid.applyRotatedCards` does right after a rotate. So the same
+      // clobber undid a turn that arrived over the socket: an undo turned the
+      // picture back and the grid's own repaint turned it forward again a
+      // moment later (#1419). Reading the snapshot rather than the array it is
+      // handed, it did that even for a correctly-rotated new record.
       ...(isSameImage && image.value?.orientation != null
```
lindkvis: Minor, but it sets a trap for the next edit: this preservation and the `pixel_sha` one below read `image.value` inline inside the object literal being assigned to `image.value`, unlike the three sibling preservations (`existingDescription`, `existingSmartScore`, `existingScore`) which capture locals first.

Correct today only because the RHS is fully evaluated before the assignment. The moment anyone splits the literal — builds `next` then patches it, or hoists a branch — these two silently read the value they were about to overwrite and the preservation becomes a no-op, which is #1419 again.

Capturing `existingOrientation` / `existingPixelSha` beside the other three costs two lines and removes it. (CLAUDE.md, Patch Reliability Policy: "Reject illogical edits. Check every patch for abrupt changes that don't fit the surrounding code.")

frontend/src/components/views/OverlaySignalsAreWired.test.js
```diff
@@ -0,0 +1,70 @@
...
+      );
+    },
+  );
+
+  it("names every signal the store actually exposes to the overlay", () => {
+    // The point of the list above is to be complete. A new `ws*Update` ref that
+    // nobody added here would slip through the per-signal checks entirely, so
+    // catch it by counting: every `ws…Update` the store declares is either in
+    // the list or is the grid's own.
+    const GRID_OWNED = new Set(["wsTagUpdate"]);
```
lindkvis: The exclusion rests on a factually wrong premise, and it leaves a hole for the one signal it skips.

The comment at line 26 says `wsTagUpdate` is "deliberately absent: the GRID consumes that one …, not the lightbox". But `ImageGrid.vue:8` binds `:tagUpdate="wsStore.wsTagUpdate"`, `ImageOverlay.vue` declares a `tagUpdate` prop, and its watcher (~line 4419) calls `fetchOverlayMetadata`.

Delete that one line of template and the overlay stops re-reading metadata on a foreign tag edit — and this guard, written precisely because "nothing that mounts ImageOverlay and sets the prop by hand can see it missing", stays green, as does the completeness count that filters it out.

docs/frontend_architecture.md
```diff
@@ -2136,7 +2136,13 @@ While the lightbox overlay is open, the user's own in-overlay edits (and any oth
...
+   - The deferral covers the **grid**, not the overlay's own content. A change to something the lightbox itself renders must therefore reach it through a dedicated signal, or it stays stale until the overlay is closed and reopened: `wsSmartScoreUpdate` (metadata panel score), `wsDetectionUpdate` (object boxes, re-read from `/pictures/{id}/detections` — otherwise a Segment run started from the overlay context menu showed nothing until reopen), `wsTextUpdate` (the text read from the picture) and `wsPixelsUpdate` (the picture's own bytes — see below) exist for that reason. The score and detection signals fire on any distinct signal key while a card is open, without gating on the payload's `picture_ids`, because signals written in one Vue flush coalesce to the last one, which may not name the open card; the text and pixel signals ARE id-gated, because each carries one gesture on named pictures.
```
lindkvis: §9.1 now says both things about `wsPixelsUpdate`, and the code does the second.

This line: "the text and pixel signals ARE id-gated, because each carries one gesture on named pictures." Line 2141: "Like the score and detection signals and for the same coalescing reason, it is **not** gated on the payload's `picture_ids`." ImageOverlay.vue:4530-4549 does not consult `payload.pictureIds`.

A reader following this sentence will assume unrelated frames are filtered out and won't go looking for the cost the other sentence accepts. The PR body says the gate was removed late; this is the leftover.

docs/integration_architecture.md
```diff
@@ -916,15 +916,59 @@ The picture-event policy lives in [`useGridRealtimeSync.js`](../frontend/src/com
...
+
+**The overlay must survive that write.** `applyRotatedCards` replaces the
+`allGridImages` array, and the lightbox re-seeds its open card from the sequence
+frozen at open whenever that prop moves - so without care an undo turns the
+picture and the grid's own repaint turns it straight back (#1419). Two things
+stop it, both on the frontend: `fetchOverlayMetadata` patches that snapshot's
+`orientation`, and the re-seed preserves `orientation` / `pixel_sha` for rows
+the patch cannot reach, exactly as the metadata merge excepts the same two
+fields from local-wins. See `frontend_architecture.md` §9.1.
   There is no `MAX_TARGETED_UPDATE` escalation here, deliberately: one read is
```
lindkvis: Inserting §8.3 between the `stack_count` bullet and its closing paragraph orphans that paragraph onto `pixels`, where its claim is false.

"There is no `MAX_TARGETED_UPDATE` escalation here, deliberately: one read is not a fetch storm, and the reload it would escalate to is precisely what must not happen while a ghost window is open" was the `stack_count` bullet's qualification (its sibling copy sits at §8.2, lines 900-908). It now ends §8.3 `pixels`, still indented as a list continuation with no list above it.

For `pixels` it is wrong: `handlePictureEvent` escalates a batch over 50 ids with `reloadOrDefer("card-content-refresh-too-large")` — exactly the reload the sentence says never happens. A 200-id rotate and a 64-id thumbnail batch both exceed it.

lindkvis (commented):
Review of the rotate/`pixels` signal change. Everything is inline. The three I'd block on are the first two comments on `useGridRealtimeSync.js` and the `before === null` branch in `refreshAfterPixelChange` — the first two are regressions this PR introduces inside its own fix path.

pixlstash/routes/pictures/_crud.py
lindkvis: Fixed in 81660de5, and one half of this does not hold.

The double apply is real and is gone: the branch now runs the applier only for a **turn** (`fields` containing `orientation`), and the non-turn `pixels` rewrites keep the ordinary deferral. For the issuing tab the duplicate remains by design, exactly as the `restored` branch above documents — suppression assumes an optimistic local op applied the change, and an undo has none, so suppressing own-origin here would re-break the case this PR exists for. What made the duplicate expensive was the escalation, addressed on the next thread.

**The silent-revert mechanism does not reproduce.** `applyRotatedCards` takes `allGridImages.value.slice()` *after* its awaits, not before — the last await is the `preloadBitmap` batch on the line above, and the slice, the patch loop and the assignment are one synchronous run with no suspension point between them (`ImageGrid.vue:4306-4324`). So a write landing during the reads is included in the slice rather than reverted by it, and two concurrent runs each read-modify-write atomically. I have left that as it is; if you were reading a different interleaving I would rather fix it than leave it, so say which.

frontend/src/composables/useGridRealtimeSync.js
lindkvis: Fixed in 81660de5. `MAX_TARGETED_UPDATE` is now checked only on the per-id `refreshGridImage` path, below the applier branch.

The cap is written for that loop — its own comment says "one /metadata + thumbnail fetch each" — and the applier is not that shape: one batched `POST /pictures/thumbnails` for the whole set. The local path already proves the size is acceptable, since `ImageGrid.runRotate` calls `applyRotatedCards(rotated)` with up to `ROTATE_MAX_IDS` ids itself. Pinned by `useGridRealtimeSync.test.js` "does not escalate a bulk turn to a grid reload" (80 ids, own-origin); I mutated the cap back onto the branch and watched it go red.

frontend/src/composables/useGridRealtimeSync.js
lindkvis: Fixed in 81660de5, along your suggested line — un-defer turns only.

`POST /pictures/rotate` and the operation-log restore now stamp `orientation` alongside `pixels`, so the branch can tell them apart. A turn runs the applier; a non-turn byte rewrite takes `card-content-bytes-overlay-deferred` under an open lightbox and the applier otherwise (its card still needs the applier rather than `refreshGridImage` — the thumbnail URL is not on `/metadata` — but it is background work in a steady stream, not a gesture waiting to land). Two new tests cover both, and inverting the turn check reds them.

frontend/src/composables/useGridRealtimeSync.js
lindkvis: Fixed in 81660de5. Confirmed by reading `closeOverlay`: the `pendingGridImages !== null` branch runs first, assigns wholesale, and clears `pendingTagFilterRefresh` / `pendingOverlayGridRefresh` with it, so the write is discarded and nothing is queued.

The turn branch now asks `grid.hasPendingGridImages()` (new accessor on the Tier-3 API, forwarded through `gridApi`) and, when a list is parked, marks the deferred refresh instead of writing in place — reason `card-content-rotate-pending-snapshot-deferred`. Only that case pays the close-time refetch. Pinned by "defers under an open overlay when a grid snapshot is waiting for close"; stubbing the accessor to false reds it.

frontend/src/components/views/ImageOverlay.vue
lindkvis: Fixed in 81660de5, by deleting the inference rather than hardening it.

With `orientation` now on the event, neither caller infers anything: `runRotate` knows from the ids the server said it turned, and the watcher knows from the field. `refreshAfterPixelChange` is gone; `refreshAfterTurn` re-reads metadata, boxes and text unconditionally, so a discarded metadata response can no longer decide whether the boxes and text are re-read. Its docstring records all three ways the comparison was wrong, yours included.

frontend/src/components/views/ImageOverlay.vue
lindkvis: Fixed in 81660de5, by the same deletion. There is no `before` any more — nothing is captured across the await except the id, which is still checked after it for the reason it always was. The navigate-away-and-back window you describe can no longer produce either the spurious re-read or the spurious skip, because no comparison happens.

frontend/src/components/views/ImageOverlay.vue
lindkvis: Fixed in 81660de5, and thank you for the video case — `orientation` being NULL for anything `Image.open` fails on made the null branch strictly worse than useless, since it fired hardest exactly where a turn was impossible.

The gate is gone with the rest of the inference, and the watcher is id-gated again, so a batch naming other pictures now costs nothing at all rather than one metadata read. Pinned by "does nothing at all for a turn that names other pictures", which asserts the total call count is unchanged.

frontend/src/components/views/ImageOverlay.vue
lindkvis: You are right, and the id gate is back in 81660de5.

I checked the macrotask argument before acting on it: `wsOrientationUpdate` is written at most once per `ws.onmessage`, each message is its own macrotask, and Vue drains the pre-flush queue on the microtask between them — so two frames cannot reach one flush, and `wsTextUpdate` being gated for the same reason is the precedent I should have followed instead of the score/detection one. The watcher comment now says this rather than the coalescing claim, and §9.1 and §8.3 say it too.

frontend/src/components/views/ImageOverlay.vue
lindkvis: Fixed in 81660de5 — `existingOrientation` / `existingPixelSha` are captured beside the other three, with a comment naming the trap.

For the record on provenance: that inline spelling arrived from `develop` in 043c0fc5 and I took it over my own hoisted version when resolving the merge, on the reasoning that the base's reviewed code should win a tie. Your point is that it is not a tie, and I agree.

frontend/src/components/views/OverlaySignalsAreWired.test.js
lindkvis: Fixed in 81660de5. You are right that the premise was false: `ImageGrid.vue:8` binds `:tagUpdate="wsStore.wsTagUpdate"` and the overlay watcher re-reads metadata off it, so the template line is load-bearing for the lightbox.

`wsTagUpdate` is in the list now, and the completeness check no longer has an exclusion set at all — an exclusion is how the hole got in, so the assertion is simply "every `ws…Update` the store declares is in this list". Deleting the `tagUpdate` binding now reds the guard; I checked.

docs/frontend_architecture.md
lindkvis: Fixed in 81660de5 — leftover from removing the gate late, exactly as you read it.

Both sentences are now one, and it is the true one: the text and orientation signals ARE id-gated, because the coalescing that ungates the score and detection signals cannot happen to a ref written from the socket. §8.3 of the integration document says the same thing so the two cannot drift apart again.

docs/integration_architecture.md
lindkvis: Fixed in 81660de5. §8.3 now sits after §8.2 closes, so the `MAX_TARGETED_UPDATE` paragraph is back under the `stack_count` bullet it qualifies, still indented as its list continuation.

The claim would also have been wrong where it landed, and is now stated properly for `pixels` in §8.3: the cap does not apply to the applier (one batched thumbnails POST, not a per-id loop) and does apply to the per-id `refreshGridImage` path — which is the code change on the second thread, so the document and the branch now say the same thing.

lindkvis:
Review addressed in 81660de5. Eleven threads replied and resolved; the `_crud.py` one is replied and left **open** deliberately, because I disagree with half of it and that is the reviewer's call, not mine to close.

**The shape of the fix.** Five of the twelve comments were the same root cause: the client could not tell a *turn* from any other byte rewrite, so it inferred one by comparing the orientation before and after a metadata read. `POST /pictures/rotate` and the operation-log restore now stamp `orientation` alongside `pixels`, and the guessing goes away with it:

- the grid un-defers only turns, so a background thumbnail batch no longer runs the applier and its blocking decode under an open lightbox;
- the overlay re-reads unconditionally instead of inferring, so a discarded metadata response cannot leave it stale with no retry;
- the id gate is back, so the `before === null` branch that treated every video as turned is gone;
- `MAX_TARGETED_UPDATE` is checked only on the per-id `refreshGridImage` path it was written for, so a 51-200 picture rotate no longer reloads the whole library in the tab that issued it;
- a turn defers after all when `pendingGridImages` is parked for close, since that branch of `closeOverlay` would discard the write with nothing queued to repair it.

**Where I pushed back.** The double-apply on the first thread is real and its escalation is fixed, but the silent-revert mechanism given for it does not reproduce: `applyRotatedCards` slices `allGridImages` *after* its awaits, so the slice, the patch loop and the assignment are one synchronous run and an interleaving write is included rather than reverted. I left that thread open in case you were reading an interleaving I have not found.

**Verification.** Each fix has an assertion that fails without it — I mutated the cap back onto the applier path, inverted the turn check, stubbed `hasPendingGridImages` to false, removed the id gate, dropped the backend field, and deleted the `tagUpdate` binding, and watched each one go red. The wiring guard now covers `wsTagUpdate` too, whose exclusion rested on a premise that was simply wrong.

Four frontend tests fail on this branch (`useDedupStore`, `LibrarySwitchOverlay.a11y`, one in `useUpdatesSocket`). They are order-dependent and pre-existing: I ran the full suite with this commit's changes reverted and got the identical four.

Issue #1394, opened 2026-09-17
Workflows & Recipes B2: hub tables, backfill and filing hook (irreversible)


lindkvis:
Part of **v1.12 Workflows & Recipes**. The design is round 2, alternative C, "Stack Opens with Grid or List" (the design bundle `design/1.12-workflows-recipes/` in the business repo). Build what the design decides and use shipped components for the chrome. Proceed on the plan's recommended answers to the open owner decisions unless the owner has recorded otherwise in that bundle's `DECISIONS.md`.

Rules for every step: one PR against `develop`, about 600 lines of diff at most, shippable alone. Until F1b swaps the route, the shipped Workflows shelf keeps working. In code and docs, `workflow_recipe` / `structural_hash` is a **variant** and the user's Recipe is a **saved recipe**. The hub tables are append-only and are never renamed.

## Depends on

| Dependency | Hard? | Check it is in `origin/develop` |
|---|---|---|
| B1 workflow identity (#1390) | yes | `git cat-file -e origin/develop:pixlstash/services/workflow_identity.py` |
| #1416 checkpoint_id as an asset (so cards are not keyed wrong first) | yes | `git grep -q checkpoint_id origin/develop -- pixlstash/services/workflow_hash.py` |

## Before you start (mandatory)

1. **Start from the latest `origin/develop`:** `git fetch origin && git checkout -b <branch> origin/develop`. If you were handed a worktree that already exists, `git fetch origin` and rebase it onto `origin/develop` before touching anything. Do not build on a stale checkout.
2. **Check that every dependency below is actually in `origin/develop`,** not just closed or badged MERGED. Merges to develop do not auto-close issues. For each dependency, find its PR (`gh pr list --state merged --search "<issue or PR number>"`), confirm `git merge-base --is-ancestor <merge commit> origin/develop`, and run the check listed.
3. **If any hard dependency is missing, stop.** Comment on this issue saying which one, and do not start. Never base a PR on another PR's branch.

## What

The hub tables that key workflows by content. **New tables only, amended into schema v2** (`CREATE TABLE IF NOT EXISTS`). **Do not bump `CURRENT_SCHEMA_VERSION` to 3.** A released build expects 2 and refuses a newer hub with `HubSchemaTooNewError`, which would lock the owner out of a downgrade. `pixlstash/hub/schema.py` amends v2 for exactly this reason, as the model shelf did.

[...]

PR #1424, opened 2026-09-17
Hash the model a PixlStash shelf loader names

### LINKED ISSUE #1416: Workflow hash: PixlStash checkpoint loader's checkpoint_id is not an asset
`PixlStashCheckpointLoader` (ComfyUI-PixlStash) names its model by `checkpoint_id`, a plain id with no file extension. `workflow_hash.structural_widget_value` nulls it, so:

- the structural (variant) hash does not change when the checkpoint does;
- `workflow_identity.slots` never sees it. Two workflows on the house loader with different checkpoints get the **same** `workflow_key`, and "other checkpoint" can never fire for them.

The VAE and CLIP loaders name their models by `*_sha256` and are fine.

Found in review of #1390 (B1).

## Fix

- Treat `checkpoint_id` on the PixlStash checkpoint loader as a topology asset in `workflow_hash`.
- This re-keys the structural hash of every graph using that loader. Decide whether that needs a `HASH_VERSION` bump and a re-file of affected hub rows, or can be amended before release.
- Land it **before the B2 backfill (#1394)**, so cards are not keyed wrong first and re-keyed later.

## Acceptance

- Two graphs differing only in `checkpoint_id` get different structural hashes and different workflow keys, shown failing under mutation.
- Graphs not using the loader keep their hashes.

lindkvis:
Closes #1416.

`PixlStashCheckpointLoader` names its checkpoint by `checkpoint_id` — a shelf
row id, no extension — and `structural_widget_value` nulled it, so the variant
hash did not move when the checkpoint did and every graph on that loader shared
one `workflow_key`.

## What changed

- `SHELF_ID_FIELD_RE` makes `checkpoint_id` a topology asset, guarded by
  `SHELF_ID_VALUE_RE` (`^\d+$`) — the node's own `isdigit()` contract. Without a
  value check the rule is name-only, and any node with a widget of that name
  would write whatever it holds into `workflow_recipe_asset`, which is kept
  forever and shared.
- `SHA256_FIELD_RE` widened to `(^|_)sha256(_\d+)?$`. **The issue says the CLIP
  loader is fine; it is not.** `clip_sha256_2` — the T5 or Llama beside a
  clip-l on Flux/SD3/HiDream — was nulled by the same anchoring. Same class,
  same fix, rather than a second card.
- An empty value on either path now names no model. A shelf widget is empty
  until Browse is clicked and the second encoder is empty on *every* SD/SDXL
  graph, and keeping it filed a junk asset row and an `asset:e3b0c442…`
  reference — the digest of the empty string — into a document whose purpose is
  to say which model went there.
- `checkpoint_id` joins `_CHECKPOINT_WIDGETS` so the chip reads `other
  checkpoint`, and `BASE_WIDGETS` in `frontend/src/utils/workflowShelf.js`,
  which mirrors it. Out of step, a workflow that has a checkpoint reported none.

## Decisions a reviewer should weigh

**No `HASH_VERSION` bump, and no re-key migration.** The issue offered either
that or amending before release; this takes the amend. The workflow library is
unreleased, so v1 has never shipped and the corrected rule *is* v1. I wrote the
migration first and dropped it: re-keying files the graph under a **new**
structural hash, so `new_recipe` is truthy, so the `revisit` guard in
`hub/workflows.py:186` does not fire and the asset names are written fresh from
the picture on disk — **including names the owner had forgotten**.
`forget_model_ghosts` documents that invariant itself ("no hash moves, so the
workflow still groups"), and a forgotten name is a privacy choice, not a cache.
Carrying forgets across a re-key is real machinery, and the hub is a separate
database an alembic migration cannot sweep, so the same step would also need to
repoint `workflow_recipe_instance.structural_hash` (its key is the instance
hash, which does not move) and clear the orphaned `workflow_recipe` rows that
would otherwise show as phantom variants.

So **existing pictures keep the grouping they have; only new reads use the new
rule.** If #1394's backfill needs the old rows corrected, that re-key is its own
step and should land first — I would rather say that than ship a half-sweep
whose residue I cannot clean.

**A shelf row id is not a content address**, and an adversarial pass was right
to push on it: `model.id` is per-machine AUTOINCREMENT, and `checkpoint_hash_task`
deletes the losing row when two paths hash alike, so a merge moves the id under
an already-filed recipe. Both argue for resolving the id to a digest at file
time. I have not done that here — it is a design change past this ticket, the id
is the only identity the executed graph carries, and a non-portable identity
that is present beats none at all, which is the bug. Documented in
`backend_architecture.md` rather than left to be rediscovered.

**Known cosmetic consequence:** the Models column now shows a bare number for
these workflows. That is the same gap the `*_sha256` loaders already have — the
frontend cannot turn an id or a digest into a name — and closing it means a
resolver, which is the same open decision.

## Tests

Four in `test_workflow_identity.py` (the checkpoint splits variants and cards,
the slot is named, the second encoder counts, empty and non-digit values name
nothing) and one in `workflowShelf.test.js`. Every one shown red by mutating the
rule it guards and restoring — including the two new value guards. The pinned-
hash test is a blast-radius guard and says so in its docstring: it cannot fail
for a change confined to these widget names.


pixlstash/services/workflow_hash.py
```diff
@@ -131,15 +131,40 @@
...
+# the extension rules below cannot see them. Without this a LoRA swap on a
+# PixlStash node would leave the recipe unchanged, which is the one error the
+# spec calls unrecoverable.
+#
+# A numbered suffix is ComfyUI's convention for a second slot on one node, and
+# it names a model exactly as the first does: the PixlStash CLIP loader's
+# `clip_sha256_2` is the T5 or Llama beside a clip-l, and `lora_name_2` is a
+# stacker's second LoRA. Anchoring on `sha256$` alone read the pair's second
+# encoder as a parameter and nulled it (#1416).
+SHA256_FIELD_RE = re.compile(r"(^|_)sha256(_\d+)?$")
```
lindkvis, line 151: `SHA256_FIELD_RE` is widened to numbered suffixes here but `LORA_DIGEST_FIELDS` is not, so only half the pair moved.

A LoRA stacker naming its second adapter `lora_sha256_2` or `adapter_sha256_2` now matches `(^|_)sha256(_\d+)?$` and becomes a topology asset (it was nulled before this PR). `_is_lora_widget` (workflow_identity.py:137) tests exact membership in `LORA_DIGEST_FIELDS = ("adapter_sha256", "lora_sha256")`, so the slot lands `is_lora=False` and `workflow_key` (workflow_identity.py:237) includes it with no mark check — swapping a character LoRA in that slot forks the workflow into a new card, which is the error `guess_mark`'s "precision beats recall" rule exists to prevent.

The filename counterpart already handles the numbered form (`LORA_FILENAME_FIELD_RE = ^lora_name(_\d+)?$`), and this PR's own comment cites `lora_name_2` as the analogy for widening.

pixlstash/services/workflow_hash.py
```diff
@@ -131,15 +131,40 @@
...
 
+# The one shelf loader that cannot address its model by digest.
+# `PixlStashCheckpointLoader` takes a `checkpoint_id`, because a model's
+# `sha256` is NULL until the background hasher has read the file and a 24 GB
+# checkpoint is listable long before that, so the node's picker writes the
+# shelf's row id instead. A bare id carries no extension and is not a digest,
+# so neither rule above saw it: the structural hash did not move when the
+# checkpoint did, and every graph on that loader shared one workflow key
+# whatever it loaded (#1416).
+SHELF_ID_FIELD_RE = re.compile(r"^checkpoint_id$")
```
lindkvis: `SHELF_ID_FIELD_RE = re.compile(r"^checkpoint_id$")` is a literal equality dressed as a pattern, and `SHELF_ID_VALUE_RE = re.compile(r"^\d+$")` is a looser spelling of `str.isdigit()` — looser exactly where it matters (see the next comment).

Two module-level constants and two `re` calls per widget buy nothing over the contract the comment already cites:

```python
if name == "checkpoint_id":
    return value if value.isdigit() and len(value) <= 12 else None
```

Shorter, and correct on both edges.

pixlstash/services/workflow_hash.py
```diff
@@ -131,15 +131,40 @@
...
+# And what such a value must look like: the node itself refuses anything that
+# is not `str.isdigit()`, so the hash rule is held to the node's own contract.
+# Without it the rule is name-only, and a third-party node with a widget of
+# that name would write whatever it holds -- prose, a newline, a kilobyte --
+# into `workflow_recipe_asset`, which is kept forever and is the tier the
+# design calls safe to share. The two guards below (no newline, 255 bytes) sit
+# AFTER this branch and would not catch it.
+SHELF_ID_VALUE_RE = re.compile(r"^\d+$")
```
lindkvis: This guard fails on both dimensions the comment above it claims it covers.

**Newline.** `re.match(r'^\d+$', '11\n')` is True — `$` matches before a final newline. A third-party node with a `checkpoint_id` widget holding `"11\n"` gets that value returned, and `assets_from_reduction` writes it into `workflow_recipe_asset` permanently, in the tier the design calls safe to share. Lines 159-165 justify this regex by noting that "the two guards below (no newline, 255 bytes) sit AFTER this branch and would not catch it", and then it doesn't catch it either. The node contract cited, `str.isdigit()`, returns False for `"11\n"`.

**Length.** `"1" * 400` matches `^\d+$`, so a 400-byte value is filed past `_MAX_FILENAME_LENGTH` — the comment's own "a kilobyte" scenario. The new test at tests/test_workflow_identity.py:521 uses `"x" * 400`, which the character class rejects, so it passes without exercising the length dimension at all; substituting `"1" * 400` turns it red. Real shelf ids are a few digits, so a cap belongs here.

pixlstash/services/workflow_hash.py
```diff
@@ -251,15 +276,31 @@ def structural_widget_value(name: str, value: Any) -> Optional[str]:
...
+    digest of the empty string) into a stored document whose whole purpose is
+    to say which model went there. A shelf id that is not a shelf id names
+    nothing either; see :data:`SHELF_ID_VALUE_RE`.
     """
     if SEED_FIELD_RE.search(name) or name == "filename_prefix":
         return None
     if _OUTPUT_PATH_RE.match(name):
         return None
     if not isinstance(value, str):
         return None
```
lindkvis, line 321: A `checkpoint_id` serialised as a JSON number is nulled here, before the new branch, so #1416 stays unfixed for that spelling.

An API prompt written by a script rather than by ComfyUI's frontend carries `"inputs": {"checkpoint_id": 11}`. The node executes it fine — `_fetch_record` does `str(checkpoint_id or "").strip()` — but `structural_widget_value` returns `None` at this line, so the structural hash again does not move when the checkpoint does. Accepting an `int` for this name matches the node's own contract, which the new guard's comment invokes as its authority.

pixlstash/services/workflow_hash.py
```diff
@@ -251,15 +276,31 @@ def structural_widget_value(name: str, value: Any) -> Optional[str]:
...
+    if SHELF_ID_FIELD_RE.match(name):
+        return value if SHELF_ID_VALUE_RE.match(value) else None
     if SHA256_FIELD_RE.search(name):
```
lindkvis, line 322: This branch still returns any non-empty value verbatim, above the newline and 255-byte guards at line 308 — and this PR widens the set of names that reach it.

A widget named `expected_sha256`, `clip_sha256_2`, or anything matching `(^|_)sha256(_\d+)?$` holding prose, a newline or a kilobyte is lowercased, written to `workflow_recipe_asset` and rendered as an `asset:` reference in the kept-forever document.

`DIGEST_PREFIX_RE` (line 148) already encodes what such a value must look like, and `hub/workflows._model_ghost_names` already applies it: "A value that is not a digest or an A1111 short hash of one names no model." Gating this branch on `DIGEST_PREFIX_RE` is the general fix — it subsumes the `or None` added below and closes the same hole the PR chose to close for `checkpoint_id` only.

pixlstash/services/workflow_hash.py
```diff
@@ -251,15 +276,31 @@ def structural_widget_value(name: str, value: Any) -> Optional[str]:
...
-        return value.lower()
+        return value.lower() or None
```
lindkvis: `or None` is a third, value-level rule change, and it re-keys recipes that are already filed — but `HASH_VERSION` is still `"v1"`.

A graph on the hub carrying a blank `*_sha256` (a download node's `expected_sha256` left blank — named as a real case in `_model_ghost_names`'s own docstring — an unset `vae_sha256`, an unpicked `clip_sha256`) previously contributed `""` to the key and now contributes nothing. Re-filing the same graph produces a different `structural_hash`, so `workflow_recipe` gains a second row for one workflow and the Workflows view shows phantom variants — with both rows stamped `v1`, the field whose docstring says a change of rule must be "visible in the data rather than inferred".

The pinned-hash test at tests/test_workflow_identity.py:429 frames this PR as "two widget NAMES" and its `_graph()` carries no `*_sha256` widget, so nothing guards this dimension.

tests/test_workflow_identity.py
```diff
@@ -405,3 +406,121 @@ def test_the_pixlstash_picture_loader_and_a_noise_mask_are_picture_sources():
     assert workflow_type(_doc(graph)) == "img2img"
     graph["13"] = _node("SetLatentNoiseMask", samples=["12", 0])
     assert workflow_type(_doc(graph)) == "inpaint"
+
+
+# ── the PixlStash shelf loaders (#1416) ─────────────────────────────────────
+
+
+def _shelf_graph(checkpoint_id: str = "11", **kwargs) -> dict:
```
lindkvis: `**kwargs` forwards a `ckpt=` that the next line discards: `_shelf_graph(ckpt="other.safetensors")` builds `_graph(ckpt=...)` and then overwrites node `"1"` with the PixlStash loader. A later test asserting two `ckpt` values hash differently would fail confusingly, and one asserting they hash the same would pass for the wrong reason.

tests/test_workflow_identity.py
```diff
@@ -405,3 +406,121 @@ def test_the_pixlstash_picture_loader_and_a_noise_mask_are_picture_sources():
...
+    # node, not two. (The node still HAS the widget, and a nulled widget keeps
+    # its name by design, so this is about the value and not the shape.)
+    def clip_loader(**widgets):
+        return _graph(
+            extra={"60": _node("PixlStashCLIPLoader", type="sdxl", **widgets)}
+        )
+
+    blank = clip_loader(clip_sha256="aa" * 32, clip_sha256_2="")
+    assert {s.widget for s in slots(_doc(blank))} == {"ckpt_name", "clip_sha256"}
+    assert structural_hash(blank) == structural_hash(
```
lindkvis: This compares a value to itself and can never fail.

Line 508 defines `blank = clip_loader(clip_sha256="aa" * 32, clip_sha256_2="")`, and the right-hand side here is that same expression spelled out — so the assertion is `x == x`. It holds under every mutation of the rule it guards, including deleting the `or None` on workflow_hash.py:303.

CLAUDE.md: "Show every new assertion failing. Break what it guards, watch it go red, restore." This one cannot go red.

lindkvis (commented):
Review of the checkpoint-loader hash change. One finding is on a file this PR does not touch, so it has no line to hang on:

**The blank `*_sha256` rows already filed can never be purged.** `hub/workflows.py:323` routes a row with `widget_name = 'vae_sha256'` and `normalized_filename = ''` — exactly the junk `workflow_hash.py:287`'s docstring identifies as the problem — to the digest branch, where `DIGEST_PREFIX_RE.match('')` is False, so it is never a ghost. "Forget model names not on the shelf" leaves them, and the matching `asset:e3b0c442…` reference stays in the stored document forever. The PR stops new ones and offers no way to clear the existing ones.

The two I'd want fixed before merge are the `LORA_DIGEST_FIELDS` half of the widening (line 144) and the self-comparing assertion at tests/test_workflow_identity.py:510.

frontend/src/utils/workflowShelf.js
```diff
@@ -11,12 +11,23 @@
 // it loads and how big it is. That is the "not named" state from the design's
 // `States.dc.html`, and it is the ordinary case rather than the exception.
 
-/** Widget names whose value is the checkpoint or diffusion model. */
+/**
+ * Widget names whose value is the checkpoint or diffusion model.
+ *
+ * The mirror of `_CHECKPOINT_WIDGETS` in
```
lindkvis: The new docstring calls `BASE_WIDGETS` "the mirror of `_CHECKPOINT_WIDGETS`" and says "the two must agree" — they don't, and nothing enforces it.

`BASE_WIDGETS` holds `diffusion_model` and `model_path`; `_CHECKPOINT_WIDGETS` (workflow_identity.py:85) holds neither. A workflow whose base model sits on a `diffusion_model` widget reads as "base" in the Models column and falls through to the "other models" chip in `differs_by`.

This PR's bug *was* this drift ("Out of step, a workflow that has a checkpoint reported none"), and the fix records the invariant as a comment with no guardrail, so the next widget added to one side drifts the same way `checkpoint_id` just did. `git grep -l 'BASE_WIDGETS\|_CHECKPOINT_WIDGETS'` returns only the two sources and the JS test.

pixlstash/services/workflow_hash.py
lindkvis: Fixed in 928c64a1. You are right that only half the pair moved, and the consequence you name — a character LoRA swap forking the card — is exactly what marks exist to stop.

Not by widening `LORA_DIGEST_FIELDS` itself, though: `detect_lora_targets` takes **one** digest slot per node on purpose ("the pack has called the widget both things, and a node carrying two names is still one adapter to load"), so turning the tuple into a pattern would change what a swap writes. The two are different questions. Added `LORA_DIGEST_FIELD_RE = ^(adapter|lora)_sha256(_\d+)?$` beside the tuple for the classification question, and `_is_lora_widget` matches on it. The tuple and `detect_lora_targets` are untouched.

`test_a_numbered_digest_slot_is_a_lora_slot_and_takes_a_mark` covers it: both slots come back `is_lora=True`, the card does not fork on a swap, and the variant hash still moves. Red when `_is_lora_widget` drops the regex.

pixlstash/services/workflow_hash.py
lindkvis: Fixed in 928c64a1, taking your version almost verbatim — two module constants and two `re` calls to express a string equality and a weaker `isdigit()` was me reaching for the pattern the file next door happened to use.

Now `SHELF_ID_FIELD = "checkpoint_id"` (still named, because `structural_widget_value` and the docstring both refer to it) and inline `value.isdigit() and len(value) <= _MAX_SHELF_ID_LENGTH`. 12 is the cap: SQLite AUTOINCREMENT tops out at 19 digits and a real shelf runs to four, so it refuses a kilobyte without ever refusing a shelf.

pixlstash/services/workflow_hash.py
lindkvis: Both holes confirmed and fixed in 928c64a1. I checked before believing it:

```
re.match(r"^\d+$", "11\n")   -> True
"11\n".isdigit()             -> False
re.match(r"^\d+$", "1"*400)  -> True
```

So the comment claimed a guard the regex did not provide, on both dimensions. `isdigit()` plus the 12-character cap closes both, and you are right that `"x" * 400` in the test never exercised length at all — the character class rejected it before length mattered. The test now iterates `"11\n"` and `"1" * 400` alongside it, and goes red if the rule is put back to `^\d+$`.

pixlstash/services/workflow_hash.py
lindkvis: Fixed in 928c64a1. Real gap — the node does `str(checkpoint_id or "").strip()`, so it runs a graph the hash was nulling, and #1416 stayed open for that spelling.

The shelf-id branch now sits above the `isinstance(value, str)` return and coerces an `int`, excluding `bool` (`isinstance(True, int)` is True in Python, and `True` is not a shelf id). `test_a_checkpoint_id_written_as_a_number_names_the_same_model` asserts the text and number spellings produce the same structural hash and the same card key, and that `True` names nothing.

pixlstash/services/workflow_hash.py
lindkvis: Taken, in 928c64a1 — this is the better fix and it replaces the `or None` rather than sitting beside it.

```python
lowered = value.lower()
return lowered if DIGEST_PREFIX_RE.match(lowered) else None
```

Lowercased first so an uppercase digest still passes. A1111 short hashes are unaffected: `DIGEST_PREFIX_RE` already admits the 10- and 12-digit forms `_short_hash` writes. And you are right that it is the same hole I closed for `checkpoint_id` only — having written the narrow guard I should have asked why the branch above it needed none.

Now both branches judge the value, which is also what makes the docstring true: `_model_ghost_names` applies this exact rule, so the hasher and the ghost finder can no longer disagree about what names a model.

pixlstash/services/workflow_hash.py
lindkvis: You have convinced me — bumped to `v2` in 928c64a1.

My reasoning was "unreleased, so v1 never shipped, so amend". That was answering the wrong question. The column is not about releases; its own docstring says a later rule is "a new value here rather than a silent reinterpretation of rows written under the old one", and this is a later rule whatever the release state. Leaving two rule sets stamped `v1` is precisely the silent reinterpretation it forbids, and `WHERE hash_version = v1` is now the only way to find the old rows, since nothing re-keys them in place.

It does not fix the phantom `workflow_recipe` rows — a re-key cannot, without the hub sweep described in the PR body — but it makes them findable, which is what the column is for. `HASH_VERSION`s comment records what changed in v2. `test_the_stored_row_records_which_rule_keyed_it` pins the new literal and keeps its rationale.

tests/test_workflow_identity.py
lindkvis: Fixed in 928c64a1 — `**kwargs` removed. Nothing passed it, and as you say the trap is a future test: `_shelf_graph(ckpt=...)` would have built the argument into `_graph` and then thrown it away with node `"1"`, so an equality assertion would have passed for entirely the wrong reason. The signature is now `_shelf_graph(checkpoint_id="11")` and takes nothing else.

tests/test_workflow_identity.py
lindkvis: You are right, and it is worse than a weak assertion — I ran a mutation sweep over this file and this line sailed through it, because the two assertions above it in the same test were carrying the failure. `x == x` cannot go red.

Deleted in 928c64a1. The set assertion above it is the load-bearing one (a blank `clip_sha256_2` must not appear among the slots), and it is red when `DIGEST_PREFIX_RE` is removed from the digest branch. Added beside it an assertion that can also fail in the other direction — a blank second encoder must **not** key the same as a filled one — plus a direct check that `asset:` + the digest of the empty string is absent from the document.

frontend/src/utils/workflowShelf.js
lindkvis: Fixed in 928c64a1, and you are right that a comment asserting an invariant is how the bug happened in the first place — I wrote "the two must agree" while they did not, which is the same move I was fixing.

Three things:

1. **They now agree.** `diffusion_model` and `model_path` were client-only; both are base-model widgets, and treating them as such means `differs_by` says `other checkpoint` for them instead of falling back to `other models`. `_CHECKPOINT_WIDGETS` renamed to `CHECKPOINT_WIDGETS` since a test outside the module imports it.
2. **A guardrail, not a comment.** `test_base_model_widgets_agree_across_the_stack` parses the `BASE_WIDGETS` literal out of `workflowShelf.js` and asserts set equality, naming client-only and server-only widgets in the failure. Same shape as `test_frontend_import_extensions_match_the_staging_allowlist`, which exists for the same reason.
3. Both comments now point at the guardrail rather than asking the next reader to remember.

Verified red by deleting `model_path` from the JS set.

lindkvis:
Answering the finding in the review body, which has no line to hang on — **the blank `*_sha256` rows already filed can never be purged.**

Confirmed and fixed in 928c64a1. You traced it exactly: `_model_ghost_names` routes any `*_sha256` widget to the digest branch, `DIGEST_PREFIX_RE.match("")` is False, so a blank is never a ghost and "forget model names not on the shelf" cannot reach it. Stopping new ones while leaving the existing ones unreachable is half a fix.

The hub has the right mechanism for this already — `CURRENT_DATA_VERSION`, the one-shot backfill counter in `PRAGMA user_version`, whose whole rationale is a step that must run exactly once over stored data rather than on every open. Bumped to 2, with `_drop_blank_recipe_assets`:

```sql
DELETE FROM workflow_recipe_asset WHERE normalized_filename = ''
```

Deleting one loses nothing — an empty name identifies nobody and resolves to no shelf row — and no hash moves and no stored document is rewritten, which is the same shape forgetting a name already has. The `asset:e3b0c442…` reference stays in the document for exactly that reason: that is what a forgotten name looks like, and rewriting stored graphs is the thing the design refuses to do.

`TestBlankRecipeAssetBackfill` covers both directions: blanks go, a named asset and a real digest stay, and the step does not re-run (a blank filed afterwards survives — the rule in `workflow_hash` is what stops new ones, not this). Red when the backfill call is removed.

**Two things I have not fixed, both named in the PR body and unchanged by this round:** the orphaned `workflow_recipe` rows a re-key leaves (phantom variants), and `workflow_recipe_instance.structural_hash` going stale. Neither is reachable from an alembic migration — the hub is a separate database — and both belong with the hub sweep that a proper re-key needs. The `v2` stamp is now what makes those rows findable when that lands.

PR #1425, opened 2026-09-17
Workflow cards: hub tables, backfill and filing hook (#1394)

lindkvis:
B2 of v1.12 Workflows & Recipes. A `workflow_recipe` row is a **variant**; what a person means by "a workflow" is a **card**. This adds the hub tables that key cards by content, derives them from the hub's own stored documents, and hooks the derivation where graphs are already filed.

## What

- **New hub tables, amended into schema v2** (`CREATE TABLE IF NOT EXISTS`, `CURRENT_SCHEMA_VERSION` stays 2): `workflow_variant`, `workflow_topology_core`, `workflow_slot_mark`, `workflow_file`, plus the owner-decision tables the later steps fill (`workflow_attr`, `workflow_default_override`, `workflow_key_pins`, `workflow_key_picture_input`, `workflow_cover`, `workflow_stack`, `workflow_stack_member`, `workflow_unstacked`).
- **`hub/workflow_cards.py`** derives a card from a stored document only: slots, the frozen LoRA mark per slot, the card key, and a per-topology cache of core hash / type / slot labels. No card row holds a readable model name.
- **Filing hook in `record_reduction`**, so import, the watched inbox and the ComfyUI extraction are all covered by one call; the import route records the file→card row and deleting a stored workflow drops it.
- **Backfill as a Task Manager pass** (`WorkflowCardBackfillFinder` / `Task`), hub-only: no picture is read, so libraries that are not attached get their cards too. It logs the grouping (cards, stacks, one-offs) for the owner gate.

## Why some things are not what the issue's wording suggests

- **No `CURRENT_DATA_VERSION` bump.** The finder's queue *is* "variants with no current card", so the pass is once-only by construction, survives an interrupted first run, and re-fills itself on a `WORKFLOW_KEY_VERSION` / `CORE_VERSION` bump. A data-version step would be skipped on the one hub that needed it most.
- **Automatic stacks are a query, not rows.** Cards sharing a `core_hash` are one stack; the stack tables are for the owner's own decisions. `STRIP_LORAS_FOR_STACKS` is the single line the owner gate may flip.

Fixed: the topology cache was rewritten once per variant; the identity rows carried a `computed_at` that made "runs twice with identical rows" untestable (dropped, and the test now wipes the rows and re-derives rather than comparing a table to itself); the filing hook ran a four-table join once per *picture* (now only for a recipe the hub had not seen); the new worker had no branch in the progress snapshot, so it would have rendered "N / N, 0 remaining" with the library's picture count; two unused indexes; `workflow_file.name` → `workflow_name`; test fixture names now carry the `test-` marker.

Disagreed, deliberately:

- **Eight tables have no writer yet.** The issue asks for them, their shape is decided here, and a hub table is append-only — adding each with its writer means one more guarded amendment of v2 per step.
- **The card key is order-dependent through the frozen mark.** That is the design the issue states ("frozen the first time the slot is seen"): re-guessing per filing would silently re-key cards the owner has named, pinned and stacked. It is now stated in the schema, the module and the doc instead of being discoverable, and the two tests that file the same graphs in opposite order say so in as many words.
- **A mark outlives `forget_asset_names`.** The readable name goes; the decision that keys the card stays, so `structural` still says the forgotten file looked like a speed LoRA. That is a classification rather than the name, and deleting it would re-key the owner's card. Asserted in the test rather than left implicit; the correction is a flip, a later step.
- The diff is over the ~600-line guidance, mostly DDL and its comments for twelve tables the issue enumerates.

## Not done here

The owner gate — running this over the week-2 dogfood copy and reporting the grouping against the measured 617 variants over 192 topologies — needs the real library. The backfill logs exactly that line on every batch.


lindkvis:
Review at head `6e8b5429`. Read the full diff plus the unchanged collaborators it leans on (`hub/db.py`, the `_apply_v2` machinery, `services/workflow_identity.py`, `base_task_finder.py`, `work_planner.py`).

Verified clean, so not repeated below: the schema amend works (`apply_migrations` re-runs `_apply_v2` unconditionally for `version >= 2`, and the new statements sit before the index list, so an existing v2 hub does pick them up — no `CURRENT_SCHEMA_VERSION` bump owed); `if new_recipe:` is once per recipe, not per picture; import and delete paths are exception-isolated; the finder matches the `MissingCheckpointHashFinder` claiming/deferral contract; ruff clean on all 13 changed Python files. All four of the "Answering the adversarial pass" disagreements hold up as argued.

Three I'd want before merge.

**F1 — `STRIP_LORAS_FOR_STACKS` is not folded into `CORE_VERSION`.** `pixlstash/hub/workflow_cards.py:58`, `:142`, `:235`. `core_hash` has no version inside its digest; `CORE_VERSION` is only a stamp column. Flipping the flag — the one line the PR offers the owner gate — changes every core hash but leaves `core_version = "v1"`, so the finder doesn't re-queue and the hub mixes two rules' stacks. `WORKFLOW_KEY_VERSION` is inside its digest and doesn't have this problem; the core lane is the asymmetric one.

**F2 — `card_grouping` ignores `core_version`.** `workflow_cards.py:348`, unlike every other query in the module. The number the owner gate reads is computed over stale-version rows mid-re-derivation. Secondary: variants with no cache row collapse into one `None` bucket and report as a single giant stack.

**F3 — the version-bump re-derivation has no test**, and it's the PR's stated replacement for a `CURRENT_DATA_VERSION` step. Four mutations stay green: deleting either `*_version` join predicate, dropping the `c.topology_hash IS NULL` half of `_VARIANT_PENDING`, and removing `and row["core_cached"] is not None` from `record_identity`'s early return — whose own comment explains why it must be there. Two small tests close all four.

Seven cheaper ones that can ride along:

- **F4** `hub/workflows.py:228` catches `(WorkflowGraphError, sqlite3.Error)` but the comment above it promises "never fatal to the filing". `record_identity` indexes into the parsed document, so a malformed one raises `KeyError`/`TypeError` and fails the caller's ingest. `routes/comfyui.py:325` already uses `except Exception` for exactly this reason, so the import path is protected and the extraction path isn't.
- **F5** `workflow_card_backfill_task.py:55` catches only `WorkflowGraphError` per item, and `..._finder.py:85` defers every hash in a failed batch for the process lifetime. With a 5 s busy timeout on a multi-process hub, a transient `database is locked` permanently retires 50 variants until restart. Inherited from `MissingCheckpointHashFinder`, but a card is cheap to retry and a 24 GB hash isn't.
- **F6** `workflow_card_backfill_task.py:72` runs a full `GROUP BY` scan and logs an INFO line per 50-variant batch, for a number only the owner gate reads. `on_all_tasks_complete()` is where a once-per-drain report belongs.
- **F7** `workflow_cards.py:188` — a third order-dependence, undocumented: `slots()` sorts by `(label, asset)`, so for WL twins sharing a label the frozen mark comes from the lexicographically smaller `asset:<digest>`. Deterministic, but decided by something meaningless. One sentence in `_freeze_marks`.
- **F8** `test_the_card_tables_land_in_the_hub` asserts on a fresh hub; the claim that matters (existing v2 hub gains them on next open) is pinned only for `SHELF_TABLES` elsewhere. Covered transitively via the shared loop, but an edit moving the card statements into the versioned list would stay green.
- **F9** `tests/test_workflow_cards.py:324` local-imports `TaskCancelledError` inside the test body; no cycle or heaviness justifies it.
- **F10** `routes/comfyui.py:300` — `record_file` now runs between filing and return, so a sqlite error in the card write makes the import response report `topology_hash: null` for a graph that was filed. Cosmetic, but the optional half shouldn't retract the essential half's answer.

One test could pass for the wrong reason: `test_the_backfill_keys_what_was_filed_before_it_and_runs_twice_the_same`. The `(3,3) → (3,0)` transition rests entirely on `v.structural_hash IS NULL` (see F3), and `wipe` drops only `workflow_variant` and `workflow_topology_core`, so `workflow_slot_mark` is carried over rather than re-derived and compared to itself. The marks are covered by `test_a_mark_is_frozen_on_first_sight`; it's the docstring's claim that's wider than what this test proves. Also `workflow_file` is in `CARD_TABLES` but has no row here, and `record_file` writes `filed_at` under `INSERT OR REPLACE`, so the byte-identical claim would not survive a re-filed file.

lindkvis:
All ten addressed in `7198453c`, plus the test note. Two are answered rather than changed.

**F1 — flag folded into the stamp.** `CORE_RULE_VERSION = f"{CORE_VERSION}-loras-{'stripped'|'kept'}"` is what `workflow_topology_core.core_version` now holds and what every query joins on, so flipping `STRIP_LORAS_FOR_STACKS` re-queues every topology instead of leaving two rules' stacks in one hub. `test_flipping_the_lora_strip_is_a_new_core_rule` pins it; with the stamp back to bare `CORE_VERSION` it goes red.

**F2 — `card_grouping` joins on `core_version`** and no longer buckets uncached rows together: a card whose cache row is missing or superseded is counted in `cards` and reported as a new `ungrouped` figure, which is in the log line. Asserted in `test_a_superseded_core_rule_re_queues_the_topologies_it_cached` (`ungrouped` 1 → 0 across the pass).

**F3 — two tests, four mutations closed.** `test_a_superseded_key_rule_re_queues_the_variants_it_keyed` and `test_a_superseded_core_rule_re_queues_the_topologies_it_cached` stamp an old version on the stored rows, which is what a bump looks like to every query here. Verified red for: either `*_version` join predicate neutered, the `c.topology_hash IS NULL` half of `_VARIANT_PENDING` dropped, and `and row["core_cached"] is not None` removed from the early return.

**F4** — the hook is now `except Exception` with the type logged, same reasoning as `_file_in_hub`. **F5** — `sqlite3.OperationalError` leaves its batch eligible; everything else still retires it. The test is renamed `test_only_an_error_that_will_not_pass_retires_a_batch` and covers cancelled / locked / other. **F6** — the report moved to `on_all_tasks_complete`, once per drain, and the test now goes through that callback rather than calling `card_grouping` directly. **F7** — the twin-label tie-break is documented in `_freeze_marks`. **F9** — import at the top. **F10** — the card write has its own handler (`_card_the_file`), so a failed card cannot retract the `topology_hash` of a graph that was filed.

**F8 — already pinned, directly.** The card tables are in `WORKFLOW_WIPE_ORDER`, which `test_an_existing_v2_hub_gains_the_tables_on_its_next_open` drops and re-opens against. I ran your edit as a mutation — card DDL moved out of `_V2_WORKFLOW_TABLES` into the versioned step-2 list — and that test fails, so the existing-hub claim is not resting on the fresh-hub one.

**The test that could pass for the wrong reason.** `workflow_slot_mark` now goes in the second wipe, so the marks are re-derived rather than read back — which meant dropping the last two wall-clock columns (`marked_at`, `filed_at`, neither read by anything), and every card table is now timestamp-free and byte-stable. `workflow_file` stays in the comparison and is empty there; the docstring says so and points at `test_replacing_a_file_moves_it_to_the_new_card`.

Suite green (`test_workflow_cards`, `_io`, `_library`, `workflows_api`, `workers_api`, `work_planner`, `architecture_guardrails`, `ci_shards`, `comfyui_workflow_extraction`, `hub_bootstrap`: 715 passed), ruff clean.

Issue #1406, opened 2026-09-17
Workflows & Recipes F4: overlay Info | Recipe tabs


lindkvis:
Part of **v1.12 Workflows & Recipes**. The design is round 2, alternative C, "Stack Opens with Grid or List" (the design bundle `design/1.12-workflows-recipes/` in the business repo). Build what the design decides and use shipped components for the chrome. Proceed on the plan's recommended answers to the open owner decisions unless the owner has recorded otherwise in that bundle's `DECISIONS.md`.

Rules for every step: one PR against `develop`, about 600 lines of diff at most, shippable alone. Until F1b swaps the route, the shipped Workflows shelf keeps working. In code and docs, `workflow_recipe` / `structural_hash` is a **variant** and the user's Recipe is a **saved recipe**. The hub tables are append-only and are never renamed.

## Depends on

| Dependency | Hard? | Check it is in `origin/develop` |
|---|---|---|
| #1390 (B1 workflow identity): every step comes after it | yes | `git cat-file -e origin/develop:pixlstash/services/workflow_identity.py` |
| #1384 overlay on AppInspector | yes | `gh pr view 1384 --json state,mergeCommit  # MERGED, then merge-base --is-ancestor` |
| #1395 (B3) | yes | `git grep -q '/workflows/cards' origin/develop -- pixlstash/authz/registry.py` |
| #1397 (B5) | yes | `git grep -q workflow_key origin/develop -- pixlstash/routes/comfyui.py` |

## Before you start (mandatory)

1. **Start from the latest `origin/develop`:** `git fetch origin && git checkout -b <branch> origin/develop`. If you were handed a worktree that already exists, `git fetch origin` and rebase it onto `origin/develop` before touching anything. Do not build on a stale checkout.
2. **Check that every dependency below is actually in `origin/develop`,** not just closed or badged MERGED. Merges to develop do not auto-close issues. For each dependency, find its PR (`gh pr list --state merged --search "<issue or PR number>"`), confirm `git merge-base --is-ancestor <merge commit> origin/develop`, and run the check listed.
3. **If any hard dependency is missing, stop.** Comment on this issue saying which one, and do not start. Never base a PR on another PR's branch.

## What

- `ImageOverlay.vue` sidebar becomes **Info | Recipe** tabs on `AppInspector`. Info is today's content, unchanged.
- `views/OverlayRecipePanel.vue`, read-only:
  - a "Matches your saved recipe X" banner (once B6 exists; hide until then);
[...]

Issue #1397, opened 2026-09-17
Workflows & Recipes B5: extended picture recipe read and workflow picture filters


lindkvis:
Part of **v1.12 Workflows & Recipes**. The design is round 2, alternative C, "Stack Opens with Grid or List" (the design bundle `design/1.12-workflows-recipes/` in the business repo). Build what the design decides and use shipped components for the chrome. Proceed on the plan's recommended answers to the open owner decisions unless the owner has recorded otherwise in that bundle's `DECISIONS.md`.

Rules for every step: one PR against `develop`, about 600 lines of diff at most, shippable alone. Until F1b swaps the route, the shipped Workflows shelf keeps working. In code and docs, `workflow_recipe` / `structural_hash` is a **variant** and the user's Recipe is a **saved recipe**. The hub tables are append-only and are never renamed.

## Depends on

| Dependency | Hard? | Check it is in `origin/develop` |
|---|---|---|
| #1390 (B1 workflow identity): every step comes after it | yes | `git cat-file -e origin/develop:pixlstash/services/workflow_identity.py` |
| #1394 (B2) | yes | `git grep -q 'CREATE TABLE IF NOT EXISTS workflow_file' origin/develop -- pixlstash/hub/schema.py` |

## Before you start (mandatory)

1. **Start from the latest `origin/develop`:** `git fetch origin && git checkout -b <branch> origin/develop`. If you were handed a worktree that already exists, `git fetch origin` and rebase it onto `origin/develop` before touching anything. Do not build on a stale checkout.
2. **Check that every dependency below is actually in `origin/develop`,** not just closed or badged MERGED. Merges to develop do not auto-close issues. For each dependency, find its PR (`gh pr list --state merged --search "<issue or PR number>"`), confirm `git merge-base --is-ancestor <merge commit> origin/develop`, and run the check listed.
3. **If any hard dependency is missing, stop.** Comment on this issue saying which one, and do not start. Never base a PR on another PR's branch.

## What

- Extend `GET /comfyui/pictures/{id}/recipe` (stays `PICTURE_SCOPED`, no inline check) with: negative prompt, **LoRA strengths** (which `detect_lora_targets` does not return today), settings, `workflow_key`, and an **A1111 branch** through `reduce_a1111`.
- The picture list gains `workflow_key` / `workflow_stack` filters, resolved to a structural-hash `IN` following the `comfyui_loras` filter pattern.
[...]

Issue #1407, opened 2026-09-17
Workflows & Recipes F5: Run popup and Make more like these


lindkvis:
Part of **v1.12 Workflows & Recipes**. The design is round 2, alternative C, "Stack Opens with Grid or List" (the design bundle `design/1.12-workflows-recipes/` in the business repo). Build what the design decides and use shipped components for the chrome. Proceed on the plan's recommended answers to the open owner decisions unless the owner has recorded otherwise in that bundle's `DECISIONS.md`.

Rules for every step: one PR against `develop`, about 600 lines of diff at most, shippable alone. Until F1b swaps the route, the shipped Workflows shelf keeps working. In code and docs, `workflow_recipe` / `structural_hash` is a **variant** and the user's Recipe is a **saved recipe**. The hub tables are append-only and are never renamed.

## Depends on

| Dependency | Hard? | Check it is in `origin/develop` |
|---|---|---|
| #1390 (B1 workflow identity): every step comes after it | yes | `git cat-file -e origin/develop:pixlstash/services/workflow_identity.py` |
| #1399 (B7) | yes | `git grep -q '/workflows/run/preflight' origin/develop -- pixlstash/authz/registry.py` |
| #1403 (F3) | yes | `git grep -q 'Node names for ComfyUI users' origin/develop -- frontend/src` |
| #1406 (F4) | yes | `git cat-file -e origin/develop:frontend/src/components/views/OverlayRecipePanel.vue` |

## Before you start (mandatory)

1. **Start from the latest `origin/develop`:** `git fetch origin && git checkout -b <branch> origin/develop`. If you were handed a worktree that already exists, `git fetch origin` and rebase it onto `origin/develop` before touching anything. Do not build on a stale checkout.
2. **Check that every dependency below is actually in `origin/develop`,** not just closed or badged MERGED. Merges to develop do not auto-close issues. For each dependency, find its PR (`gh pr list --state merged --search "<issue or PR number>"`), confirm `git merge-base --is-ancestor <merge commit> origin/develop`, and run the check listed.
3. **If any hard dependency is missing, stop.** Comment on this issue saying which one, and do not start. Never base a PR on another PR's branch.

## What

- **`io/RunDialog.vue`** (`AppDialog` lg; shipped chrome, 16px gutter), one popup for every source:
[...]

PR #1426, opened 2026-09-17
Recipes: recipe section in the overlay

### LINKED ISSUE #1313: Recipes: recipe section in the overlay
A Recipe section in the overlay sidebar, beside Description, Faces and Tags:

- Models, with verified badges and strengths, linking to the shelf.
- Prompt, settings, and inputs (the resolution lock).
- Generate variants….
- The workflow, with Copy and Download, linking to the Workflows view.

Design: the recipe panel in the design system project.

Depends on: #1311, Pikselkroken/ComfyUI-PixlStash#23, #1307

lindkvis:
Closes #1313.

A **Recipe** section in the lightbox sidebar, between Tags and Metadata: the
models the picture used with the strength each LoRA was loaded at and a tick on
any the recipe named by digest, the prompt, the sampler settings, the resolution
lock, Generate variants…, and the workflow with Copy, Download and Open.

## How

`GET /comfyui/pictures/{id}/workflow` is the endpoint the overlay already calls
once per picture, so it carries the new fields rather than a second route
existing: `model_slots`, `settings`, `inputs`, `topology_hash`, from a new
`pixlstash/services/picture_recipe_service.py`. The models are read through
`workflow_hash.reduce_api_graph` — the same reduction the hub files as
`workflow_recipe_asset` — so the shelf match is a lookup rather than a second
normalisation, and the verified tier is `fetch_picture_counts`'s read backwards.
Settings come through `workflow_parameters.describe_parameters`, so the names
agree with the run panel's.

Two fields vary by credential. Shelf rows (`model_id`/`verified`) and the
resolution lock go to a fully-unscoped owner only; a scoped token still gets the
filename and the strength, which are in the graph the route serves it anyway.
New `AuthService.is_unscoped_owner_request` is the predicate, and
`require_unscoped_owner` now delegates to it so the two cannot drift.

The ComfyUI tab was **moved** out of `OverlayMetadataPanel.vue`, not copied —
that file loses 454 lines and its tab band, leaving one plain list.

`?model=<id>` and `?topology=<hash>` land the two shelves on the right row.

The pre-push review found no secrets and nothing private, and demonstrated one
real leak with a live probe: a picture-scoped token was served
`inputs[].input_picture_id` and `pixel_sha` for a picture the gate refuses it on
every other route. Fixed — the lock is owner-only and the route no longer reads
it for anyone else — with a test that fails when the gate is removed, carrying
both a positive control and a 403 control on the referenced picture.

Also from that pass: the verified badge survived a `v-if` inversion (now
asserted per chip); the duplicate `SELECT id, filename FROM model` is gone
(`recipe_asset_index` returns the map it already built); `inputs` sort by node id
numerically rather than as text; a setting two nodes disagree about is qualified
by its node instead of drawing two undifferentiated "Steps" rows; the three new
response models dropped `extra="allow"` so they are a real disclosure boundary on
a credential-varying route; off-grid spacing is on the token scale; a thumbnail
that will not load falls back to the designed "gone" tile; five route tests
became three to spend two fewer `Server` boots in the gate.

Two of its findings I answered differently rather than obeying:

- **`ModelShelf` not widening filters like `WorkflowShelf` does.** It was right
  that doing nothing is worse than either option, wrong that the two should
  match: `WorkflowShelf` has one Show axis that is safe to widen, the shelf has
  several plus collapsed groups and closed stacks, and clearing a reader's
  filters behind their back is not an improvement. It now says the model is
  hidden and offers *Show everything*.
- **`extra="allow"` on `ComfyUIPictureWorkflowResponse` itself.** Left alone —
  it predates this change and tightening it would alter what the existing fields
  serve.

## Not done here

- The design-system project has no recipe-panel spec, though the issue names
  one; this is built on the shipped overlay-panel idiom and the tokens.
- #1406 (v1.12 F4) will restructure this sidebar into Info | Recipe tabs and add
  the A1111 branch. This is the section #1313 asked for, not that.
- A1111/Forge pictures get no Recipe section: the route 404s without a ComfyUI
  workflow. That is #1397's job.

## Checks

20 + 7 backend unit tests, 3 route tests, 12 component tests. Every new
assertion was mutation-checked — 18 mutations, each went red, each restored
green. Frontend 4410 passed, lint clean; the touched backend suites 598 passed;
`ruff` and `render_backend_architecture.py --check` clean. A pathological
200-node graph costs ~11 ms.


lindkvis:
`backend-windows shard 4` is red and no test failed in it.

The shard reports `190 passed, 542 deselected`, then prints `Exiting with
status 0` and dies with a Windows access violation inside `os._exit` at
`tests/process_teardown.py:168` — the fast-exit path whose own comment reads
"win32: native exit-time destructors crash intermittently". The shell sees 139.

Same signature, same line, same `Exiting with status 0`, on **develop** in run
35274092885 (`backend-windows shard 4`, `154 passed`), with none of this branch
in it. `soak/windows-segfault` (#1420's branch) is already measuring this exact
"intermittent 139" and is not settled.

I ran shard 4's own file list here with `--ci-shard 4/4`: the same 190 tests
pass, exit 0. Nothing in this diff writes to the hub or the vault — the only
statement it adds is a `select`.

So I have not changed anything for it. Re-running the job should clear it.

lindkvis:
Re-ran that one job, no code changed: `backend-windows shard 4` is green and all
17 checks pass.

One detail for whoever picks the soak back up: of the three route tests this
branch adds, the time-balanced sharder put exactly one in shard 4, so the diff
did add an `env` build to that shard. That is the only way it touches this at
all — the crash is still in `os._exit`, and develop hit it with none of these
tests in the run.

lindkvis:
Code review(xhigh · 15 findings)
  ⎿  frontend/src/components/views/OverlayRecipePanel.vue
       ●  196 [correctness]    The `unloadable` Set persists across lightbox navigation and its keys collide between
                               pictures, so one broken thumbnail permanently marks that node/position slot "gone" for
                               every later picture.
       ●  299 [simplification] `downloadWorkflow` is a seventh hand-rolled copy of the same blob-download dance.
     pixlstash/services/picture_recipe_service.py 
       ●  272 [correctness]    The node qualifier added to a repeated setting label does not disambiguate: `node_title`
                               falls back to `class_type`, so two untitled samplers produce two identical labels.
       ●  159 [correctness]    `_names_a_model` misses suffixed digest widgets, so the second text encoder of a
                               Flux/SD3/HiDream picture is dropped from the model list.
       ●  133 [correctness]    A malformed API prompt drops the model list entirely instead of falling back to the names
                               the route already has in hand.
       ●  199 [efficiency]     `recipe_asset_index(hub)` — three full scans of `model` and `model_file` — moves from a
                               shelf-wide, once-per-operation index onto the lightbox's per-picture path.
       ●  159 [correctness]    A blank `*_sha256` widget produces a nameless model chip, because `_names_a_model`
                               accepts any value on a digest-named widget without checking it looks like a digest.
       ●  283 [simplification] `_node_order` is a near-verbatim copy of `workflow_parameters._node_order`, and the
                               copy's `isdigit()` can raise the ValueError its own docstring promises not to.
       ●  120 [correctness]    Model names in the Recipe chips are now lowercased and stripped of their subfolder, where
                               the old ComfyUI tab printed the graph's raw name.
       ●  102 [simplification] `describe_recipe` re-checks `owner` for `inputs` although the route already passes `[]`
                               when the caller is not the owner, so the branch is dead in the only call site.
     frontend/src/components/views/ModelShelf.vue
       ● 3131 [correctness]    `?model=<id>` silently does nothing when the model's Show block is unticked, because
                               `store.rows` holds only the blocks that were fetched, so `known` is undefined and the
                               function returns with no selection and no notice.
       ● 3135 [correctness]    The "Show everything" notice action clears the filters but never selects or focuses the
                               model, because `honouredModel` was already set before the notice was pushed.
       ● 3139 [correctness]    The notice offers `store.resetFilters()` for two of the three hiding cases it names, and
                               resetting filters does not open a collapsed group or a closed stack.
     pixlstash/routes/comfyui.py
       ● 2793 [correctness]    The route's OpenAPI description promises a scoped token "the filename … which it can
                               already read out of the `workflow`", which is false for digest-addressed loaders; it gets
                               a raw 64-hex string as the model name.
     frontend/src/components/views/ImageGrid.vue
       ● 2026 [correctness]    `openRemixDialog`'s fallback can remix the wrong picture when the lightbox's picture is
                               not in `allGridImages`.

● 15 findings on #1426, reported above. The shape of it:

  Likely to bite a real user
  1. OverlayRecipePanel.vue:196 — the unloadable Set survives lightbox navigation and is keyed by graph node id, so one 404
     poisons that slot for every later picture in the session.
  2. ModelShelf.vue:3131 / :3135 — the new ?model=<id> deep link does nothing when the model's block is unticked, and the
     "Show everything" escape hatch widens the filters but never actually selects the row (WorkflowShelf does it in the right
     order).
  3. picture_recipe_service.py:159 — SHA256_FIELD_RE is end-anchored, so clip_sha256_2 is invisible: dual-encoder
     Flux/SD3/HiDream pictures list one of two text encoders.
 
  Regressions vs the old ComfyUI tab — a WorkflowGraphError now drops the whole Models group instead of falling back to
  workflow_info["models"] (:133), and chips render the normalised name, so SDXL/Base_1.0.safetensors shows as
  base_1.0.safetensors (:120).

  Perf — recipe_asset_index(hub) (three full table scans) moved from once-per-shelf-operation onto the per-picture workflow
  route, so arrow-keying the lightbox rescans model + model_file per keypress.

  Want me to fix a subset? The top three plus the _sha256_2 regex are the ones I'd take before merge.

✻ Baked for 16m 21s · done 1:32 AM

※ recap: Reviewed PR #1426 and reported 15 findings, mostly recipe-panel and model-shelf bugs. Next: tell me which to fix, or
  I'll start with the top three plus the clip_sha256_2 regex. (disable recaps in /config)

lindkvis:
Rebuilt onto the v1.12 design, which I had not read. It is the Workflows &
Recipes round-2 artifact plus `plans/pixlstash-workflows-recipes-implementation-plan.md`
in the business repo — I had looked in the two Claude Design projects and the
`1.12-recipe-panel` bundle, whose decision ("Direction A, sidebar-native") is
what I built and is superseded.

**The sidebar is now `AppInspector`'s own tab band, Info | Recipe**, not a
Recipe section under Tags. The Recipe tab is drawn in the design's order —
Workflow + Open, Prompt, Models as chips with strengths, then Settings, Seed and
Negative in one kv grid, action pinned in a footer. `AppInspector` gained the
on-dark tab inks its own prop docs said a lightbox pane must add first.

Two things I had dropped as YAGNI are in because the design names them:
**Seed** and **Negative**. The seed is served as text as well as a number —
ComfyUI draws up to 2^64-1 and a JS `Number` loses digits above 2^53, so the
old field would have printed the wrong seed for about half of real ones.

The **workflow JSON with Copy and Download is kept**, at the owner's direction.
The design drops it in favour of Open; pasting a graph into ComfyUI and saving
the file are real uses of it, so it stays as a collapsed box at the foot.

**Four things the design draws are absent rather than faked**, each gated on a
step that is not in `develop`: the "Matches your saved recipe X" banner and
**Save** (saved recipes), the workflow's name and its stack (#1395 B3), the
"workflow: 8" override column (#1397 B5's owner-only route), and **Run…** as the
Run popup (#1407). *Generate variants…* stands in, visible-and-disabled with the
reason as its tooltip when it cannot run — the shape the design gives an A1111
picture. #1406 F4 is where those land.

Checks: 4414 frontend, 281 backend, ruff and the generated route table clean.
Five new mutations run red and restored green. Four overlay suites needed the
`vuetifyStubs` barrel mock, because the tab band renders a real `VIcon`.

lindkvis:
`checks` was red on the frontend suite: 7 tests in
`ImageOverlayTurnRefresh.test.js`, which landed on `develop` (#1423) after this
branch forked. The failure is mine, not develop's — the Info | Recipe tab band
makes `AppInspector` render a `VIcon` imported from the `vuetify/components`
barrel, and that suite mounts `ImageOverlay` without the Vuetify plugin, so the
icon's setup throws and takes the mount with it.

Merged `develop` in (no force-push) and reproduced it exactly: 7 failed, 4447
passed, 259 files, same as CI. The fix is the repo's own
`testing/vuetifyStubs.js` helper, which exists for "suites that only need the
Vuetify tags to resolve" — the same four-line `vi.mock` four sibling overlay
suites already carry on this branch.

**That is a test file, so here is why it is not papering over.** No assertion
changed and none was removed; the stub only lets the component mount. I
mutation-checked it afterwards against the code it guards: ignoring the turn
signal fails 5 of 7, removing the repeated-key guard fails 1, and acting on a
signal naming other pictures fails 1 — then green again restored. The suite
still measures what it claims to.

The alternative code fixes were worse: dropping the tab icons contradicts the
design, and making `AppInspector` rely on globally-registered `<v-icon>` would
turn a loud failure into silently missing icons for its three other consumers.

`vi.mock` is per-file by construction and the helper's docstring forbids
hoisting it into `testing/setup.js`, so a sixth overlay suite will need it too;
`frontend_architecture.md` now says so beside the tab-band entry.

Post-merge: 4454 frontend and 326 backend pass, ruff and the generated route
table clean.

lindkvis:
Merged `develop` (no rebase, no force-push). Two conflicts, both in the recipe
read, because #1428 landed **B5 (#1397) — the extended picture recipe read**,
which is the sanctioned version of backend work this branch had built ad hoc.

**`pixlstash/utils/comfyui_utilities.py` — resolved in develop's favour, and
mine had a real bug.** I had followed the negative conditioning chain with the
positive walker, so a node taking both inputs (a ControlNet applier, a
conditioning combine) would have dragged the negative walk onto the positive
prompt. B5's `_follow_prompt_api` is side-aware and does not. My duplicate walk
is deleted; `extract_comfy_workflow_info` now calls B5's public
`extract_recipe_extras` instead. My footprint in that file is 8 additive lines,
and `git diff origin/develop` shows **no line of develop's removed**.

I also dropped my `negative_prompt` key from `extract_generation_info`'s three
returns: B5 deliberately split that out so only the API format is read for it,
and my tests asserting the old shape were re-pointed at what this branch
actually adds (that the workflow read *surfaces* it, and `seed_text`) rather
than re-testing B5's code, which its own suite covers.

**`tests/test_comfyui_recipe_route.py` — kept both sides.** Purely additive;
the one repair needed was that a conflict boundary fell inside develop's
`_upload_one`, which I restored whole. No line of develop's was lost there
either.

## One thing I did not decide, because it is not mine

There are now **two overlapping recipe reads**: B5's `/comfyui/pictures/{id}/recipe`
(negative prompt, settings, strengths, `workflow_key`, an A1111 branch) and this
branch's fields on `/comfyui/pictures/{id}/workflow`.

I did not collapse them because `/recipe` still calls `_read_object_info(comfyui_url)`
unconditionally — a live ComfyUI round-trip per request — and the Recipe tab
reloads on every filmstrip step. Pointing the tab at it would put a network call
behind every arrow-key.

Neither read is a superset today: B5 has the A1111 branch and typed settings
with non-finite refusal; this branch has shelf resolution (`model_id` /
`verified`), the resolution lock, `seed_text`, and `width`/`height` (absent from
B5's `_SETTING_FIELDS`, so the design's Size row cannot be built from it).

The right end state is one read — `/recipe` gaining a no-preflight mode and
those fields, with my `/workflow` additions deleted. That changes a contract
merged hours ago and belongs to #1397/#1406, so I have flagged it rather than
taken it unilaterally. Say which way and I will do it.

Post-merge: 4454 frontend, 418 backend, ruff and the generated route table clean.

lindkvis:
Independent review run over this PR; both of its blockers were real and are
fixed in `938a4e73`.

**BLOCK 1 — two-sampler graphs lost their Settings block.** `_settings` qualifies
a repeated label with its node (`steps (Hires Fix)`); the panel looked up the
bare `steps`, got nothing, and skipped the row. Reproduced: a hires-fix graph
showed **no Steps and no CFG** — emptiest on the graphs most worth reading. The
panel now groups by the bare name and renders one row per node, pairing Sampler
with its own node's scheduler and Size with its own node's height. Two tests
cross that seam using labels copied from the service's real output, and both
mutations bite. One of my mutation attempts silently didn't apply because
prettier had reflowed the target across three lines — re-run against the real
text, which is the only reason I caught that the pairing test did guard.

**BLOCK 2 — migration 0122 is deleted, not patched.** It NULLed `comfyui_models`,
which makes `write_comfyui` true, and for a picture whose file is unreachable
(external drive detached at the first start after upgrade) the task's
early-return branch writes `"[]"` over real models and LoRAs, with `downgrade()`
unable to restore them. `comfyui_extraction_task.py`'s own comment names that
hazard verbatim and I walked into it. Two further claims in its docstring were
false: `persist` writes the prompt only when the new value is not None, so a
wrong prompt that should become empty stays; and the predicate column
`workflow_topology_hash` is hub-written, so a hub-absent library matched zero
rows. It also nulled `text_embedding` library-wide, which I had described as
"one metadata read each".

Regenerating the stored fields is still owed, and needs its own change: skip the
write when the file is unreachable, clear the prompt when the new read has none,
a predicate that works without a hub, an honest cost note, and a test like
0118's. The code fix stands alone — every import and rescan from now on reads
the executed graph. The changelog fragment now says exactly that instead of
promising a backfill.

Also fixed from the same review: the Recipe tab showed the **previous** picture's
prompt and models until the next fetch returned (`comfyMetadata` is now cleared
before the request); **Copy** was offered for API-format JSON that cannot be
pasted into ComfyUI, re-introducing a bug the `OverlayMetadataPanel` guard had
already fixed — the guard is back and asserted in both directions, and my old
test had been asserting the bug; and the Recipe tab is **no longer disabled** on
a picture with no recipe, because a disabled button fires no pointer events so
its explanation was unreachable — the design gives A1111 pictures the same tab
and the panel already explains itself in a sentence. That let me revert the
`AppInspector` change entirely rather than alter a shared widget for four
consumers.

Docs corrected: `backend_architecture.md`'s chunk-selection paragraph said
`find_comfy_workflow` drives the extraction backfill and `find_comfy_api_prompt`
is only for what runs — both false after this change. `integration_architecture.md`
gained the `/workflow` contract (§2.2b) including the `seed_text` rule, the
node-qualified `settings` rule, and an explicit note that it overlaps `/recipe`
and that A1111 pictures are readable there and 404 here.

Left open deliberately, with the reasoning on record: the `/workflow` ↔ `/recipe`
duplication (needs `/recipe` to skip its pre-flight — #1397/#1406's call), the
two deep links answering the same gesture differently, and `recipe_asset_index`'s
three table scans per overlay open (measured 2.5 / 4.9 / 32 ms at 500 / 2k / 10k
models).

426 backend, 4457 frontend, ruff and the generated route table clean.

lindkvis:
Reconciled. There is now **one recipe read**, and the split between the two
routes is by question rather than by caller:

| Route | Answers |
|---|---|
| `GET /comfyui/pictures/{id}/recipe` | what the picture was **made with** — prompts, models with strengths, settings, seed, shelf rows, resolution lock, `workflow_key`, `topology_hash` |
| `GET /comfyui/pictures/{id}/workflow` | the graph's **bytes**, in the editor's format, for Copy/Download/paste |

**`?preflight=false`** is what made one read possible. B5's `/recipe` called
`_read_object_info` unconditionally — a ComfyUI round-trip per request — which
is why the tab could not use it. The flag skips it, `preflight.checked` is then
false (which already means *the question was not asked*), and the Remix dialog
keeps the default because it is about to run the thing. A test makes
`fetch_object_info` raise and asserts the lightbox read never touches it, with
the default as the control; removing the flag's effect turns it red.

**What moved, and what I deleted rather than kept.** `model_slots`, `inputs`,
`seed_text` and `topology_hash` moved onto `/recipe`; `/workflow` is back to the
shape it had before this branch. My `describe_parameters`-based settings reader
is **gone** — B5's `extract_recipe_extras` is the one implementation, extended
with `width`/`height` (the design's Size row) and `guidance` (Flux has no CFG).
That also let me revert the `FEATURED_NAMES` rename I had made only for that
import.

Two consequences worth stating plainly:

- **The panel gains A1111 pictures.** `/workflow` 404s on them; `/recipe` reads
  their infotext. The settings grid renders whatever keys arrive, so A1111's own
  vocabulary works without branching on `source` — asserted.
- **The panel loses the second sampler's values.** My reader qualified a
  repeated label by node (`steps (Hires Fix)`); B5's is first-node-wins and
  documents that as "what this graph says". I took B5's, so the two hires-fix
  tests I added last round are deleted — the capability is gone, and asserting
  it would be asserting nothing. If that matters it belongs in
  `extract_recipe_extras`, where both callers get it, not in a second reader.

One of B5's own tests changed: it asserted the exact settings dict, and its
fixture already carried `width`/`height` on the scheduler node unread. The
assertion is still exact, it now covers them.

The workflow JSON box is **lazy** — `/workflow` is fetched only when the box is
opened, so the graph, the one large thing here, is not shipped on every
filmstrip step.

`integration_architecture.md` §2.2b now states the whole contract: the two
routes, the pre-flight flag, and the three field rules (`seed_text` is what a
client prints; `model_slots[].model_id`/`.verified` and `inputs` are owner-only;
`settings` is one dict whose first-node-wins limitation is explicit).

424 backend, 4456 frontend, ruff and the generated route table clean.

PR #1427, opened 2026-09-17
Workflows & Recipes B6: saved recipes in the vault

### LINKED ISSUE #1398: Workflows & Recipes B6: saved recipes in the vault

## Depends on

|---|---|---|

## What

- A new vault table via Alembic: a conditional create guarded by `inspector.has_table` (baseline `create_all` already makes it on a fresh vault), with `__all__`. Fields: `id, name, position, workflow_key, prompt, negative, loras JSON [{filename, sha256, strength}], overrides JSON by parameter address, seed NULL, keep_seed, source_picture_id NULL, created_at`.
- Routes: `GET/POST /recipes`, `PATCH/DELETE /recipes/{id}`, `PUT /recipes/order`, `GET /recipes/{id}/export` (the export body lands in B8 if simpler).
- **Ownership:** a recipe belongs to the workflow it was saved from and runs on any member of that workflow's stack. The stack's Recipes tab lists every member's recipes, and on Unstack each recipe stays with its own workflow.
- **Credit computed on read:** kept pictures of the stack's variants whose `comfyui_positive_prompt` equals the recipe prompt and whose `comfyui_loras` names match. Seed and strengths are ignored. No link table.

## Acceptance

- Migration test on a fresh and an existing vault. All routes `OWNER_ONLY` with both-direction authz tests. Credit tests include a non-match (different LoRA).

lindkvis:
Closes #1398. Step B6 of v1.12 Workflows & Recipes: the place a saved recipe lives, and the five routes that read and write one. Backend only — the Recipes tab is F6.

## What this adds

- **`saved_recipe`, a vault table** (`db_models/saved_recipe.py`, migration `0121`). A saved recipe is authored and holds the owner's prompt, so it cannot be rebuilt the way the hub's workflow rows can and it has to travel with a snapshot. The create is conditional (`inspector.has_table`), because the baseline's `create_all()` already builds it on a fresh vault.
- **`GET`/`POST /recipes`, `PATCH`/`DELETE /recipes/{id}`, `PUT /recipes/order`**, all `OWNER_ONLY` in `ROUTE_POLICIES`, rowed in the coverage matrix, and `/api/v1/recipes` added to `READ_BLOCKED_GET_PATHS` (the host-capability guardrail derives that list, so it is not optional).
- **Ownership by `workflow_key`** (decision D10): a recipe belongs to the workflow it was saved from and lists across that workflow's stack, resolved in the hub by `effective_stack_keys` — an explicit `workflow_stack_member` row, then `workflow_unstacked`, then the automatic `core_hash` group. Nothing names a stack, so an Unstack leaves each recipe with its own workflow.
- **Credit computed on read, no link table**: one grouped vault query per stack, matched in Python on prompt plus normalised LoRA names. Seeds and strengths are ignored, because no picture row stores a strength.

`GET /recipes/{id}/export` is deliberately not here. The issue allows it to land in B8 ("if simpler"), which is where export already sits together with the forgotten-name guard and the CSO review of what an export may carry.

## Answering the pre-merge adversarial review

Most findings were real and are fixed in the second commit: credit now counts only pictures the extraction pass has actually read (a NULL `comfyui_loras` is the never-checked sentinel, and folding it in credited every un-read picture to a recipe saved with no prompt and no LoRAs); prompts are compared stripped on both sides, so a recipe saved from a picture credits that picture; an unknown `source_picture_id` and a null `name`/`prompt` are answered as 422 and as "clear it" rather than surfacing the vault's constraints as 500s; `overrides` and `seed` got the ceiling the module claimed for its other fields; and a reorder now permutes the positions its own rows hold, so one tab's reordering cannot displace another workflow's recipes or land the next save in the middle of them. Six mutations that the first version's tests survived are now each red, and the migration test asserts the table's shape against both the model's build and the migration's.

Three findings I am answering rather than changing:

- **The `core_hash` grouping is expressed in two places** (`card_grouping` and `effective_stack_keys`). They answer different questions — a report over every card, and one card's stack — and merging them would make the per-key read pay for a full grouping. Worth revisiting when B3/B4 add the card reads that will want both.
- **Binding one SQL variable per key has no chunking.** The runtime's SQLite allows 250,000 host parameters (3.45, limit read back at runtime); the largest plausible stack is three orders of magnitude below that.
- **`_hub()`'s 503 is hard to reach**, since `Server` assigns `self.hub` unconditionally. It mirrors `routes/workflows.py` exactly and guards the same `getattr`, so I would rather remove both together than leave the two modules disagreeing.

One more fix reaches beyond the feature: `server.py`'s `RequestValidationError` handler passed pydantic's error list to `JSONResponse` unencoded. A validator raising `ValueError` puts the exception object in the error's `ctx`, so the handler for a 422 failed with a 500 instead — the first validator in this repo to raise one hit it. It now encodes first, as FastAPI's own handler does.

No changelog fragment: nothing here is visible to a user until the Recipes tab lands.


lindkvis:
## Review: approve with nits

Backend review plus an independent adversarial authz/security pass. No blocker. The core — vault row, migration, authz declaration, credit matching, reorder closure — is correct, and the tests are unusually good: four mutations run, four bit. Findings are split into **verified by running code** and **reasoning only**.

One result reaches beyond this PR and is the most valuable thing in it: **the `server.py` fix repairs a live 500 already shipping on `develop`.** See the last section.

---

### Should-fix

**S1 — neither new hub query filters `key_version`, unlike the other readers. `hub/workflow_cards.py:421,469`. VERIFIED.**

`_VARIANT_JOIN` carries `AND v.key_version = ?` at :80, which is what makes `unidentified_variants` and `variant_counts` treat a superseded row as pending-with-no-card. Both new queries omit it (as does `card_grouping`, so the new code is consistent with that one reader and inconsistent with the other three). Confirmed by reading the two new statements: the `workflow_stack_member` lookup and the `workflow_variant WHERE workflow_key IN (…)` both go unfiltered. Stamping every variant superseded still returns `keys(A) = [B, A]` and two variant hashes.

The consequence is **orphaning**. A `WORKFLOW_KEY_VERSION` bump or a slot-mark flip re-keys the variant, and nothing migrates `saved_recipe.workflow_key`. Reproduced by re-keying: under the old key, 4 rows exist but credit resolves to 0; under the new key, 0 rows and the tab is empty; unfiltered, the 4 rows are still there. The model docstring is carefully worded — it claims survival of regrouping, Unstack and a `CORE_VERSION` bump, and all three genuinely hold — so nothing here is a false claim. But there is no detection, no repair and no test for the one bump that breaks it.

This is the item I would fix before merge, for a reason specific to this table: a saved recipe is **authored** and holds the owner's own prompt, so unlike the hub's derivable workflow rows it cannot be rebuilt. Silently losing content that by design cannot be regenerated is worse than the usual stale-read bug.

**S2 — `effective_stack_keys` and `card_grouping` do disagree. `hub/workflow_cards.py:391` vs `:341`. VERIFIED.**

The PR judges these two consistent. They are not. `effective_stack_keys` excludes cards holding `workflow_stack_member` or `workflow_unstacked` rows; `card_grouping` consults neither. Reproduced with A and B sharing a `core_hash`, C alone, and A unstacked: `keys(A) = [A]` and `keys(B) = [B]`, while `card_grouping` reports `{stacks: 1, stacked_cards: 2, one_offs: 1}`. So the gate reports one live two-card stack where the shipped read path sees three singletons.

That matters because the gate exists to judge whether `STRIP_LORAS_FOR_STACKS` over-groups — it is now measuring a grouping the Recipes tab will not use. Both docstrings are individually honest, so nothing is a lie; one of them should either gain the exclusions or state that it deliberately reports the pre-decision rule. I accept the PR's argument that merging the two implementations would make the per-key read pay for a full grouping — the ask is a decision recorded in the docstrings, not a refactor.

**S3 — the degenerate-recipe sweep is only half-closed, and the open half is untested. `services/saved_recipe_service.py:194`. VERIFIED.**

The NULL sentinel is real (mutation 1 below proves it), but it closes only the `comfyui_loras IS NULL` half. A picture that *was* read and yielded no prompt holds `comfyui_loras = "[]"` with `comfyui_positive_prompt = NULL`, keying as `("", frozenset())` — exactly a recipe saved with an empty prompt and no LoRAs. Measured: such a recipe credited 2 pictures where a tight sentinel gives 0. A whitespace-only stored prompt does the same, crediting 1. The population is real rather than hypothetical: `workflow_io.py` documents two ways a *filed* graph yields no prompt (`ConditioningZeroOut`, and disagreeing samplers), and prompt-free upscale graphs are ordinary.

To be fair to the design: the behaviour is defensible, since credit also ignores `overrides` and `seed` for the same stated reason it ignores strengths. The defect is that nobody chose it and nothing tests it, while the docstring claims the sentinel stops crediting "a library's worth of pictures it never made" — true of the NULL half only.

If the intended answer is 0, the fix is exact: both writers of `comfyui_positive_prompt` (`tasks/comfyui_extraction_task.py:233`, `utils/image_utils.py:1074`) coerce via `... or None`, so **no row ever stores an empty-string prompt**. That makes `.where(Picture.comfyui_positive_prompt.is_not(None))` precise, with no legitimate row excluded.

**S4 — `PUT /recipes/order` with a large id list is a 500. `routes/recipes.py:132`. VERIFIED.**

`recipe_ids: list[int]` has no `max_length`, and the whole list goes into one `id.in_(...)`. Measured against a live server: 40,001 ids → 404 (fine), 100,001 → 404 (fine), **300,001 → 500 `sqlite3.OperationalError: too many SQL variables`**. This is the same "never surface a DB constraint as a 500" class the PR handles correctly for the foreign key.

Not a regression and not a blocker: there are **14 pre-existing unbounded id-list fields across 8 route modules** (`characters_faces` ×2, `config` ×2, `library_layout` ×2, `picture_sets` ×2, `pictures/_import` ×2, `reference_folders` ×3, `stacks` ×2, `tag_predictions` ×1). But `routes/dedup.py:1290` already shows the house pattern (`max_length=500`), and one field closes the new instance.

---

### Nits

- **Set-collapse in credit matching.** `lora_names` is a `frozenset`, so a recipe using `[Ada@1.0, Ada@0.2]` — a stacked LoRA, a materially different look — keys identically to one using Ada once. Measured credit 2.
- **`GET /recipes` unfiltered is uncapped** (30 saved → 30 returned), where the workflows route caps at 60.
- **The 422 body echoes the whole oversized field** (>30 KB response). Pre-existing behaviour, but the new ceilings make it easy to hit.
- **`_decode(recipe_id=<a sentence>)`** at `services/saved_recipe_service.py:346` — deliberate, but the parameter name misleads; rename it.
- **`PUT /recipes/order` is declared after `PATCH`/`DELETE /recipes/{recipe_id}`.** Safe today since there is no `PUT /{recipe_id}`, but it is a latent shadowing hazard.
- **Typo** in `tests/test_migrations.py`: "The The same run covers both".

### Reorder edge cases: correct, two unspecified

The closure invariant **holds** in every case — positions are permuted within the request's own row set, the multiset is conserved, uniqueness is preserved, and no outsider moves. Duplicate id → 400 (tested). Unknown id → 404 with nothing applied (tested and mutation-checked). Empty list → 200 no-op. The next `POST` lands after `max(position)` (tested). Concurrent delete → 404 refusing the whole gesture, and it is race-free: the select and the updates share one session inside one `vault.db.run_task`.

Two cases are closed but unspecified and untested: a **partial subset** (`A1(0) A2(1) A3(2)` + `[A3, A1]` → 200, giving `A3(0) A2(1) A1(2)`, which wedges A2 between the two rows the owner reordered), and a **foreign-workflow id**, which is accepted and permuted across workflows because the route is not workflow-scoped at all. Both deserve a line of documented intent, and a test.

---

- **`tests/test_saved_recipes.py`: 26 passed.** `ruff check` and `ruff format --check` clean on `pixlstash tests scripts`.
- **Coverage-matrix arithmetic exact**: `len(ROUTE_POLICIES) == 336`, `owner_only == 157`.
- **Four mutations run, four bit:** dropping `.where(Picture.comfyui_loras.is_not(None))` reddened the never-read-credit test (`assert 2 == 0`); making the reorder write `range(len(ids))` instead of permuting reddened the cross-tab test; reverting `server.py` to `content={"detail": detail}` reddened the oversized-field test with `TypeError: Object of type ValueError is not JSON serializable`; replacing the all-or-nothing reorder guard with a silent filter reddened the unknown-id test.
- **Migration 0121 is clean**: conditional `has_table`, re-inspects after CREATE for the inspector-cache subtlety, both indexes guarded by name, `__all__` present, single head, correct order, no application logic, symmetric downgrade, and the `picture`-absent early return matches 18 sibling migrations. `tests/test_migrations.py` asserts the table shape **twice** — against the baseline `create_all()` and against the migration's own CREATE — covering columns, both index names and the `SET NULL` FK. That is the right way to hold two definitions of one table in step.
- **Authz is clean.** Five routes all `OWNER_ONLY`, declared only in `ROUTE_POLICIES` with no inline ladder, `GET /recipes` added to `READ_BLOCKED_GET_PATHS`, and the test correctly pins the declaration itself because the writes are refused by the READ-token middleware ahead of the gate. Independently confirmed: no ad-hoc scope check is added anywhere, the four `resolved_inline=True` routes are untouched, and `READ_BLOCKED` arithmetic is complete.
- **Credit symmetry is right**: SQL groups on raw values, Python accumulates with `+` so groups normalising to one key are summed rather than overwritten, and it uses `get` not `pop` so a group counts for every matching recipe.
- **Exception handling**: three `except` clauses, no bare `pass`, each logged with row/column/value or re-raised with `from exc`.
- **Error paths**: unknown `source_picture_id` → 422 on both verbs, and the test proves a real id still works, so the check is not merely refusing the field; null name/prompt → `""`; oversized `overrides`/`seed` → 422.
- **Non-finite floats are a non-issue here, contrary to first appearances.** Constructing `RecipeLora(strength=inf)` in Python does keep `inf` and would break a listing render — but over the real wire against a live server, `POST /recipes` with `Infinity`/`NaN` returns 201 with `strength: null`, and `GET /recipes` returns 200. FastAPI's response encoding neutralises it end to end, and `overrides` behaves the same. Recorded because it reads convincingly from the model alone and dies on the wire.
- CI gating correct: `tests/test_saved_recipes.py` is added to the Linux `backend` job in correct alphabetical position. The Windows "OS-sensitive" list is a curated subset and owes no entry.
- Secret/private-information scan clean on added lines; the fixture uses the sanctioned `example-` marker.

---

### The `server.py` change fixes a live bug on `develop`

Encoding-first is correct (it matches FastAPI's own handler) and complete. I dumped the resulting body: `ctx.error` encodes to `{}` because `jsonable_encoder` falls back to `vars(exc)`, but pydantic has already copied the text into `msg` (`"Value error, overrides is longer than 20000 characters serialised."`), so nothing the client needs is lost. The password short-circuit above is unaffected.

It also fixes the **class**, and the class had a live member. Exactly **2** validators in `pixlstash/routes/` raise a bare `ValueError` that reaches this handler: the new `recipes.py:51`, and the pre-existing `dedup.py:1297` — which I confirmed is a `@model_validator(mode="after")` on `VerdictBatchRequestModel`, not a handler-body raise like the other four `ValueError` sites in `routes/`. Exercised both ways: `POST /api/v1/dedup/verdicts/batch` with a duplicate signature returns **500 on `develop`** and **422 with this PR**.

Two asks follow:

1. **A regression test naming `dedup`.** Right now the only thing guarding a dedup endpoint is a recipes test, which the next person to touch `recipes.py` can delete without knowing what they broke.
2. **Consider a changelog fragment** — not for the recipes feature, which is correctly fragment-free until the tab lands, but for the dedup 500, if that bug reached a release. A user hitting a 500 on a duplicate-signature batch is a user-visible fix.

One note for future validators: this handler now makes a validator's `msg` text caller-visible by contract. No current validator embeds a path, a secret or the input value, so nothing leaks today — a one-line warning comment would keep it that way.

---

### Before merge

- **Rebase.** This branch is 7 commits behind `develop`, and its own copy of the guardrail file is missing `test_base_model_widgets_agree_across_the_stack`, which exists on `develop`.
- **Resolve the registry/matrix conflict with #1429.** Both PRs insert new rows at the same anchor in `pixlstash/authz/registry.py` and `docs/authz-coverage-matrix.md`, so the second to land conflicts. This is loud — git stops — not a silent undeclared-route hole. Replaying the merges in a sandbox: #1428 merges clean, then #1429 clean, then #1427 conflicts in exactly those two files. A union resolution gives **339 declared**, and on that merged tree the guardrails pass (86) along with all three branches' own suites (248).
- **One trap in that resolution:** `test_coverage_matrix_document_matches_the_registry` parses only the row tables, **not** the prose "CURRENT: N declared" header bullets. After a union merge both bullets (336 and 334) survive and neither equals the true 339, and **CI still passes**. Whoever lands second must union-resolve *and* re-derive the header count by hand.

---

lindkvis:
Thanks — this found real things. Everything below is pushed as `2c84d898`, on top of a rebase onto current `develop`.

## Should-fix

**S1 — `key_version`.** Fixed: both new reads now filter on `WORKFLOW_KEY_VERSION`, so a variant filed under a superseded rule neither joins a stack nor credits its pictures. Guarded by `test_a_variant_keyed_by_a_superseded_rule_neither_stacks_nor_credits`, which seeds two stale rows — one giving a fourth card a place in A's automatic group, one hanging a second variant off B — because the two queries fail independently; both filters were unguarded before it, and removing either now reddens it. The docstring says what the filter costs: mid-re-derivation a stack shrinks to itself, which reads as smaller rather than wrong.

The orphaning half is a debt, and I have written it down rather than papered over it — in `SavedRecipe`'s docstring and in `backend_architecture.md`: a re-keying leaves the recipe addressed by a key no variant carries, it is worse here than for a hub row because a saved recipe cannot be re-derived, and **whoever bumps `WORKFLOW_KEY_VERSION` re-keys this table in the same change** (the old key is recoverable, since the new one comes off the same stored documents). No bump exists yet, so there is nothing to migrate today and a migration written now would be untested against the rule it is supposed to follow.

**S2 — the two groupings.** You are right that they disagree, and I was wrong to call them consistent. Recorded as you asked, in both docstrings: `card_grouping` reports the **automatic** rule before the owner's decisions, deliberately, because the gate it feeds asks whether `STRIP_LORAS_FOR_STACKS` groups well — a question about the rule, not about what has since been moved out of it. `effective_stack_keys` answers what a card's stack *is*. A reader comparing them should expect the difference once a card has been moved.

**S3 — the prompt-free half.** Decided, documented and tested, and I went the other way from the suggestion. A picture that *was* read and carries no prompt now matches a recipe with no prompt on purpose: a prompt-free graph is ordinary (upscale, `ConditioningZeroOut`, disagreeing samplers), those pictures really were made by that stack with that look, and the stack bounds the answer. Excluding NULL prompts would make a recipe saved on an upscale workflow read 0 for ever, which is the same defect pointing the other way. The NULL-`comfyui_loras` sentinel stays, because that one is absence of evidence rather than evidence of absence. `test_a_picture_read_with_no_prompt_matches_a_recipe_with_no_prompt` pins it, and adding your `comfyui_positive_prompt.is_not(None)` filter reddens it.

**S4 — unbounded id list.** Fixed with `max_length=500`, the `routes/dedup.py` house pattern, and a test that a 100,001-id list is a 422 while an ordinary reorder still works.

## Nits

- **Set-collapse.** Fixed properly rather than documented: `lora_names` is now `lora_key`, a **sorted tuple with duplicates kept**, because the picture side writes one entry per loader node, so `[Ada, Ada]` is a real value on both sides. A recipe stacking Ada twice no longer credits pictures that loaded it once, and vice versa.
- **`_decode(recipe_id=…)`.** Renamed to `where`. The rename earned itself immediately: fixing the log call revealed a format-argument mismatch that would have raised inside the warning path, which the corrupt-JSON test caught.
- **Route order.** `PUT /recipes/order` now precedes the `{recipe_id}` routes, with the shadowing reason in a comment.
- **Typo.** Fixed.
- **`GET /recipes` uncapped — not changing.** It is the owner's own complete list of their own recipes; a silent cap would hide recipes with nothing saying so, which is worse than a long response. The workflows route caps *picture ids* for a six-tile strip, which is a different thing. If this list ever gets big the answer is pagination with a total, not truncation.
- **The 422 echoing the oversized field — not changing here.** It is FastAPI putting `input` in the error, unchanged by this PR, and suppressing it globally would remove the field that tells a client which value was rejected. Worth its own change if somebody wants the whole app to stop echoing rejected bodies.

## Reorder edge cases

Both now documented in `reorder_in_session` as intended behaviour, with a test each: a **subset** moves only the rows it names (`[A3, A1]` out of `A1 A2 A3` leaves A2 between them — two rows sent, two rows moved), and the route is deliberately **not workflow-scoped**, because a stack's tab reorders recipes belonging to several members on every gesture.

## The `server.py` fix

Both asks done. `tests/test_dedup_tiers_api.py::test_a_duplicate_signature_is_refused_as_a_bad_request_not_a_crash` names the endpoint that has the validator and explains why it exists; reverting the encoder turns it red. And a changelog fragment, since `cb656898` is contained in `v1.10.0` — that 500 shipped. The fragment covers the dedup fix only; the recipes feature stays fragment-free until the tab lands. I also put your warning on the handler: whatever a validator writes in its `ValueError` reaches the caller, so it must not carry a host path, a credential or the input value.

## Before merge

Rebased onto `develop` (`ab3016dd`), so the missing guardrail test is back. Counts re-derived after the rebase: `len(ROUTE_POLICIES) == 336`, `owner_only == 157` — unchanged, since #1429 has not landed yet. If it lands first, whoever resolves the conflict must union the rows **and** re-derive the header bullet by hand, exactly as you say — the guardrail parses the tables and never reads that prose.

PR #1428, opened 2026-09-17
Workflows & Recipes B5: extended picture recipe read and workflow picture filters

### LINKED ISSUE #1397: Workflows & Recipes B5: extended picture recipe read and workflow picture filters

lindkvis:
Closes #1397. Part of v1.12 Workflows & Recipes (B5).

## What

`GET /comfyui/pictures/{id}/recipe` now reports the whole recipe, not only what
replays it:

- **negative prompt**, by the same conditioning walk as the positive one,
  parameterised by the side it started on (a node taking both must be followed
  on the caller's side, or the negative chain arrives at the positive prompt);
- **settings** (steps, cfg, sampler_name, scheduler, denoise) read from **any**
  node that names one, not from the sampler — the shipped `Flux2-Klein`
  templates put steps on a scheduler, the sampler name on a `KSamplerSelect`
  and the CFG on a `CFGGuider`, so a sampler-only read answers with one field
  out of five for our own workflows;
- **LoRA strengths** per slot (`detect_lora_targets` gains `strengths`), picked
  out by the slot's own numeric suffix;
- **`workflow_key`**, a hub read on `picture.workflow_structural_hash` — never
  a derivation, so an unfiled picture honestly has none;
- **an A1111 branch** through `reduce_a1111`: `source: "a1111"`,
  `available: false`, with the prompts, seed, model, LoRAs and fields read off
  the reduced nodes so this route and the hub agree what the recipe is.

`GET /pictures` gains `workflow_key` / `workflow_stack`, resolved against the
hub into the card's variants and matched as an `IN` on
`picture.workflow_structural_hash`. An empty resolution is a filter matching
nothing, never the absence of one — that holds for a card nothing was made
with, an empty parameter value, an absent hub and a hub that will not answer.
The `find_unassigned` and smart-score-candidate branches are threaded too;
those name each filter rather than splatting them, so the filter would
otherwise have been silently ignored on them.

Route paths and access policies are unchanged, so `ROUTE_POLICIES` is untouched.

## Security (chief-security-officer line)

The new file-derived fields — negative prompt, settings, LoRA strengths, and
the A1111 fields — are read from the picture's own embedded metadata and expose
no more to a share-token holder than the `positive_prompt` beside them, or than
the `/comfyui/pictures/{id}/workflow` sibling which already hands the same
holder the whole graph.

**`workflow_key` is the one that needs stating rather than asserting.** It is an
opaque digest over that same graph, but it is *not* purely a function of the
file: it folds in which LoRA slots the topology marks structural, and that mark
was frozen from the filename of whichever picture of that topology was filed
first in the library. A holder of two pictures differing only in a LoRA can
therefore learn one bit the files alone do not give them — whether that slot
counts as part of the workflow — decided by a published rule over a filename.
One bit per topology; no filename, nothing about a picture they cannot see.
That is the whole exposure, and it is written down in the code and in
`backend_architecture.md` rather than papered over. What a key *groups* stays
an owner-only question.

Two hardening fixes the adversarial pass found, both reproduced first:

- **Non-finite floats.** This change was the first to emit floats from this
  route. `json.loads` accepts the `Infinity`/`NaN` literals and responses render
  with `allow_nan=False`, so a crafted `prompt` chunk (or an A1111
  `<lora:x:inf>`) turned the read into a 500 for anyone holding a share token.
  Refused in `_finite_setting`, `_lora_strengths` and `_a1111_strengths`, with
  tests that would raise if one got through.
- **`reduce_a1111` on a request path.** It had one caller, a background pass,
  and its own documented quadratic regex bounded per line at 8,000 characters —
  a line *count* multiplies that. `parse_infotext` now searches only the last 32
  lines, which the grammar loses nothing by: A1111 escapes newlines, so its
  fields line is the last.

## Review objections not taken

- **No scoped-token test for the new query params.** Nothing here is an authz
  change: no policy, gate or scope code is touched, the filter only appends an
  AND predicate, and every scope branch was threaded. Worth a second opinion if
  a reviewer disagrees.
- **Keeping the `workflow_stack_member` half of `variants_in_stack`** although
  nothing writes those tables yet. Dropping it would answer a stored stack with
  an empty grid the day one is written, and the contract tells clients an empty
  grid means "no pictures". It is now tested by writing the rows.
- **`workflow_unstacked` is not consulted** — nothing writes it either, and the
  step that does owns making its readers agree. Said so in the docstring.
- **No changelog fragment.** Nothing in the app changes yet; the Workflows
  screen step is what a release note would describe.
- The diff is over the ~600-line guidance at ~1,010, of which ~470 is tests and
  documentation.


lindkvis:
## Review: request changes

Backend review plus an independent adversarial authz/security pass. Findings are split into **verified by running code** and **reasoning only**; where a claim of yours was checked and holds, that is recorded too, because several of them do.

The headline result: the listing filter — the part I considered highest-risk — **is correct**, in all four fail-closed cases, and every scope branch genuinely threads it. The defects are elsewhere, and two of them are in exactly the class this PR set out to harden.

---

### Blockers

**B1 — `routes/comfyui.py:1206`: `int(recipe.seed)` returns 500 on crafted metadata. VERIFIED.**

`"seed": int(recipe.seed) if recipe.seed else None`, guarded only by `a1111_recipe.py:313`'s `seed if seed.isdigit() else None`. `str.isdigit()` accepts strings `int()` refuses. Confirmed on CPython 3.11.15:

| seed text | `isdigit()` | `int()` |
|---|---|---|
| `²` (U+00B2) | `True` | `ValueError: invalid literal` |
| 4301 digits | `True` | `ValueError: Exceeds the limit (4300 digits)` |

A picture whose `parameters` chunk is `x\nSteps: 20, Seed: ²` turns `GET /comfyui/pictures/{id}/recipe` into an HTTP 500 for any share-token holder — the 4301-digit variant fits inside the 8,000-character fields-line ceiling, so both are reachable. This is the same threat model the PR cites for its float hardening: every float path was guarded and the one `int()` was not. The `A1111Recipe` docstring's "None when it is absent **or not an integer**" is therefore false as written.

Population is **1** — the only other consumer (`tasks/comfyui_extraction_task.py:343`) keeps the value as text. The house fix already exists at `utils/workflow_hash.py:319` (`isdigit()` plus a length cap). Worth making the route's conversion defensive too, so the two cannot drift apart later.

**B2 — `services/a1111_recipe.py:169`: the 32-line window still permits ~20 s CPU per request, and this PR is what exposes it. VERIFIED, measured.**

`_MAX_FIELDS_LINE = 8_000` is enforced **per line** inside `_parse_fields`, while `_MAX_FIELDS_LINES = 32` bounds the search loop — so the ceiling multiplies rather than caps. Measured against this branch's own regex:

| input | cost |
|---|---|
| one 1,011-char line | 0.007 s |
| one 2,011-char line | 0.030 s |
| one 4,011-char line | 0.118 s |
| one 7,991-char line | **0.75 s** |
| 32 such lines (250 KB) | **19.98 s** |

(4× per doubling — the quadratic is real, not theoretical. Absolute ms are host-specific; the shape is what matters.)

Paying full price 32 times needs a line that contains `Steps:` — clearing the cheap substring guard — while yielding no `Steps` field, so the loop keeps walking: `'xSteps: 20,' + 'a'*7980` does it, since the captured key is `xSteps`. Before this PR the handler returned `no_prompt_chunk` and never reached `reduce_a1111`; the A1111 branch is what puts this on a request path, as your own comment at :158 acknowledges.

Two mitigations are real and worth stating: the handler is a sync `def`, so it burns a threadpool worker rather than blocking the event loop, and the pathological bytes must live in a picture the caller already holds. Neither makes it safe — it is uncached and repeatable, and a handful of concurrent requests exhausts the default threadpool and starves every other sync endpoint.

Also: the per-line ceiling was mis-extrapolated when it was chosen. 8,000 characters is ~0.75 s, not the ~0.14 s a naive scale from the comment's 19,000-chars/0.82 s datapoint suggests. The fix is to bound the **aggregate** characters scanned (8,000 total across the window, not per line), or to remove the quadratic from `_PARAM_RE`.

---

### Should-fix

**S1 — `utils/comfyui_utilities.py:384`: the conditioning walk can cross sides. VERIFIED.**

`for key in ("conditioning", side)` tries the generic key **first**. Given `KSampler.negative → X`, where `X = {conditioning: →"2", negative: →"3"}`, `2 = "POSITIVE"` and `3 = "NEGATIVE"`, the result is `negative_prompt: "POSITIVE"` — verbatim the failure the docstring says `side` exists to prevent. Impact is a mislabelled prompt rather than disclosure (the caller already has both prompts, and the whole graph via the `/workflow` sibling), so it is not a blocker. Fix is to iterate `(side, "conditioning")`.

The rest of the walk is sound and was checked rather than assumed: a ControlNet-shaped node consumed by both sides resolves correctly on each; a cycle terminates immediately; `_MAX_FOLLOW_DEPTH = 8` means a 12-hop chain honestly returns `None`; a dangling reference returns `None`. No visited-set is needed — a 2-way branch at depth 8 is ~256 calls, microseconds, not a DoS vector.

**S2 — `settings` collides on the key `steps` with two different types.** The ComfyUI branch emits `steps` as `int` and `cfg` as `float`; the A1111 branch emits `steps`, `cfg_scale`, `sampler`, `size` and `denoising_strength` **all as strings** (`'20'`). That contradicts the one-type-across-both-branches rule this module states for `strengths` in its own docstring. The integration doc telling clients to render A1111 as name/value pairs mitigates but does not remove it, because `steps` is documented as a ComfyUI key and exists on both branches — precisely where a client will reach.

**S3 — `utils/comfyui_utilities.py:541`: `_finite_setting` validates the value's kind, not the field's type. VERIFIED.** Input `{"cfg": NaN, "denoise": Infinity, "steps": "twenty", "sampler_name": 12345, "scheduler": ["9", 0]}` yields `{'steps': 'twenty', 'sampler_name': 12345}`. The non-finite floats are correctly refused, but `steps` can be any string and a 400-digit int passes through and renders. Same unexamined int/str door as B1.

---

### Nits

**N1 — `workflow_structural_hashes` is reachable from the query string. VERIFIED.** It is newly a declared `Picture.find` parameter but never popped, and the resolver only writes the key when it resolves — so `?workflow_structural_hashes=abc` reaches `PredicateFilter`, where pydantic refuses `str`→`List[str]` and returns 500. This is **not** a filter bypass: it can only add an `IN`, and the resolver overwrites it whenever the real filter resolves (confirmed). Population is **5** — this adds the fifth member of a pre-existing class (`comfyui_models_filter`, `comfyui_loras_filter`, `tags_filter`, `tags_rejected_filter` are all equally unpopped; only `impossible_sources` is popped). One defensive `pop` closes the new instance; the class is a separate follow-up.

**N2 — `_a1111_strengths` (`comfyui.py:1155`) swallows `(TypeError, ValueError)` unlogged.** Every other new `except` in the diff logs with context, and this is the one exception to CLAUDE.md's rule. It is defensible — `""`, `None`, `"heavy"` and `"0.8,0.5"` are all ordinary A1111 input and logging would be noise — but `logger.debug` would satisfy the rule at no cost.

**N3 — `_picture_workflow_key` re-reads the picture** with a second `Picture.find` when the handler has already resolved it.

---

### On the security bound this PR documents

The PR argues that `workflow_key` leaks "one bit per topology; no filename, nothing about a picture they cannot see." **That bound is too tight, and it does not hold.** Brute-forcing the full structural-mark set from a single observed key is cheap — the LoRA-slot set is tiny, so 2^n over it recovers the mark set uniquely and exactly for a 3-LoRA topology. The recovered bits that disagree with the holder's own filename-guess encode information about the **first-filed** picture of that topology (frozen by `hub/schema.py`'s `workflow_slot_mark INSERT OR IGNORE`), which may be outside the holder's scope.

What leaks is a filename-derived *classification*, not filenames, prompts or pixels, so this stays **low severity** and I am not treating it as a blocker. But the written bound is what a future reader will rely on, so it should either be restated honestly in both the code comment and `backend_architecture.md`, or `workflow_key` should be returned to the owner only. Documenting a bound more favourable than the code enforces is the part worth fixing.

Separately, and to your credit: **`ROUTE_POLICIES` is correctly untouched**, the "no authz change" claim is otherwise true and was verified (registry count byte-identical to `develop` at 331), no ad-hoc scope ladder is added anywhere, and the new filters compose as a narrowing `AND` with the scope predicate on every branch — no BOLA, no oracle beyond the `workflow_key` bit discussed above.

---

- **The filter fails closed in all four cases.** Card nobody used → `[]`; `?workflow_key=` empty → `[]` (presence-not-truthiness works because `dict(request.query_params)` retains the key); absent hub → `LookupError` → `[]`; raising hub → caught, logged, `[]`; neither parameter → `None`. And the empty list really matches nothing: on SQLAlchemy 2.0.54 `in_([])` compiles to `IN (SELECT 1 FROM (SELECT 1) WHERE 1!=1)` and returns zero rows against a populated table, while `in_(['h1'])` returns the row.
- **Threading verified by enumeration** across the smart-score `PredicateFilter` (including the `is None` fast-path short-circuit), **both** the `count_only` and listing branches of `find_unassigned`, and all four `**query_params` sites.
- **Settings precedence is deterministic, so that is not a defect.** The graph arrives via `json.loads`, so `.values()` follows JSON text order — the same document gives `steps: 10` across repeated reads, and swapping the nodes in the text flips it to 40. The genuine caveat (a hires-fix graph reporting pass 1's steps beside pass 2's cfg and denoise) is already documented in the docstring and both architecture docs.
- **`workflow_key` is a pure read.** Single `SELECT`, no fallback, unfiled pictures short-circuit before the hub is consulted, and an absent or raising hub yields `None` rather than a 500.
- **The 32-line grammar argument is defensible.** Prompt newlines land before the fields line and the walk goes backwards, `text.strip()` removes trailing blanks, and A1111 forks append extension data (`Hashes`, ADetailer, ControlNet) as comma-separated fields *on* that line rather than as new lines after it. I could not construct a realistic infotext the window misses — it would need more than 31 appended lines. The objection in B2 is to the cost, not to this reasoning.
- **The float refusal is complete.** All three float-emitting paths on the route (`_finite_setting`, `_lora_strengths`, `_a1111_strengths`) are guarded before render; the population was grepped and no unguarded path remains.
- **The new tests bite: 3 of 3 mutations went red.** Removing `_finite_setting`'s `isfinite` broke both the equality assertion and `json.dumps(allow_nan=False)`; removing `_a1111_strengths`' `isfinite` let `inf`/`-inf`/`nan` each report a strength while the `"0.8"` control still passed; turning `_resolve_workflow_filter`'s `return []` into `None` made both the absent-hub and broken-hub cases return the whole library. The coverage gaps map exactly onto B1, S1 and S3 — no int case, no side-crossing case, no per-field type case.
- `ruff check` and `ruff format --check` are clean on all 10 changed Python files.
- CI gating is correct and owes nothing: `tests/test_comfyui_recipe_preflight.py` already exists on `develop` and is already listed in both the Linux `backend` job and the Windows subset, so the untouched `ci.yml` is right.

### Your "objections not taken" — all five defensible

1. **No scoped-token test** — agreed, and the substance was checked rather than taken on faith: no policy, gate or membership code is touched, the filter only appends a narrowing `AND`, and every scope branch was enumerated. One nuance worth recording rather than objecting to: hub reads are library-global, so `?workflow_key=X` is an oracle for "is my picture on card X" — but that is the same bit the route's own `workflow_key` already hands the holder, subject to the correction above.
2. **Keeping the `workflow_stack_member` half** — correct trade. Dead-but-tested with a stated reason beats a grid that silently empties the day the table is written.
3. **`workflow_unstacked` not consulted** — defensible, documented, and the docstring assigns the over-reporting to the step that writes it. Right owner.
4. **No changelog fragment** — correct per CLAUDE.md. API surface only, nothing user-visible, and a wrong entry is worse than a missing one.
5. **Diff size** — fair; roughly 470 of ~1,010 lines are tests and documentation, and the docs here are load-bearing.

### What was not verified

The gated suite was not run end-to-end. Findings were verified by executing this branch's own code in isolation (`reduce_a1111`, `parse_infotext`, `_parse_fields`, `_finite_setting`, `detect_lora_targets`, plus `in_([])` against a real SQLite engine), not through the live handler and `response_model`. So B1's 500 is verified at the conversion, not at the HTTP boundary; the filter results are verified at the SQL layer, not over the wire. All 18 checks on `f516673` are green, but the suite contains no test for B1, S1 or S3, so green does not speak to them.

---

lindkvis:
Addressed in `de234355`. Every finding is taken; two are fixed deeper than asked, and I disagree with nothing. Reproducing each before fixing it was worth it — one of your nits turned out to be the same door as a blocker, and my own first fix for S3 had a bug in it.

### B1 — `int(recipe.seed)` 500. Fixed at the source, then defensively at the call.

Reproduced both vectors: `"²"` and `"1" * 4301` are `isdigit()`-true and raise in `int()` (`invalid literal`, `Exceeds the limit (4300 digits)`).

You were right that the `A1111Recipe` docstring was the real defect — it promised "not an integer" and tested `isdigit()`. So `reduce_a1111` now asks `int()` itself (`_integer_text`), which makes the field honest for both consumers rather than patching the one that converts. The route's conversion is defensive too, as you asked, and logs a warning if it ever fires — that log firing would mean the two have drifted, which is the thing worth knowing.

### B2 — the 32-line window. You were right that bounding the window was the wrong fix; the quadratic is gone.

Your measurement stands and my per-line ceiling was indeed mis-extrapolated (8,000 chars is 0.14 s on this host, not the ~0.04 s the old comment implied). But the cause is `_PARAM_RE`: `\w[\w \-/]+` before the literal `:` backtracks over every length of a colon-free run. Bounding that run at 63 characters removes it:

| input (`"xSteps: 20," + "a"*n`) | before | after |
|---|---|---|
| 8,000 chars | 0.143 s | 0.0024 s |
| 32,000 chars | 2.338 s | 0.0106 s |

Linear, so a line count is no longer a multiplier on anything and the aggregate bound you suggested isn't needed. Real infotext parses byte-identically (checked against an ADetailer + ControlNet + quoted-`Hashes` line; longest genuine key is 19 chars). The quoted-value branch was measured too — not quadratic, 0.0004 s. `_MAX_FIELDS_LINE` and the 32-line window stay as bounds on absurdity, with their comments corrected to say they are no longer what holds the line. There is a ratio test on the regex so a future change cannot quietly reintroduce it.

### S1 — side crossing. Fixed, `(side, "conditioning")`.

Reproduced your graph exactly; it returned `"POSITIVE"`. Embarrassing, since it is the failure the parameter exists to prevent — and my own "control" test for it asserted `!=`, which survives the swap. That test is gone (the exact-value assertions beside it bite) and the side-crossing case is now a unit test on your graph shape. Thank you for checking the rest of the walk rather than assuming it.

### S2 / S3 — settings are typed per field.

Both were the same unexamined int/str door you named. `_SETTING_FIELDS` is now a type map (`steps` int, `cfg`/`denoise` float, `sampler_name`/`scheduler` str) and `_typed_setting` takes the **field's** type, not the value's kind — so your `{"steps": "twenty", "sampler_name": 12345}` now yields `{}`. On the A1111 branch a value that is wholly a number is reported as one, so `steps` is a number on both branches; `size: "512x768"` and `sampler: "Euler a"` stay text, as the contract's name/value-pairs description says.

Worth flagging: my first cut of this had a real bug, caught by probing it rather than by reading it. `math.isfinite` **overflows** on an integer too large for a float, and raising out of the loop cost the whole settings block rather than one field. Guarded and tested.

### N1, N2 — both taken.

`workflow_structural_hashes` is popped unconditionally, before the early return, with a comment recording that it is the fifth member of a pre-existing class and that this closes only the instance the change adds. `_a1111_strengths` logs at debug, with the reason debug is the right level written next to it.

### N3 — `_picture_workflow_key` re-reads the picture. Not changed, and I'd rather say why than quietly skip it.

The handler already makes two primary-key reads (`_resolve_picture_file`, `_picture_source_origin`), so this is a third of the same shape on a dialog endpoint. Folding it in means widening `_picture_source_origin`'s select and return shape, and that function is documented tightly around one job. Happy to do it if you want the read count down, but it looked like more coupling than it buys.

### The security bound — you're right, and it's restated in both places.

"One bit per topology" was too favourable. The holder has the graph, so a 2ⁿ sweep over the LoRA-slot mark assignments recovers the set exactly; the disclosure is **one bit per LoRA slot** of that topology, about the first-filed picture, which may be outside their scope. The code comment and `backend_architecture.md` now say that, and say explicitly that an earlier version of the note claimed less — since the reason this matters is that a later reader relies on the written bound.

I have not moved `workflow_key` to owner-only: the step asks for it on this route, and it stays a filename-derived classification rather than a filename. Both docs name owner-only as the fix if you judge the bit-per-slot too much, and that call is yours, not mine.

### Verification

361 tests green across the seven touched files. Five mutations run and each went red before restoring: seed guard back to `isdigit()`, regex key run unbounded, generic-key-first walk, untyped settings, `isfinite` on the int path. `ruff check` / `format --check` clean; the generated route table still matches.

Thank you for the "verified correct" section — the `in_([])` compilation and the branch enumeration are exactly the parts I could not have asserted honestly myself.

Issue #1177, opened 2026-09-05
v1.11.1: backlog from the 1.11.0 release review


lindkvis:
Backlog from the 1.11.0 `develop` -> `main` release review (eleven chunk
reviewers, 975 files, +78,215/-9,139). Everything filed here was explicitly
graded **can wait** — the blockers went out separately as #1175 and #1176.

**Item numbers are stable.** A completed item keeps its number and moves to the
Done table rather than being removed; new items continue from 67.

---

## Status

**44 of 55 done, 9 open** (plus items 63–65 surfaced later). Safety and
security are clear. Everything still open is either routed or waiting on one
unanswered question.

The second wave landed as seven PRs — #1199 through #1205 — each one taken past
an independent adversarial review as well as CI. Three of those reviews found
defects in code that was already green and bot-approved; the findings are in
#1206.

---

### Correctness

26. **UI-graph subgraph inlining expands exponentially before its depth guard
    fires** — `services/workflow_hash.py:604`.
29. **Explicit `default_device: "cuda"` is still rewritten to `"cpu"`** on a
    failed GPU check. The owner asked for a device and silently got another.
65. **`host` is never reset when the external listener is disabled** —
    `electron/src/main.ts`, `writeServerSettings`. Turning remote access on sets
    `cfg.host = '0.0.0.0'`; turning it off leaves it. Confirmed still true on
    `develop`. The Electron backend ignores `host` unless
    `external_server_enabled` is set, so there is no exposure in the desktop
    app — but the standalone server does bind `config["host"]`.

### Performance

48. **Views publish loads every member `Picture` ORM row into memory** —
    `services/views_service.py`.

### Cleanup

54. **`FolderEditor`'s reference-create branch and the commit service's
    reference mode are unreachable from the frontend.** Every create-mode
    `openReferenceFolderEditor()` call routes to the mapping wizard, which
    hard-codes `mode: "local_import"`. Deliberate — adding an existing folder is
    "Add a library" now — so this is dead-path tidying, not a regression.
63. **`setup:inspect` takes an arbitrary path** — `electron/src/main.ts`.
    **Narrowed by #1201**: the channel is now gated to the wizard page, so what
    is left is the arbitrary-path read from within setup itself. Read-only, and
    an existence oracle that also returns a picture count, a byte total and free
    space.

### Blocked on something this machine does not have

4. **Purge repair repoints `file_path` only** —
[...]

PR #1429, opened 2026-09-18
Workflows & Recipes B3: GET /workflows/cards and workflow card detail

### LINKED ISSUE #1395: Workflows & Recipes B3: GET /workflows/cards and workflow detail

## Depends on

|---|---|---|

## What

- `GET /workflows/cards`, alongside the old routes: cards, effective stacks, aggregates, covers, one-off and hidden counts. B9 later moves it onto `GET /workflows`.
- `GET /workflows/{key}`: detail, with defaults and their provenance.
- `GET /workflows/{key}/pictures`.
- Effective stack order: a manual assignment, then unstacked, then the automatic group. Rows for keys that left a group are ignored, and new members are appended in cover-rank order.
- **Computed per request in the vault**, with no aggregate table:
  - one `GROUP BY workflow_structural_hash` for count, rated count, Σscore and last used, mapped to cards in memory;
  - cover rank as the Bayesian mean `(C·m + Σscore)/(C + n_rated)` with C=5, then picture count;
  - cover pictures from one `ROW_NUMBER()` window (top 3 by score, smart score, recency);
  - one-off count (fewer than 3 pictures, no rating, no saved recipe, not imported) and hidden count.
- Defaults: the mode of each featured parameter over instance documents of pictures rated 4★ and up, falling back to all pictures. Provenance reads "from your best pictures" until edited.
- "Differs by" per member via `workflow_identity.differs_by`. A stack card shows the union.
- Deferred: the "unusual default" fact chip.

## Acceptance

- Every route is `OWNER_ONLY` in `ROUTE_POLICIES` and the coverage matrix, with both-direction authz tests.
- **Timing measured on the dogfood copy.** The budget is about three queries and well under 100 ms. State the number in the PR.
- Assertions go in `tests/test_workflows_api.py` (module-scoped server). Route table regenerated with `scripts/render_backend_architecture.py`, and the integration doc §2 updated.

lindkvis:
Closes #1395. v1.12 Workflows & Recipes, step B3.

Three owner-only read routes for the workflow **card** — the topology plus its
non-LoRA models plus the LoRA slots marked structural, so swapping a character
LoRA keeps the same card. The shipped topology list is untouched and keeps
working; F1b swaps the route later.

| Route | What it answers |
|---|---|
| `GET /workflows/cards` | The grid: cards in cover-rank order, one card per stack, and counts of the one-offs and hidden cards it left out |
| `GET /workflows/cards/{workflow_key}` | One card: its variants, and the value each featured parameter starts from with where that value came from |
| `GET /workflows/cards/{workflow_key}/pictures` | Picture ids for one card, newest first |

## The payload is the one the frontend already documents

`frontend/src/utils/workflowCard.js` is merged and writes the card down at the
top of the file; `docs/frontend_architecture.md` promises `GET /workflows/cards`
"needs no mapping layer". The first draft of this branch invented a different
shape — `workflow_key`/`pictures`/`cover_picture_ids`, a sibling `stacks[]`
array — which would have left `WorkflowCard.vue` reading `picture_count`,
`rating` and `stack_size` off a payload that had none of them, and
`InfoPopover.vue` keying a `v-for` on an `undefined` label. That was caught in
review and the route now serves the documented document: `key`, `name`, `type`,
`imported`, `models`, `loras`, `differs_by`, `picture_count`, `rating`,
`covers`, `stack_size`, `saved_recipe_count`, `defaults`, with `mark` carrying
B1's own `structural`/`recipe` vocabulary rather than a translation of it. It
adds `topology_hash`, `variant_count`, `member_keys` and `rank` beside them,
which a caller that only knows the document ignores.

Verified by running the shipped helpers (`isStack`, `checkpointModel`,
`loraChips`, `factChips`, `ratingLabel`, `cardAccessibleName`) over a real
payload: all nine checks pass, and the accessible name comes out as
"stack of 2 workflows, checkpoint …, LoRAs: recipe LoRA slot, differs by: other
checkpoint, 2 nodes differ, 4 pictures, rated 4.5 of 5".

## What it computes, and where

**No aggregate table.** Two vault queries — one `GROUP BY
workflow_structural_hash` and one `ROW_NUMBER()` window — joined in memory to
the hub's card rows on a structural hash, plus a third vault query only on a
library where somebody has actually chosen a cover.

- **Cover rank** is the Bayesian mean `(C·m + Σscore)/(C + n_rated)`, `C = 5`,
  with *m* the library's own mean rating. Picture count breaks ties.
- **`rating` is not `rank`.** `rating` is the plain mean of the stars a card
  has, `null` when it has none; `rank` is the smoothed number the grid is
  ordered by. Showing the rank as the rating tells somebody their never-rated
  workflow is rated 3.7.
- **Cover strip**: a card's best three, picked in memory out of the best three
  of each variant. The in-memory order mirrors the window's `NULLS LAST`
  exactly — a missing score sorts *below* the repo's `-1.0` failed-metric
  sentinel, which reading it as `0.0` gets backwards. Served as thumbnail URLs
  with the same cache-buster the grid's own tiles use. An owner-chosen cover
  (stored by `pixel_sha`) leads the strip and falls back when it is binned.
- **Effective stacks**: a manual assignment, then an unstacking, then the
  automatic group by `core_hash`. Positions are filed under the core hash the
  row was written against, and *that keying is what ignores a card which has
  left its group* — it now groups under a different core hash, so its stored
  position is looked up in a dict it is not in. The grid draws one card per
  stack, the cover, wearing `stack_size` and the union of `differs_by`.
- **Defaults** are the mode of each featured parameter over the instance
  documents of pictures rated 4★ and up, falling back to every picture.
  Addressed by `(slot_label, input_name)`, never by node id. Capped at the
  newest 200 runs: an instance keys on the seed, so an uncapped read parses one
  stored graph per generated picture.
- The "unusual default" fact chip is deferred, as the issue says.

## Timing

Measured on a copy of the dogfood hub and vault (13,166 kept pictures, 8,808
carrying a variant, 963 rated; 629 variants → 245 cards):

- `GET /workflows/cards`: **median 77 ms** — 67 drawn cards, 33 stacks, 112
  one-offs.
- `GET /workflows/cards/{key}`: median 78 ms.

Roughly half of it is `describe_differences` reducing a graph per stacked card;
reducing each document twice per comparison instead of four times
(`workflow_identity._slots`) took that from 45 ms to 35 ms.

## Three deliberate deviations

1. **The paths.** The issue asks for `GET /workflows/{key}` and
   `GET /workflows/{key}/pictures`. The second would be permanently shadowed by
   the mounted `/workflows/{topology_hash}/pictures`; the detail route alone
   would have fitted, but splitting three sibling routes across two prefixes to
   save one segment reads worse than keeping them together. B9 moves the
   prefix, not the shape.
2. **`saved_recipe_count` is always 0** and the one-off predicate's fourth
   clause, "no saved recipe", is not implemented: saved recipes have no table
   yet. The predicate carries a `ponytail:` comment naming where it goes.
3. **The diff is larger than the ~600-line guide.** Most of it is the 900-line
   test file and this codebase's comment density. It is one step and merges
   alone.

Known and accepted: hiding one member of a two-card stack leaves the other
standing alone with `stack_size: 1`. The `workflow_stack_member` rows are
untouched and unhiding restores it — the grid is reporting what it drew.

## Authorization

All three are `OWNER_ONLY` in `ROUTE_POLICIES` and in
`docs/authz-coverage-matrix.md` (334 declared, +3 `owner_only`, nothing
retargeted). They read every non-deleted picture in the vault, so they disclose
the library's size one card at a time and a thumbnail URL for every card's best
three; the detail adds the parameter values the owner's best pictures were made
with. Both directions are asserted: the owner 200s, every resource-scoped share
token is 403'd with a live in-scope positive control beside it.

## Tests

`tests/test_workflows_api.py`, in the module's existing module-scoped server.

Every new assertion was mutation-checked: **24 mutants, all killed**. Six
survived the first pass and each one was a test passing for the wrong reason,
fixed in the fixture rather than papered over:

- the Bayesian prior was untested because the fixture's library mean was exactly
  4.0 and the test wrote `4.0` as a literal — the ratings are now 5, 4, 0 and 2,
  so the mean is 11/3 and equals no card's own mean;
- four soft-delete filters were dead because the only binned picture sat on a
  card the one-off clause removed anyway — two binned pictures now sit on cards
  that *are* drawn, rated above everything kept beside them;
- the hidden-card test passed because the hidden card was also a one-off;
- `COVER_DEPTH` was untested with only three pictures on the card;
- the owner-chosen-cover path had no coverage at all;
- the ⓘ label was never duplicated, so it needed a second sampler slot.

One survivor turned out to be dead code rather than a coverage gap — an explicit
"has this card left its group" comparison that the position keying already made
unreachable — and was removed.


lindkvis:
Shard 1 was red on `test_every_untemplated_owner_class_get_is_on_the_read_blocked_belt`: `GET /api/v1/workflows/cards` is the first **untemplated** owner-class GET this branch adds, and the belt in `pixlstash/auth.py` owes one entry per such route.

Reproduced locally with the gate's own flags and sharding (`--force-cpu --ci-shard 1/8` over the job's file list): `1 failed, 601 passed`, the same single failure and nothing else.

Fixed in d8be62f9 by adding the path to `READ_BLOCKED_GET_PATHS`. Two notes for the reviewer:

- The route was **already refused at runtime** by the existing `/api/v1/workflows/` entry in `READ_BLOCKED_GET_PREFIXES`, so this is not a hole that was open. The entry is owed because the rule that frozenset is checked against is arithmetic over the registry — every untemplated owner-class GET — rather than "every one a prefix does not already happen to catch". Excepting prefix-covered paths would put the judgement back in, which is the thing that left `/insights` and `/moves/pending` off the belt in #1177. The added comment says so.
- The two templated card routes need nothing: `/api/v1/workflows/` already covers them, and `test_read_blocked_get_prefixes_cover_only_owner_class_gets` holds that the prefix swallows only owner-class GETs.

`test_authz_host_capability_16_3.py` and `test_architecture_guardrails.py` pass (86), and the assertion was confirmed to bite by removing the entry again and watching it go red.

lindkvis:
## Review: approve with changes

Backend review plus an independent adversarial authz/security pass. Nothing merge-blocking outright; four items belong on this branch. Findings are split into **verified by running code** and **reasoning only**.

Two things first, because they set the tone in opposite directions.

**The soft-delete work is the best part of the change.** Five new query sites, five filter, and each of the five is individually mutation-killed by a *different* named test (9 of 9 module-wide). The PR's claim that four filters were dead in the first pass and are now live is true and verifiable — `busy_binned.png` and `forgotten_binned.png` are doing real work, the latter flipping its card's `provenance`. That is exactly the discipline CLAUDE.md asks for.

**But the headline test claim does not hold.** Against "24 mutants, all killed": **four load-bearing mutants survive.** That is believable for the set chosen — the claim is not dishonest — but the set has holes, and one of them sits on the invariant this PR itself calls easiest to get subtly wrong.

---

### Should-fix

**C1 — the "newest 200 runs" cap has no coverage at all. `services/workflow_library_service.py:393-425`. VERIFIED.**

Three independent survivors, 101 tests green under each:

| mutation | result |
|---|---|
| `DEFAULT_SAMPLE` 200 → 1 | survives |
| `.limit(limit)` → `.limit(10000)` | survives |
| `order_by(….desc())` → `.asc()` | survives |

The **ordering** is the sharper loss. Flipped, a card run 5,000 times answers with the mode over its *oldest* 200 runs — defaults that drift backwards in time and never converge on what the owner actually does now. One fixture card with 4 instances plus a monkeypatched `DEFAULT_SAMPLE` kills all three.

**C2 — the cover-ordering rule is right, but its `score` half is untested. `services/workflow_card_service.py:218`. VERIFIED.**

To be clear, since this one is easy to misread: **the shipped code is correct.** `-inf if candidate.score is None else candidate.score` is the right rule, and the docstring above it already explains precisely why `or 0.0` would be wrong, citing CLAUDE.md's `-1.0` failed-metric convention. The defect is coverage, not behaviour.

Replacing it with `candidate.score or 0` leaves **all 101 tests green**. The fixture proves the NULL-below-`-1.0` rule for `smart_score` only (`busy_three` = -1.0, `busy_four` = None, so that mutant does die), but **no kept picture carrying a structural hash has `score IS NULL`** — the drawn cards are 5/4/0/0/2, and every NULL-scored row is either soft-deleted or has no workflow. Relatedly, the test docstring at `tests/test_workflows_api.py:3063` claims "a cleared 0 sorts above an unrated picture (NULL is last, not zero)" — the fixture contains no such picture, so that sentence currently asserts nothing. Same defect class as the six the PR fixed in its first pass.

The failing case was constructed and run: add one kept picture on `BUSY_RECIPE_A` with `score=None, smart_score=0.9` — unrated but quality-scored, the commonest kind of picture in a real library. Correct code keeps the strip as `[busy_one, busy_two, busy_three]`; under `score or 0` it reads as 0, ties the cleared zeros, wins on `smart_score` and **takes the third cover slot**, so the in-memory strip disagrees with the SQL window that selected the candidates.

**C3 — `name` is `null` for every card the owner has not renamed, and the shipped consumers do not handle it. `routes/workflows.py:243`. VERIFIED.**

Confirmed at the schema: `name: str | None = Field(None, description="The owner's own name, if they gave one.")`, sourced from `workflow_attr.name`, which is written only on an explicit rename. So `null` is the **dominant** case, not an edge. Confirmed on a real payload dumped from the route: both cards return `"name": null`. Running the shipped helpers and templates over it:

- `WorkflowCard.vue:75` `{{ card.name }}` → **empty**, and that row is the card's only identifying text;
- `InfoPopover.vue:25` → **`aria-label="About null"`**;
- `cardAccessibleName` drops it via `.filter(Boolean)`, so the accessible name begins "stack of 2 workflows, checkpoint …" with nothing naming the card.

Worth noting: the PR body's own verification quotes that very string — *"stack of 2 workflows, checkpoint …, LoRAs: recipe LoRA slot, differs by: …"* — as evidence the nine helper checks pass. The nameless card is visible in the quoted evidence. This needs a decision either way: serve a derived fallback, or state in `integration_architecture.md` §2.3 that the client owns it and file a frontend issue.

**C4 — a stack member's detail carries `differs_by` while declaring `stack_size: 1`. `services/workflow_card_service.py:399` vs `:438`. VERIFIED.**

Real payload for `GET /workflows/cards/{FORGOTTEN}`: `stack_size=1`, `member_keys=[]`, `differs_by=["other checkpoint", "2 nodes differ"]`. `factChips` (`frontend/src/utils/workflowCard.js:75`) takes the **non-stack** branch on `stack_size: 1`, which still spreads `card.differs_by` — running it gives `["other checkpoint", "2 nodes differ", "img2img"]`, and the accessible name reads `facts: other checkpoint, 2 nodes differ, img2img`. `WorkflowCard.vue:92` also omits the "differs by" prefix for a non-stack, so those chips arrive with no context, and "other checkpoint" as a standalone *fact* is a difference from a cover the payload never names.

`test_a_stack_shows_the_union_…:3199` asserts this deliberately — it was assessed against the payload rather than against the consumer. This also compounds the PR's own "known and accepted" hidden-member case: fixing C4 fixes that too, since the surviving card currently keeps difference chips against a cover no longer on screen.

C3 and C4 both sit on the exact contract seam this PR exists to honour, which is why I would land them here rather than defer.

---

### Nits

- **Tie-break comment says the opposite of the code. `workflow_card_service.py:632`. VERIFIED.** The comment reads "on a tie the value that sorts first"; `max(key=(count, str(value)))` picks the one that sorts **last** (measured: `{20,30}→30`, `{9,30}→9`, `{'a','b'}→'b'`). Deterministic and asserted — the mutant dies — so only the comment is wrong.
- **`differs_by` re-reduces the stack cover once per member** (`:389`). Instrumented `_reduce` = 2 for a stack of 2, so a stack of N reduces the cover N−1 times. The seam already built for `_slots` would hoist it.
- **Module docstring says "six hub reads"**; the instrumented count is **8**. (The vault claim is accurate — see below.)
- **Import block out of order**, `routes/workflows.py:53-55`: `ImageUtils` sits between the `hub.*` and `services.*` groups. Ruff-clean, since `I` is not enabled.
- **`pictures.get(chosen.get(key, ""))`** at `:534` — a `workflow_cover` row with `pixel_sha=''` would give every card the same cover. **Reasoning only, not reproduced.**
- The mode is computed over *distinct instances*, not pictures, so a 50-picture same-seed batch counts once. Defensible, but worth a docstring line.

---

- **`rating` vs `rank` is clean everywhere.** `rating` never sorts, `rank` is never presented as a rating; the `rating → self.rank` mutant kills 2 tests.
- **The comparator mirrors the SQL window.** `-inf` / `datetime.min` / `id desc` match `nullslast(desc())` ×3 plus `id.desc()`, and `-1.0 > -inf` so the failed-metric sentinel outranks a missing score in **both** passes. The chosen-cover path resolves by `pixel_sha`, filters deleted, leads the strip and falls back when binned — all mutation-killed. (Subject to C2.) Removing `nullslast` from the SQL survives, but that is an *equivalent* mutant — SQLite orders NULL last under `DESC` anyway — not a gap.
- **The removed dead code was genuinely dead, verified rather than accepted.** Walked the leaver, joiner and returner cases, then mutated the keying away (positions flattened into one bucket) and `test_a_member_row_for_a_card_that_left_the_group_is_ignored` **went red**. One residual: two `auto` rows sharing a `core_hash` merge last-wins — deterministic, untested, and unreachable until the write route lands.
- **Defaults are otherwise sound.** The fallback is tested (`BEST_SCORE` 4→2 dies), the cap **cannot** be polluted by unrelated imports (narrowed by `variants.in_()` before the limit), and addressing is by `(slot_label, input_name)` throughout, with `rounds=None` matching `slots`.
- **No N+1 anywhere.** Instrumented per route: grid = 8 hub reads + 1 vault read-task (2 SQL), rising to 2 vault tasks once a cover is chosen; detail = 15 + 3; pictures = 1 + 1 — constant in card count. `describe_differences` batches one `variant_documents` call for all members (`:375`); `slot_marks` and `asset_names` are batched and chunked. The PR's "two vault queries plus a third only when a cover was chosen" is **accurate**.
- **Exceptions**: 5 clauses, all narrow, all logged with key/hash plus the exception and a stated consequence. No bare `except`, no `pass`, no `# noqa`.
- **`ruff check` and `ruff format --check`** both clean on `pixlstash tests scripts` (750 files).
- **The d8be62f CI fix is complete**, verified by running the test rather than reading it. The belt test derives `untemplated = {path without "{"}`, so only `/api/v1/workflows/cards` was owed an entry; both templated routes are covered by the existing `/api/v1/workflows/` entry in `READ_BLOCKED_GET_PREFIXES`, which the same test's second half checks. 183 passed across `test_authz_host_capability_16_3.py`, `test_authz_gate_step3.py` and `test_ci_shards.py`.
- **Authz is clean and the negatives are real.** Three routes `OWNER_ONLY`, matrix arithmetic complete (331+3 = 334; owner_only 152+3 = 155), guardrails green, no inline scope ladder in any handler (`ensure_secure_when_required` is transport, not object scope), and both directions tested with a live in-scope positive control. Independently mutation-proved: removing the gate's owner guard (`require_unscoped_owner` → `return`, the single non-docstring occurrence) turned `test_the_insertion_preview_is_owner_only` **red** (200 instead of 403). `assert_real_route` does genuine route resolution, so the negatives are not vacuous 403s from the pre-routing middleware — which is the specific silent coverage loss CLAUDE.md warns about.
- **Test hygiene follows the module rules**: `fresh_library` re-seeds hub *and* vault per test so the destructive tests do not leak, finders are detached and re-asserted gone, the hub wipe orders children before parents, and assertions name cards rather than counting. No new test file, so nothing owed to `ci.yml`.
- Secret/private-information scan clean on added lines.

### The four declared deviations

1. **Paths under `/workflows/cards/`** — defensible and correct. `/workflows/{key}/pictures` really would be totally and silently shadowed, since both segments are 64-hex.
2. **`saved_recipe_count` always 0** — defensible. `InfoPopover.vue:67` always draws the row, so `0` is honest where absent would not be, and the `ponytail:` comment names where clause 4 goes. Accepted.
3. **Diff size** — accepted; 847 of 2,490 lines are tests, the rest is this repo's comment density, and it merges alone.
4. **"Known and accepted" hidden-member `stack_size: 1`** — defensible as stated, **but** it compounds C4 above: the surviving card keeps difference chips against a cover no longer on screen. Fixing C4 fixes this one.

---

### Before merge

**Resolve the registry/matrix conflict with #1427.** Both PRs insert new rows at the same anchor in `pixlstash/authz/registry.py` and `docs/authz-coverage-matrix.md`. Replaying the merges in a sandbox: #1428 merges clean, then this one clean, then #1427 conflicts in exactly those two files — loud, not a silent undeclared-route hole. A union resolution gives **339 declared**, and on that merged tree the guardrails pass (86) along with all three branches' own suites (248).

**One trap in that resolution:** `test_coverage_matrix_document_matches_the_registry` parses only the row tables, **not** the prose "CURRENT: N declared" header bullets. After a union merge both bullets (336 and 334) survive, neither equals the true 339, and **CI still passes**. Whoever lands second must union-resolve *and* re-derive the header count by hand.

Suggested disposition: land C1–C4 here (C2 is a paste-ready test, C1 is one test, C3 and C4 are on the contract seam). The nits are cheap while the branch is open.

---

lindkvis:
Thanks — the four should-fix items and all six nits are in `42a4ea91`. Every claim below was re-measured rather than reasoned about, and I've re-run the mutants you named.

## C1 — the cap had no coverage. Fixed.

You were right that all three survived. The fixture's forgotten card now has four runs instead of one: three at `steps: 20` and a newer one at `77`. So the mode over every run is 20 and the mode over the newest one is 77, which makes the cap, its `LIMIT` and its `ORDER BY` each observable on their own. `test_defaults_read_the_newest_runs_and_stop_at_the_cap` monkeypatches `DEFAULT_SAMPLE` to 1 and asserts 77; the sibling that reads the same card uncapped still asserts 20.

All three mutants now die: `DEFAULT_SAMPLE 200→1`, `.limit(limit)→.limit(10000)`, `.desc()→.asc()`.

## C2 — the `score` half was untested. Fixed.

Confirmed exactly as you described: no kept picture on a drawn card had `score IS NULL`, so the docstring sentence asserted nothing. Added `busy_unrated.png` — unrated, `smart_score: 0.9`, on `BUSY_RECIPE_A`. Read as a zero it ties the two cleared zeros, wins on its smart score and takes the third cover slot, which is what your constructed case predicted. `candidate.score or 0` now fails two tests. The docstring was rewritten to name the picture that proves each half rather than describing one the fixture lacked.

## C3 — `name` was null for every un-renamed card. Fixed in the backend.

Decided rather than deferred: the server now resolves it, because `frontend_architecture.md` promises no mapping layer and asking the client to derive a display name *is* a mapping layer. Order is the owner's name → the workflow file that runs the card, without its extension → `Untitled workflow`.

The file name is the interesting rung — it is what people actually call their workflows, and it is the one identifying string not already on the card, since the checkpoint has its own row and the type its own chip. `card_index`'s existing left join now carries `MIN(workflow_name)` so a card with several files names itself the same way twice.

**The stand-in wording is a design call and I am not attached to it** — `Untitled workflow` is what 133 un-named, un-imported cards would all read as. If the F1 screen wants something else, that is one line. What is not negotiable is that the field stops being null, since `{{ card.name }}` renders empty and `InfoPopover` puts it in an `aria-label`.

## C4 — chips without the size that explains them. Fixed.

`stack_size` and `member_keys` are now set on **every** member of a stack, not only its cover. The grid is unchanged — it still draws the cover alone — but a member opened on its own reports `stack_size: 2` and names the stack, so `factChips` takes the stack branch and the chips keep their "differs by" context.

Two tests hold both directions: a member's detail agrees with its cover about the same stack, and after unstacking both cards report `stack_size: 1` with `differs_by: []`. As you predicted, this also settles the "known and accepted" hidden-member case — chips and size can no longer disagree.

## Nits

- **Tie-break comment** — you are right and I had it backwards. Measured `{20,30}→30`, `{9,30}→9`, `{'a','b'}→'b'`. Comment now says "the value whose text sorts LAST", with the `9 over 30` example, and notes that any total order would do since what matters is that there is one.
- **Cover re-reduced per member** — done, using the same seam as `_slots`: `reduce_stored_document` + `differs_by_reduced`, with `differs_by` unchanged as the document-taking wrapper. Instrumented on the dogfood copy: **132 → 99 reductions** (one per stacked card instead of one per comparison), and the grid's median goes 77 → **75 ms**. Worth doing beyond the 2 ms because the old cost grew with *stack size*, and `STRIP_LORAS_FOR_STACKS` is deliberately generous about what groups together.
- **"six hub reads"** — instrumented: **8**. Corrected, along with the timing figure in the same docstring.
- **Import order** — fixed; `find_card` joins the other `hub.*` imports and `ImageUtils` goes after `services.*`.
- **Empty `pixel_sha`** — reproduced, since you flagged it as reasoning-only. It needs both a blank cover row and a picture stored with a blank sha; with both, every card in the grid took that picture as its cover. Guarded, and `test_an_empty_chosen_cover_covers_nothing_rather_than_everything` fails without the guard.
- **Mode over distinct instances** — docstring line added saying a fifty-picture same-seed batch votes once, and why that is the intended reading.

## Verification

- **14 mutants run against the changed logic, 14 killed**, including every one you reported surviving.
- Re-ran the shipped frontend helpers over a fresh payload (cover *and* member): 12 checks pass. The accessible name now opens with the card's name instead of "stack of 2 workflows…", and the member's facts come back as `["other checkpoint", "2 nodes differ"]` with its stack context intact.
- 698 passed across the workflow, guardrail, authz and CI-shard suites; `ruff check` and `ruff format --check` clean.

## The #1427 conflict

Noted, and thank you for replaying the merges — particularly the trap that `test_coverage_matrix_document_matches_the_registry` parses only the row tables, so two stale header bullets would survive a union merge with CI green. I have not touched the registry or the matrix in this push, so the conflict is unchanged and still loud. Whoever lands second unions the rows *and* re-derives the header count by hand; I am happy to be that one if this merges first.

lindkvis:
`develop` moved (#1427 saved recipes, #1428 B5) and `951fb899` merges it in. Merge, not rebase, since the branch is pushed and reviewed.

## The two conflicts

Both were the union you predicted, and both sides' changes are independent and additive — mine adds card routes to `workflows.py`, theirs adds the whole of `recipes.py`. Neither makes the other wrong, so nothing was dropped.

- **`pixlstash/authz/registry.py`** — union, my three card routes kept with the `workflows.py` block they belong to and the `recipes.py` section after it. `len(ROUTE_POLICIES) == 339`, `owner_only` 160, exactly the figure you got in your sandbox.
- **`docs/authz-coverage-matrix.md`** — the header bullet, and **your trap was real**: a plain union leaves two `CURRENT` bullets at 334 and 336, neither equal to 339, with CI green because `test_coverage_matrix_document_matches_the_registry` parses only the row tables. Resolved by re-deriving the count from the merged tree rather than adding the two bullets up: one `CURRENT` bullet at 339, #1398's demoted to *superseded*, and my own 334 bullet dropped rather than kept beside its successor — it described a base that never existed on `develop` once #1427 landed. The new bullet says in as many words that it was re-derived by hand and why no test would have caught it otherwise.

`pixlstash/auth.py` auto-merged and kept both belt entries (`/api/v1/recipes` and `/api/v1/workflows/cards`); the matrix row tables auto-merged with all eight new rows present.

## What the merge changed beyond the conflicts

Reading `develop` rather than just resolving text: **#1398 landed the thing two of my stubs were waiting for**, so they were wrong after the merge rather than merely unspelled.

1. **`saved_recipe_count` was hardcoded to `0`** with a comment reading "No table yet (a later step)". `saved_recipe` is now a vault table keyed by `workflow_key` — my card key — and `InfoPopover` always draws that row, so a hardcoded zero does not read as "not implemented", it reads as "you have saved none". Now a real count, from one `GROUP BY workflow_key` folded into the grid's existing read task (three statements, still one round trip).
2. **The one-off predicate's fourth clause was unimplemented**, carrying a `ponytail:` comment saying saved recipes "arrive with their own step". They have. Without the clause, saving a look on a workflow with two unrated pictures folds that card into the one-off count and takes it off the grid — the owner's own recipe sitting on a card they can no longer see. The clause is now in, and the comment with it.

Both are pinned by tests, and both mutants die: hardcoding the count back to `0` fails two, dropping the clause fails one. The fixture also now wipes `saved_recipe` per test — it is keyed by card rather than by picture, so nothing else in the vault reset reached it, and a leaked recipe changed the next test's one-off count.

## Checks

- 710 passed across the workflow, saved-recipe, guardrail, authz and CI-shard suites — including `test_saved_recipes.py`, which the merge brought in.
- `test_architecture_guardrails.py` + `test_authz_host_capability_16_3.py`: 86 passed, so the registry, the matrix and the read-blocked belt all agree on the merged tree.
- `ruff check` / `ruff format --check` clean.
- Route table regenerated; `integration_architecture.md` §2.3 updated for the real `saved_recipe_count` and the four-clause one-off rule.

The PR reports `MERGEABLE` again. Not merging it — that is still yours.

PR #1431, opened 2026-09-19
Feat/apple silicon acceleration

lindkvis:
Support acceleration on Apple silicon


lindkvis (commented):
Reviewed `origin/develop...` this branch. ruff clean; the two new test files pass (105 tests) and are gated in `ci.yml`; detection verified against real torch on a CUDA box (`available_accelerator()` -> `cuda`, unaffected).

One blocker and two worth fixing.

**1. `empty_device_cache()` is called with an argument it does not take**

`tagger_plugins/clip_service.py:166` and `:252` call `empty_device_cache(previous_device)`, but `utils/vram_utils.py:167` is `def empty_device_cache() -> bool:` — no parameters. It raises `TypeError`.

The call sits inside `except RuntimeError:`, so the `except Exception:` below is a sibling and cannot catch it. It escapes `encode_image_batch` / `encode_face_crops` after the model has already been moved to CPU and `self._device` reassigned.

This is also a regression on CUDA: that fallback worked before (`torch.cuda.empty_cache()`). Nothing tests either method, which is why the suite is green.

Fix is giving the wrapper the optional argument its callee already has:

```python
def empty_device_cache(device=None) -> bool:
    return empty_accelerator_cache(device)
```

plus a test driving a fake OOM through the fallback, in both vocabularies.

**2. `resolved_device` is dead and gets persisted**

`startup_checks.py:456` sets `self._server_config["resolved_device"]`. Nothing reads it, it has no `_` prefix, and `server.py:531` calls `persist_server_config` on that dict six lines after `run()` — so it lands in the owner's config file on every accelerated boot. That is the case `server_config_io.py`'s docstring exists to prevent. Delete it, or name it `_resolved_device`.

**3. The CLIP workflow's sibling gates were not swept**

`inference/workflows/clip_embedding.py:71` and `:86` still read `== "cuda"` / `!= "cuda"`. CLIP does run on Metal after this change, so there it gets a flat uncapped batch of 128 and reports 0 MB to the reservation accounting — the same class fixed in `workflows/tagging.py`, on the memory model where the module docstring says it matters most. The CLIP per-image figures are CUDA measurements, so opening the gate needs a measured pair or the placeholder note WD14 already carries.

`face_embedding.py:89` and `description.py:259` share the spelling but are correct by accident — both models stay on CPU by documented hold. Worth a comment saying so.

**Smaller**

- `services/config_service.py:151` — `collect_vram_from_torch` is still CUDA-only, so no device-memory telemetry on Apple Silicon while the budget slider now shows a Metal-derived ceiling.
- `work_planner.py:682` — the metric is legitimately CUDA-only, but the log line reads `"torch.cuda.is_available() is False in the planner thread"`, which is the exact sentence this PR is about.
- `tagger_plugins/joycaption.py:445` still picks `cuda if available else cpu`, so JoyCaption runs on CPU on a Mac — while the new `_precision_help()` warns about device memory being system RAM, which reads as though it were on the GPU. Route it, or state the hold like Florence-2 does.
- `wd14.py:207` — dropping the Darwin `worker_count = 0` is justified by `_run_dataloader`'s timeout + `num_workers=0` retry (confirmed present), but it is an untested change on macOS's spawn start method.
- `startup_checks.py` MPS branch computes `free_mb` from `total_mb` before `if total_mb <= 0: raise`. Harmless, reads backwards.
- `VALID_DEVICE_SETTINGS` lacks `""` while `_HOST_RESOLVED_SETTINGS` has it, so a blank `default_device` resolves at runtime but hard-fails validation. Probably intended; worth a line.

**Good**

One probe table, the two drifted allowlists now one object with an identity test, both directions asserted throughout (accelerator used, no-accelerator still CPU, `--force-cpu` wins, explicit `cpu` honoured, CUDA control), real captured OOM strings next to the SQLite control, and holds documented with the measurement that justified them. The `ctx_id=-1` trap note in `face_extraction_task.py` is the kind of thing that saves someone a day.

lindkvis:
All of the above addressed in f332978e, plus the shard-3 failure.

- `empty_device_cache(device=None)` now takes the device its spill callers pass. Covered by four new tests through the real `ClipService` recovery path, MPS- and CUDA-worded; reverting the signature turns all five red.
- `resolved_device` removed.
- `clip_embedding.py` batch cap and VRAM estimate gate on `is_accelerated`, with the CUDA-measured figures marked unverified on Metal. `face_embedding.py` and `description.py` keep `== "cuda"` and now say why.
- `collect_vram_from_torch` answers on unified memory; planner log reworded; JoyCaption device pick documented; `startup_checks` guard moved above the subtraction; the `""` / `VALID_DEVICE_SETTINGS` difference is noted at the constant.
- `test_every_cuda_ort_session_site_is_built_from_the_budget` (shard 3) listed WD14's three inline provider sites, which the ladder collapsed into one. Repaired to check the invariant where it now lives - ladder picks the providers, is handed the budget's options, puts them on the CUDA provider. Both halves mutation-checked.
- Added `changelog.d/1431.md`; the branch had no fragment and this is user-visible.

Not done: no fragment-level change to the WD14 `num_workers` decision on macOS, and the CoreML/MPS figures are still the ones measured on the 8 GB machine.

PR #1432, opened 2026-09-19
Workflows & Recipes B4: workflow writes, stack writes, CHANGED_WORKFLOWS event

### LINKED ISSUE #1396: Workflows & Recipes B4: workflow writes, stack writes, CHANGED_WORKFLOWS event

## Depends on

|---|---|---|
| #1395 (B3) | yes | `git grep -q '/workflows/cards' origin/develop -- pixlstash/authz/registry.py` |

## What

- `PATCH /workflows/{key}` (name, notes, hidden); `PUT /workflows/{key}/slots`, `/defaults`, `/pins`, `/inputs`.
- **Slot-mark flip with carry-over in the same transaction.** On a split every new key inherits the user attributes; on a merge the member with the most pictures wins.
- Stack writes: `POST /workflows/stacks` (stack together; merging keeps the first-selected stack's name and cover), `PUT /workflows/stacks/{id}/order`, `POST /workflows/stacks/{id}/unstack`, `POST /workflows/{key}/unstack`. A stack left with one member dissolves. A stack whose members are all hidden is hidden.
- `EventType.CHANGED_WORKFLOWS`: a "look again" signal carrying keys, reason (`imported|changed|stacks|recipes`) and the originating client id. Emitted by the inbox, import and every write above.

## Acceptance

- Stack overrides and attributes survive a `CORE_VERSION` bump and a mark flip (split and merge carry-over), shown failing under mutation.
- All routes `OWNER_ONLY` in the registry and coverage matrix, with both-direction authz tests. Integration doc §8 (events) and route table updated.

lindkvis:
Closes #1396. v1.12 Workflows & Recipes, step B4: the write side of the
workflow cards B3 shipped the reads for.

## What

- `PATCH /workflows/{key}` (name, notes, hidden), and `PUT /workflows/{key}/slots`,
  `/defaults`, `/pins`, `/inputs`.
- Stack writes: `POST /workflows/stacks`, `PUT /workflows/stacks/{id}/order`,
  `POST /workflows/stacks/{id}/unstack`, `POST /workflows/{key}/unstack`.
  A stack left with one member dissolves.
- `EventType.CHANGED_WORKFLOWS` — a "look again" signal carrying the keys, a
  reason (`imported|changed|stacks|recipes`) and the originating tab's client
  id. Emitted by the watched inbox, the import, every write above, and the
  saved-recipe writes.
- New: `hub/workflow_card_writes.py` (the only module that writes card rows)
  and `services/workflow_events.py` (one emitter, so the five emit sites agree
  about the envelope).

Nothing user-visible yet: the shipped Workflows shelf is untouched and the card
grid still has no screen (F1b).

## The one that needed thinking about

`PUT /slots` changes identity. A LoRA slot's mark decides whether that LoRA
reaches the card key, so flipping one re-keys **every variant of the topology**
— one card splits into several, or several merge into one — and the re-key and
the carry-over of everything the owner said about those cards are one
transaction. A split **copies** the attributes to every successor; a merge
gives them to the member with the most pictures, ties broken on the key so the
winner does not depend on row order.

**It also carries the saved recipes**, which is the migration
`db_models/saved_recipe.py` names this change as owing: those are *vault* rows
keyed on the card key, and a saved recipe is authored — unlike a hub row it
cannot be rebuilt from anything. Left behind it is addressed by a key no
variant carries and its workflow's tab silently stops listing it. That write
is a second database, so it cannot be inside the hub's transaction; it is the
first thing after it, and it is logged.

Two cases that look like edge cases and are not: a variant whose stored
document will not parse keeps the key it is on, so its card survives even when
a sibling moved; and `workflow_file.workflow_key` moves with it, or an imported
file stops saying it runs the card.

## Shown failing under mutation

Restored after each:

| Mutation | Went red |
|---|---|
| `_copy_rows` returns early (the carry-over does nothing) | the split and the merge tests |
| merge winner picked by fewest pictures | `..._keeps_the_name_of_the_one_with_most_pictures` |
| successor order inverted (`SlotMarkResult.key` is the small half) | `..._sends_the_caller_to_the_half_with_the_pictures` |
| `workflow_cover` + `workflow_unstacked` dropped from `_KEYED_TABLES` | `..._carries_everything_the_owner_said_and_the_file_with_it` |
| the `workflow_file` re-key removed | same |
| the saved-recipe re-key does nothing | `..._takes_the_saved_recipes_with_it` |
| `unstack_card` writes no `workflow_unstacked` row | `..._survive_a_regrouping`, `..._dissolves_a_stack_of_one` |
| `_drop_thin_stacks` returns early | `..._dissolves_a_stack_of_one` |
| `surviving` is the new keys alone | `..._will_not_reduce_keeps_its_card_and_its_attributes` |
| `set_stack_order` leaves a key in its old stack | `..._cannot_leave_a_card_in_two_of_them` |
| the whole `CHANGED_WORKFLOWS` wire branch disabled | `..._reaches_a_client_whatever_its_grid_filters` |
| the `reason` allowlist emptied | `..._goes_out_as_its_own_frame` |
| the duplicate-address guard removed | `..._twice_is_refused_rather_than_a_500` |

## Answering the pre-push adversarial pass

It found no secrets and nothing private. Of what else it found, these were real
and are fixed in this PR: the saved-recipe orphaning above; `set_stack_order`
deleting by `stack_id` only, which left a card a member of two stacks at once;
`PUT .../order` minting a stack for an id nothing holds (now 404, like its
`unstack` sibling) and silently dropping members omitted from the list (now 400
— it is a complete ordered list or nothing); a split copying
`workflow_stack_member` verbatim, which put two cards on one `position` and
moved the stack's cover to whichever digest sorted first rather than to the
half with the pictures; and `/defaults` and `/inputs` answering 500 rather than
422 on a body naming one parameter twice. Its four "unasserted behaviour"
findings are the four new mutation rows above. The doc errors it named
(`WsBroadcasterMixin.WORKFLOW_CHANGE_REASONS`, which is module level; the
`workflows_changed` row filed under events that carry no payload; the §8
behaviour column reading as shipped when nothing listens yet) are corrected,
and `AUTO_STACK_PREFIX` now has one home that the producer, the reader and the
writer all take it from.

**Two I disagree with**, left as they are:

- *A stack thinning to one member should keep the survivor's "out of its group"
  row.* `stack_together` deliberately clears that row — stacking a card by hand
  is the owner reversing the decision to keep it out — so after that gesture
  there is no standing decision for the dissolve to preserve. Reinstating one
  would resurrect a decision the owner already took back.
- *The authz negatives are verb-belt only.* True, and the test says so in as
  many words. It is not closeable with the credentials the product mints:
  `create_token` allowlists `ALL` and `READ`, and a resource-scoped token
  **must** be `READ` (`auth.py`), so no mintable credential both carries a
  scope and survives the non-GET belt to reach the gate on a write. The gate's
  runtime `OWNER_ONLY` enforcement is proved centrally on GETs, where a READ
  token does reach it (`tests/test_authz_gate_step3.py`); per route, the
  declaration is pinned. That is the same coverage every other owner-only write
  in this repo has, and closing it properly means a mintable write-scoped token,
  which is a change to the token model and not to this step.

## Docs

Integration §8 (the `workflows_changed` frame) and §2.3 (the nine mutators, and
the "no mutators" line it used to carry), `backend_architecture.md` §the
Workflows view (the write half, the re-keying, the stack rules) plus its
regenerated route and event tables, the coverage matrix, and the
`routes/workflows.py` docstring that said "Nothing here mutates".

## On the diff size

Over the ~600-line guidance, and I did not find a way to cut it that was not
just cutting function: nine routes, a re-keying transaction with both carry-over
directions and a cross-database migration, and the tests the acceptance asks
for. Roughly half the diff is tests and docs.


lindkvis (commented):
Read the whole diff. ruff clean; the gated files pass here (335 tests across `test_workflows_api.py`, `test_workflow_io.py`, `test_websocket_auth.py`, `test_architecture_guardrails.py`, `test_authz_gate_step3.py`, `test_openapi_response_schemas.py` — the one failure is the private-address guardrail firing on local `electron/release` and `node_modules` build output, identical on `origin/develop`).

The authz half is in good order: nine routes, nine registry entries, nine matrix rows, and the negative test is honest about measuring the verb belt rather than the gate — live-token control, `assert_real_route` on every path, positive control next to it. I agree with both of your pushbacks; the token-model one in particular is not this step's to close.

One real defect, found by following the asymmetry between how the hub half and the recipe half treat a card that survives a flip. It is reproducible and I have put the repro on the thread. The other three are smaller.

pixlstash/services/saved_recipe_service.py
```diff
@@ -183,6 +183,49 @@ def list_in_session(
...
+    the one holding most of the card's pictures. The row names a workflow and
+    not a variant, so which half of a split it was saved from is not recorded
+    anywhere and cannot be recovered - sending it to the biggest is a choice,
+    and the alternatives (copying it to every successor, or dropping it) are
+    both worse: one invents recipes the owner never saved, the other is the
+    silent loss this function exists to close.
+    """
+    moved_count = 0
+    for old_key, successors in moved.items():
+        if not successors or successors[0] == old_key:
```
lindkvis: **A saved recipe leaves a card that is still there.**

`moved` is not the same set as "cards that went away". `_rekey_variants` knows the difference — it builds `surviving` at `workflow_card_writes.py:335` for exactly this reason — but `moved` is built from `successors`, which only drops the `news == {old_key}` case. A card can survive while still appearing as a key in `moved`: the won't-reduce branch leaves a variant on the old key, its sibling moves, and the old key is then in `moved` with a successor that is not itself.

`successors[0] == old_key` does not catch that, so every recipe on a live card is moved off it.

Reproduced against this branch — the PR's own won't-reduce fixture, plus an Unstack so the recipe list cannot expand across the stack:

```
POST /workflows/{merged}/unstack        -> 200
POST /recipes   {workflow_key: merged}  -> 201, id 1
(corrupt FLIP_RECIPE_B's document, as test_a_variant_that_will_not_reduce... does)
PUT  /workflows/{merged}/slots          -> 200
GET  /workflows/cards/{merged}          -> 200   # the card is still there
GET  /recipes?workflow_key={merged}     -> []    # its recipes are not
GET  /recipes?workflow_key={new key}    -> [1]
```

Without the Unstack it is invisible: both halves share a core hash, so they are auto-stacked and `/recipes` expands across the stack. One Unstack — a gesture this PR ships the route for — and the recipe is gone from the tab it was saved on. That is the failure this function's docstring says it exists to close, arrived at from the other side.

It is also an asymmetry with the hub half, which is the tell: `_rekey_variants` **copies** attributes to every successor and keeps the old key's rows when it survives (`test_a_variant_that_will_not_reduce_keeps_its_card_and_its_attributes` asserts exactly that), while this **moves**. On a card that is gone, moving is right and the docstring's argument holds. On a card that is still there it has no argument behind it — leaving them is the obvious answer.

Cheapest fix is to stop lying in `moved`: have `_rekey_variants` include the old key among its own successors when it is in `surviving`, then `old_key in successors` here is the whole condition. Worth a test in the shape of the repro above, and worth checking `SlotMarkResult.key` against the same case while you are there.

pixlstash/routes/workflows.py
```diff
@@ -675,4 +909,350 @@ def list_card_pictures(
...
+        moved = flip_slot_marks(
+            hub,
+            card.topology_hash,
+            payload.marks,
+            read_variant_picture_counts(server.vault),
+        )
+        # The migration `db_models/saved_recipe.py` says a re-keying owes this
+        # table. A second database, so it cannot be in the hub's transaction;
+        # it is the first thing after it, and it is logged.
+        rekeyed = saved_recipe_service.rekey_recipes(server.vault, moved)
```
lindkvis: The success path logs; the failure path does not, and that is the one the comment above promises to cover ("it is the first thing after it, and it is logged", "the log below is what says so, since nothing else would").

`rekey_recipes` is a second database with no try/except around it. If it raises, the hub transaction is already committed, the route answers 500, and the recipes are left on a key no variant carries — the exact state `rekey_in_session`'s docstring is written against — with nothing in the log naming the topology or the `moved` map that would let anyone repair it.

A `try/except` that logs `moved` at error level before it re-raises (or that logs and still returns 200, since the hub write did land) would make the promise true. It is three lines and it is the only record that would exist.

pixlstash/hub/workflow_card_writes.py
```diff
@@ -0,0 +1,588 @@
...
+    workflow_key,
+)
+
+logger = get_logger(__name__)
+
+# Every table the owner's decisions about a card live in, all keyed on
+# ``workflow_key``. Ordered children-last is not needed - none of them
+# references another - but the tuple IS the carry-over's definition of "the
+# user attributes", so it is the one place a new card table has to be listed.
+_KEYED_TABLES = (
```
lindkvis, line 49: > A table added to the card's vocabulary belongs in that tuple, or a mark flip silently drops it - which is the failure this module is shaped to make impossible to forget.

The shape makes it easy to *find*, not impossible to forget — nothing fails if the next card table is not added here. Your own mutation table has "drop `workflow_cover` + `workflow_unstacked` from `_KEYED_TABLES`" as a row, which is the same observation: it took remembering.

It is derivable, so it can be enforced. The tuple is today exactly the hub tables carrying a `workflow_key` column, minus the two `_rekey_variants` handles by `UPDATE`:

```
tables with a workflow_key column: workflow_attr, workflow_cover,
    workflow_default_override, workflow_file, workflow_key_picture_input,
    workflow_key_pins, workflow_stack_member, workflow_unstacked,
    workflow_variant
_KEYED_TABLES:                    the same, minus workflow_file and
                                  workflow_variant
```

A guardrail that scans `hub/schema.py` for `workflow_key` columns and asserts the set equals `_KEYED_TABLES | {"workflow_file", "workflow_variant"}` is about eight lines, and it turns "somebody will notice" into "the build says so". The two exclusions want naming in it, since they are the interesting half.

pixlstash/routes/workflows.py
```diff
@@ -675,4 +909,350 @@ def list_card_pictures(
...
+                detail=f"This workflow has no LoRA slot called {unknown[0]!r}.",
+            )
+        if getattr(server.vault, "library_uuid", None) is None:
+            # The flip decides its merge winner on the vault's picture counts
+            # and has to move the vault's saved recipes afterwards. With no
+            # library open it could do neither, and a re-key that skipped both
+            # would pick an arbitrary winner and orphan the recipes.
+            raise HTTPException(
+                status_code=503,
+                detail="No library is open, so a workflow cannot be re-keyed.",
```
lindkvis, line 1130: Both 503 branches — this one and `/inputs` at :1225 — are unasserted. `/inputs` even declares `503` in its `responses=`, so it is on the published contract with nothing holding it.

Minor next to the rest of this suite, and the branch is cheap to reach (a `workflow_env` whose vault reports no `library_uuid`). Worth one test each, mostly so the ordering here stays put: the 422 on an unknown slot label has to come before the 503, or a caller with no library open gets told about the wrong problem.

pixlstash/services/saved_recipe_service.py
lindkvis: Fixed in 839f0ba2, and the fix is in `moved` rather than only here.

`_rekey_variants` now lists a surviving old key among its own successors, so the mapping stops claiming a card went away when a variant is still sitting on it. This condition is then just `old_key in successors`, which is the question that was being asked all along.

One thing that fell out of it: the ordering. `pictures_per_new` was summed over `new_keys`, which has no entry for a key only an un-re-keyable variant is on — so the surviving half ranked at zero pictures and `SlotMarkResult.key` would have sent the caller to the smaller half of its own split. It is now summed over where every variant sits *after* the flip, which is the same number in every case where everything moved.

`test_a_surviving_card_keeps_its_saved_recipes` is the repro from the thread, Unstack and all, and it asserts the response too (`merged in moved[merged]`). Shown failing under both mutations, restored after each:

| Mutation | Went red |
|---|---|
| this line back to `successors[0] == old_key` | `..._keeps_its_saved_recipes` |
| `moved` back to `news` alone | `..._keeps_its_saved_recipes` |

Contract updated in integration §2.3 (a key in `moved` may list itself, and what a client must check before deciding a card is gone) and in `backend_architecture.md`.

pixlstash/routes/workflows.py
lindkvis: Fixed in 839f0ba2. `try/except` that logs the whole `moved` map at error level and re-raises.

Re-raise rather than answer 200, because a saved recipe is authored and quietly stranding one is the thing `rekey_in_session` exists to stop — a 500 is the only thing that tells the owner at all. The comment says why the map and not the count: a second `PUT` cannot repair it, since the marks asked for are by then the marks in force, so the flip re-keys nothing and the log is the only record of which key each recipe set belongs on.

`test_a_re_keying_that_cannot_move_the_recipes_says_which_keys_to_repair` drives it with a raising `rekey_recipes`, asserts the key is in the log, and asserts the hub half landed — which is what makes the record necessary. Deleting the log line turns it red.

pixlstash/hub/workflow_card_writes.py
lindkvis: Fixed in 839f0ba2 — `test_a_mark_flip_carries_every_table_a_card_key_appears_in` in `test_architecture_guardrails.py`, derived exactly as in the thread.

Two things I added on top of what you suggested. The `derived` set is asserted *disjoint* from `_KEYED_TABLES` as well, because the failure in that direction is worse and silent: a derived table copied by the carry-over would duplicate a variant onto both halves of a split. And there is a teeth test, since a regex that silently matched nothing would pass for any tuple — the shape this kind of guardrail fails in.

Dropping `workflow_cover` from the tuple turns it red. The docstring claim is now backed rather than restated; `backend_architecture.md` records the derivation and names the two exclusions.

pixlstash/routes/workflows.py
lindkvis: Fixed in 839f0ba2 — one test each.

`test_a_re_key_needs_an_open_library_and_says_so` asserts the ordering you flagged as the reason for the test: a bad slot label is still a 422 with no library open, so a caller who typed one is not sent looking at the wrong problem.

Worth noting for anyone writing the next one: `Vault.library_uuid` is a read-only property, so the tests patch `_library_uuid`, its backing attribute. Patching the property name with `raising=False` succeeds and proves nothing but that a fake can be attached.

Removing the `/inputs` 503 branch turns that test red.

PR #1433, opened 2026-09-19
MCP: make it reachable, findable and worth using

lindkvis:
Follow-up to #1370, from five problems hit in use.

**1. Registered for one directory only.** `claude mcp add` defaults to `--scope local`, so the one-liner the Connect AI agent dialog handed out registered the server only for the directory it was run in. Everywhere else the agent had no PixlStash tools at all and answered questions about the library by reading `vault.db` with sqlite3. The command now passes `-s user`.

**2. The address was the wrong one.** The desktop shell serves its window from an ephemeral loopback port picked per launch, and the dialog baked that into the client's config, so it died on the next restart. `pixlstash-mcp` now defaults `--url` from the `port` in `server-config.json`; the dialog emits no address at all. `PIXLSTASH_URL` then explicit `--url` still win.

**3. Not on PATH without installing PixlStash into your own Python.** An MCP client spawns `pixlstash-mcp` by name with no shell and no login profile. The app already solves exactly this for `pixlstash-cli` with a shim in `~/.local/bin`; this adds the second one, on the same **Shell command** setting and in the same directory, forwarding to the interpreter the build already ships. Windows goes straight to that interpreter because its launcher owns no console — which matters more here than for the CLI, since stdout is the JSON-RPC channel.

**4. A closed port said only "connection refused".** The configured port is optional on the desktop — bound only when remote access is on. That now has its own message naming the cause, printed at start-up rather than waiting for the first tool call.

**5. It did not cover the library, or introduce itself.** Asked how many picture sets exist, an agent went to the shell: there was no tool for it, and `initialize` returned no `instructions`. Added `list_sets`, `list_characters`, `list_projects` and `set_id`/`character_id`/`project_id` filters, and instructions that name the sqlite shortcut and why a raw row can be wrong (a picture's stored `project_id` is re-derived from membership before the API returns it).

No new routes and no registry entries: the three list routes are already `_LIST_AWARE`. A picture-scoped token gets narrowed set and character lists but is refused `/projects` by resource type, with a test pinning that difference.

Backend 26 tests (was 14), electron 264, frontend 4467, ruff and tsc clean.

It is five fixes in one PR because they are one story — the server was unreachable, unfindable and unhelpful in turn. Happy to split if that reviews better.


pixlstash/mcp_server.py
```diff
@@ -180,8 +317,33 @@ def fetch(path: str, params: dict) -> tuple[int, str, bytes]:
...
+            logger.warning("[mcp] %s closed the connection on %s", base, path)
             raise ToolError(
-                f"Could not reach PixlStash at {base}: {exc.reason}"
+                f"{base} accepted the connection and closed it without "
+                "answering. That is what a TLS listener does when it is sent "
+                "plain HTTP, so PixlStash is probably serving https on this "
+                "port. Check `require_ssl` in its server-config.json, and "
+                "point this server at that file with --server-config."
+            ) from exc
+        except ssl.SSLError as exc:
```
lindkvis: This branch is unreachable for the case it documents, so the message it wants to give is never the one the owner sees.

`urllib.request.AbstractHTTPHandler.do_open` wraps everything `h.request()` raises -- the TLS handshake included -- in `URLError`, and `except urllib.error.URLError` sits above this clause. Reproduced against a local TLS listener with an untrusted self-signed certificate: what comes out is `Could not reach PixlStash at https://...: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate`, via `unreachable_message` -- never the `--server-config` hint.

`ssl.SSLError` here only catches a failure during `response.read()`, which is not what the docstring is about. The fix is a line in `unreachable_message`, beside the `ConnectionRefusedError` case: `isinstance(reason, ssl.SSLError)`. Worth noting none of the new tests exercise this branch, which is why it is green.

pixlstash/mcp_server.py
```diff
@@ -376,9 +637,32 @@ def main(argv: list[str] | None = None) -> int:
...
+    # simply start later, and exiting would have the client give up for good.
+    warn_if_unreachable(fetch, url)
+    serve(fetch)
     return 0
 
 
+def warn_if_unreachable(fetch: Fetch, url: str) -> bool:
+    """Probe the server once; return whether it answered."""
+    try:
+        fetch("/tags", {"limit": 1})
```
lindkvis: The probe cannot see a bad token. `fetch` *returns* `(401, ...)` for an HTTP error rather than raising (the `HTTPError` handler above), so a wrong or expired `PIXLSTASH_TOKEN` -- the most common misconfiguration after the port -- makes this return `True` and print nothing, and the owner still first hears about it at the first tool call. `_get(fetch, "/tags", {"limit": 1})` raises `ToolError("PixlStash answered 401: ...")` and the existing `except ToolError` already prints it.

Separately, `limit` is inert: `GET /tags` takes no paging params and does a `GROUP BY` over the whole tag table, first materialising every scope-allowed picture id when the token is scoped. That is a fairly heavy query to run once per MCP client process at start-up; `/pictures/count` would be a cheaper liveness check.

frontend/src/components/settings/ConnectAgentDialog.vue
```diff
@@ -22,20 +21,24 @@ const props = defineProps({
 
 const emit = defineEmits(["close", "created"]);
 
-// `--url` is the server root: pixlstash-mcp appends `/api/v1` itself, so
-// passing API_BASE_URL unchanged would send it to `/api/v1/api/v1/...`.
-// Computed, not read at setup: AccountSection mounts this dialog permanently
-// and it spends nearly all of its life closed.
-const serverUrl = computed(() => API_BASE_URL.replace(/\/api\/v1\/?$/, ""));
-
+// No `--url`, deliberately. The desktop shell serves this window from an
```
lindkvis: Right for the desktop case, but it also removes the only correct answer for the remote one. `API_BASE_URL` comes from `deriveBackendUrl()`, so for an owner who reached this dialog over the LAN it named the real host; both blocks now emit a bare `pixlstash-mcp`, which resolves against the *client* machine's `server-config.json` and defaults to `127.0.0.1`.

The README covers it ("Pass `--url` only to point it at a different machine"), but the dialog is what gets copy-pasted. A line next to the blocks when `window.location.hostname` is not loopback would close the gap.

electron/src/main.ts
```diff
@@ -579,6 +595,20 @@ function applyShellCommand(): void {
...
     const { launcher, windowsHub } = shimForwardsTo();
     const installed = syncShim(shellCommand, launcher, shimPath(), windowsHub);
+    // The MCP shim rides the same preference and the same directory. Its own
+    // success is deliberately not folded into `shimReachable`, which is about
+    // the CLI hint shown in Settings.
+    const mcp = mcpShimForwardsTo();
+    // Its own config path, not the platform default: the desktop keeps a
+    // separate server-config.json, and reading the wrong one gets the wrong
+    // port and the wrong scheme.
+    syncMcpShim(
```
lindkvis: The result is dropped while `syncUserPath` below follows only the CLI shim, and the two can disagree: `shimBlocked` is per file, so if the user already has their own `pixlstash` in that directory the CLI shim refuses (`installed === false`) while `pixlstash-mcp.cmd` is written anyway. On Windows that removes the PATH entry for the directory the MCP shim just landed in -- written and unreachable.

Either OR the two into `installed`, or warn when they diverge.

tests/test_mcp_server.py
```diff
@@ -175,6 +195,209 @@ def test_an_unknown_protocol_version_gets_ours():
...
+        thread.start()
+        fetch = mcp_server.http_fetch(f"http://127.0.0.1:{port}", "example-token")
+        result = _call(fetch, "list_tags")
+        thread.join(timeout=5)
+
+    assert result["isError"] is True
+    text = result["content"][0]["text"]
+    assert "closed it without answering" in text
+    assert "https" in text and "--server-config" in text
+    assert context.verify_mode is not None  # keeps the import honest
```
lindkvis: This line and the `context = ssl.SSLContext(...)` it refers to are dead: the fake listener never wraps the socket, so the server context is unused, and the import is already honest -- `ssl` is used by `test_the_servers_own_certificate_is_trusted_and_still_verified` just above. Dropping both makes the test say only what it tests.

The test itself is good: `RemoteDisconnected` really is reachable (`h.getresponse()` is outside urllib's inner `try`, and `ConnectionResetError` is not a `URLError`), and driving it through a real socket proves it.

lindkvis (commented):
Reviewed `origin/develop` dafb6c5f against the PR head 04dd4312 (`git diff origin/develop...refs/remotes/pr/1433`).

The core move is right: one `server-config.json` deciding port, scheme and certificate, passed explicitly by the desktop shim, and `ssl_context_for` pinning the self-signed certificate rather than turning verification off. No new routes, no migrations; the four routes the new tools reach already carry `_LIST_AWARE` entries in `authz/registry.py`, so nothing is owed there. Default stays on loopback, so the bearer token has no default path off the box.

Five findings below. The first two are the ones I would fix before merge; the others are small.

pixlstash/mcp_server.py
lindkvis: Confirmed and fixed in 055d9ee0. I reproduced it against a local TLS listener with an untrusted self-signed certificate before changing anything, and got exactly what you describe: `Could not reach PixlStash at https://...: [SSL: CERTIFICATE_VERIFY_FAILED]`, via `unreachable_message`. The `except ssl.SSLError` clause never ran.

`unreachable_message` now has an `isinstance(reason, ssl.SSLError)` branch beside the `ConnectionRefusedError` one, and the dead clause is gone. `test_an_untrusted_certificate_says_so_and_names_the_config` covers it; removing the branch turns it red.

pixlstash/mcp_server.py
lindkvis: Right on both counts, fixed in 055d9ee0.

The probe now goes through `_get`, so a non-200 raises `ToolError` and the existing `except ToolError` prints it. Verified live against the desktop backend: a bad token now prints `pixlstash-mcp: PixlStash answered 401: Not authenticated` at start-up, where before it printed nothing.

Also switched to `/pictures/count` and dropped the inert `limit`. `test_the_start_up_probe_sees_a_rejected_token` asserts both the False return and that the path asked for is `/pictures/count`, so probing through `fetch` or going back to `/tags` each turn it red.

frontend/src/components/settings/ConnectAgentDialog.vue
lindkvis: Agreed, and it was a real hole: I removed the address for the desktop case and took the remote one with it.

055d9ee0 splits them. The dialog computes `remoteUrl` from `window.location`; on loopback it stays empty and nothing is baked in, and off loopback both blocks carry `--url <origin>` again, plus a `.cad-warn` line saying the agent must be able to reach that address and trust the certificate if https is on. `names the address when PixlStash is being viewed over the network` stubs `location` to a LAN host and asserts the flag, the `args` array and the warning; jsdom runs on localhost, so without the stub that branch had no coverage at all.

electron/src/main.ts
lindkvis: Fixed in 055d9ee0: `syncUserPath(installed || mcpInstalled, ...)`, with the divergence you describe written down beside it.

I went with OR rather than a warning because the failure is silent and Windows-only, and the directory is worth keeping for either file. `shimBlocked` being per file is exactly the case: a user with their own `pixlstash` gets the CLI shim refused while `pixlstash-mcp.cmd` is written, and taking only the CLI answer stripped the PATH entry for a file that had just landed in it.

tests/test_mcp_server.py
lindkvis: Both gone in 055d9ee0. The `ssl.SSLContext` and the `assert context.verify_mode is not None  # keeps the import honest` were cargo-culted; `ssl` was already used by the certificate test above, so the comment was not even true. The test now sets up only the plain socket it actually drives.

PR #1434, opened 2026-09-19
Workflows & Recipes F1a: WorkflowsView on /workflows-next

### LINKED ISSUE #1402: Workflows & Recipes F1a: WorkflowsView on a temporary route: grid, sort, empty state, stack panel (Grid)

## Depends on

|---|---|---|
| #1401 (F0) | yes | `git cat-file -e origin/develop:frontend/src/components/widgets/WorkflowCard.vue` |
| #1395 (B3) | yes | `git grep -q '/workflows/cards' origin/develop -- pixlstash/authz/registry.py` |

## What

- `views/WorkflowsView.vue` + `stores/useWorkflowsStore.js` on a **temporary route with no sidebar entry** (e.g. `/workflows-next`, removed in F1b):
  - Toolbar: Sort with Your ratings, Recently used, Picture count. Add maps to today's import.
  - Card grid, with the column count from the real container width.
  - Empty state "Nothing found yet": *Drop a workflow file*, *Open watched folder*, *Connect ComfyUI* (gone once connected), and the watched folder path as a line. Built on the `visual-language.md` §9 empty-state rule.
- `panels/StackPanel.vue`, **Grid only**:
  - ▸ opens one DOM element after the last card of the stack's row (`grid-column: 1 / -1`), pushing later rows down.
  - The notch is `.tbm-caret` positioned by `--notch` from the column index (centre of the column; left with one column), never from a stored pixel offset.
  - A `ResizeObserver` gives the column count, and the panel re-anchors on resize.
  - Header (`--bar-height`): name, "N workflows", "N hidden · Show", Unstack all, close. The Grid|List switch arrives in F2.
  - Member cards at full size. The cover member wears the shipped `.stack-cover-flag`. Each member's special row says how it differs from the cover.
  - **Neutral panel**: `--panel` ground, 1px border, `--radius-lg`, bottom margin `--space-4`. **No olive wash**: the open card is marked only by the rotated ▸.
- **One stack open at a time.** Opening a second closes the first and keeps the new card on screen.
- No virtualisation.
- **Keyboard:**
  - One tab stop with a roving cursor, the shape of `useGridKeyboardNav.js`. Members are indices in the same flat list, so Down from the stack's row enters the panel at the nearest column.
  - Space selects. Enter toggles the panel on a stack card and opens ⓘ on a workflow card. Shift+F10 / Menu key opens the card menu.
  - Esc closes the innermost first (popover → menu → panel → selection). Closing the panel returns focus to the stack card.
- **ARIA:**
  - One `treegrid`. Cards are rows with `aria-selected`, and stack rows add `aria-expanded` / `aria-controls`. Members are `aria-level=2` rows with `aria-posinset`/`aria-setsize`.
  - The panel header is a row with one gridcell holding a `role="toolbar"` named after the stack. The notch is `aria-hidden`.
  - A live region announces "X opened, N workflows, M hidden".
- Selection: a click selects and feeds the inspector. ▸, the layered count and double-click toggle the stack. Ctrl/Shift multi-select can mix top-level cards and members.

## Acceptance

- Tests: one stack open, the panel re-anchors on resize, keyboard across the panel boundary, Esc order. Frontend architecture doc §4/§5 updated (say that one-open-at-a-time deliberately differs from the picture grid).

lindkvis:
Closes #1402. v1.12 Workflows & Recipes, step F1a: the Workflows screen that
replaces the shelf in F1b, on a temporary route with no sidebar entry so the
shipped `/workflows` keeps working beside it.

## What

- `views/WorkflowsView.vue` + `stores/useWorkflowsStore.js` on `/workflows-next`.
  Toolbar: Sort (Your ratings / Recently used / Picture count) and Add, which is
  today's workflow import unchanged. Card grid, `--wf-columns` computed from the
  real container width rather than `auto-fill`, so the painted track and the
  cursor's arithmetic cannot disagree. "Nothing found yet" empty state.
- `panels/StackPanel.vue`, Grid only. One element at `grid-column: 1 / -1` after
  the last card of the stack's row, so opening pushes later rows down. The notch
  is `.tbm-caret` aimed by `--notch` from the card's column index, recomputed on
  every resize. Neutral `--panel` ground, no olive wash — the open card is marked
  by its rotated ▸ alone. One stack open at a time.
- `GET /workflows/cards` gains `last_used`. It is the only field *Recently used*
  can read: `rank` is a rating and `covers` is an order, not a date.

## Notes for the reviewer

**`.stack-cover-flag` moved** from `ImageGrid.css` (scoped) to `App.css`
(global), because the picture grid and this panel now wear the same chip and a
scoped copy in two sheets is how the two drift. Nothing else declares it.

**The diff is ~2400 lines against the step's ~600 guideline.** It is one
shippable unit — a view, its store, its panel and their tests — and I did not
find a split where the halves merge independently. The largest single reduction
available is sharing the empty-state and toolbar CSS with `LibraryEmptyState` /
`WorkflowShelf` (~115 re-typed lines), which I left alone: it means refactoring
a screenshot-tested first-run screen, which belongs in its own change.

**Two controls the design lists are not here**, and the reason changed while
this was in review. B4 (#1396) merged and shipped the endpoints; the blocker is
now the read side. `GET /workflows/cards` serves no `stack_id` for
`POST /workflows/stacks/{id}/unstack` to be addressed by, and it drops hidden
cards *before* grouping, so a stack's hidden-member count is not in the payload
at any level. Each needs one more field on the card. The grid-level `hidden` /
`one_offs` counts are a different number and are shown. The Grid|List switch is
F2's, and Shift+F10 opens ⓘ because F1a has no card menu to open.

## From the pre-push review

An adversarial pass over the diff found no secrets and nothing private. Of its
other findings I fixed the substantive ones, each now with a test that fails
when the fix is reverted:

- Closing a stack from the card's own ▸ could leave the grid with **zero tab
  stops**, stranding a keyboard user. The cursor is now an entry id, not an
  index — which also fixes it silently moving to a different card on a resort.
- ArrowDown out of a panel landed in the wrong column whenever the last member
  row was ragged (five members across four columns). Vertical movement now steps
  by whole rows over the padding holes.
- The panel rendered nothing until every member request resolved, and its header
  counted arrivals — "1 workflow" over a stack of six, permanently if a request
  failed. It now draws the cover immediately and counts `stack_size`.
- `aria-owns` on the open stack row. The panel sits after the *last card of the
  row*, so in DOM order its rows follow that card, not the expanded one, and a
  screen-reader user pressing Down got the right-hand neighbour instead of the
  children. Also added `aria-multiselectable`.
- The live region announced twice per open, the first time as "0 workflows", and
  never announced the close.
- `openStack` re-issued every member request on a double-click; non-stack cards
  could be opened into a panel; `measure()` read `clientWidth` including the
  grid's own padding, over-counting columns near a boundary.
- Docs: §5 claimed Shift+F10 opened a card menu that does not exist, and that
  ▸ returned the cursor to the stack card. `last_used` and `rank` are now in the
  canonical card shape in `utils/workflowCard.js` and in
  `integration_architecture.md`.

**One fix is outside this feature.** `SearchResultBar.test.js` mounted 33
components and unmounted none, so `Tooltip.vue`'s `onUnmounted` ran after jsdom
teardown and threw. Adding two test files shifted the scheduling enough to
surface it about one run in three. I did not establish that it never happened
before this branch, only that it does happen with it and does not after the fix
— four consecutive clean full runs.

I disagree with one finding: it called the `aria-hidden`, `tabindex="-1"`
layered-count button an unreachable control. It is decoration that happens to be
clickable, ▸ is the announced control, and the two now do exactly the same thing
(the `.stop` that made them differ is gone), which is what makes hiding the
second one honest.

## Tests

4494 frontend tests and `tests/test_workflows_api.py` green; ruff and eslint
clean; `render_backend_architecture.py --check` OK. Every new assertion was
shown failing against the code it guards and green again after restoring.


frontend/src/components/views/WorkflowsView.vue
```diff
@@ -0,0 +1,844 @@
...
+    key: card.key,
+    card,
+    cardIndex,
+  }));
+  const openKey = store.openStackKey;
+  if (!openKey) return cards;
+  const at = cards.findIndex((entry) => entry.key === openKey);
+  if (at < 0) return cards;
+  const cols = Math.max(1, columns.value);
+  const rowEnd = Math.min(cards.length, (Math.floor(at / cols) + 1) * cols);
```
lindkvis: `Math.min(cards.length, …)` breaks the invariant the rest of this file is built on. When the open stack sits in the last, **incomplete** card row, `rowEnd` becomes `cards.length`, which is not a multiple of `cols` — so from the block onwards `index % columns` no longer names the column a row is drawn in.

Worked through: 6 cards, 4 columns, stack at index 5 (column 1), 2 members gives `[c0…c5, m0, m1, hole, hole]`. ArrowDown from the stack card is `verticalStop(9)`; index 9 is a hole, 13 is out of range, and the fallback `firstStop(9, +1)` only scans *forward*, so it returns `null` — the cursor does not move at all and the panel is unreachable by ArrowDown. ArrowUp from `m1` (panel column 1) lands on `c3`, drawn in column 3.

Every fixture in `WorkflowsView.test.js` opens the stack at index 1 of a full first row, so the suite cannot see this. Padding the card list out to a whole row before splicing would restore the invariant, and the `verticalStop` fallback wants to scan back toward the list rather than further out of it.

frontend/src/stores/useWorkflowsStore.js
```diff
@@ -0,0 +1,237 @@
...
+    hidden.value = 0;
+    loaded.value = false;
+    error.value = "";
+    members.value = {};
+    inflight.clear();
+    openStackKey.value = null;
+    selectedKeys.value = [];
+  }
+
+  onScopeDispose(onSessionReset(reset));
```
lindkvis, line 252: No epoch stamp, so a session reset does not stop the previous session's responses from landing. `fetchCards` and `openStack` both `await` and then write unconditionally — a `listWorkflowCards()` still in flight when the credential changes resolves *after* `reset()` and puts the old session's cards, and their cover thumbnail URLs, into the new one.

`useWorkflowShelfStore`, the store this one's docstring contrasts itself with, carries `let epoch = 0`, bumps it in `resetForSession()` and guards every write with `if (mine !== epoch) return`, under the comment "an in-flight fetch that resolves after a session reset must not write its rows into the new session's store". The same hazard applies here. `sessionReset.test.js` seeds synchronously, so it would stay green either way.

frontend/src/stores/useWorkflowsStore.js
```diff
@@ -0,0 +1,237 @@
...
+      };
+    } catch (err) {
+      error.value = errorMessage(err, "Could not read this stack.");
+      console.warn(
+        `[workflows] could not read the members of ${coverKey}`,
+        err,
+      );
+      // The cover alone rather than nothing: the panel then says what it can
+      // instead of hanging on a spinner with an error nobody reads.
+      members.value = { ...members.value, [coverKey]: [card] };
```
lindkvis: The failure is cached as an answer. This write plus the `if (members.value[coverKey] || …) return` guard on L136 makes one dropped request permanent: reopening the stack never retries, so the panel reads "N could not be read" for the rest of the session even once the network recovers.

`useWorkflowShelfStore.loadSamples` documents exactly this mistake ("Writing an empty array on the error path made the 'already have them' guard permanent") and settles it with three states — in flight, answered, failed — where a retry clears the failed set and asks again.

frontend/src/stores/useWorkflowsStore.js
```diff
@@ -0,0 +1,237 @@
...
+    } finally {
+      inflight.delete(coverKey);
+      membersLoading.value = false;
```
lindkvis: `membersLoading` is one boolean shared by every stack, so it clears on whichever request finishes first rather than on the open one. Open `b`, then click `e` before `b`'s members land: `b`'s `finally` clears the flag while `e` is still fetching, and `StackPanel`'s `pending` flips from "reading the rest…" to "N could not be read" — the panel reporting failure for requests that are still running, which is the specific lie the `countLabel` comment above it sets out to avoid.

Clearing only when `coverKey === openStackKey.value` would be enough.

frontend/src/components/panels/StackPanel.vue
```diff
@@ -0,0 +1,230 @@
...
+        aria-level="2"
+        :aria-posinset="index + 1"
+        :aria-setsize="size || members.length"
+        :aria-selected="selectedKeys.includes(member.key)"
+        :tabindex="cursorKey === member.key ? 0 : -1"
+        :data-key="member.key"
+        @click="emit('select', member.key, $event)"
+      >
+        <div class="stack-panel__cell" role="gridcell">
+          <WorkflowCard :card="member" />
```
lindkvis: Members are served with the **whole stack's** `stack_size`: `workflow_card_service.py` sets `member.stack_size = len(stack.member_keys)` for every member, not just the cover (deliberately, so a member opened alone carries its difference chips). So `isStack()` is true for each row here, and `WorkflowCard` renders both the layered-count button and the ▸ toggle on every member in the panel. Neither does anything — no `@toggle` is bound, and `store.openStack(memberKey)` would not find the key in `cards` anyway.

Two visible consequences: the ▸ carries `aria-expanded="false"` with no `aria-controls`, so a reader is offered N collapsed expandables that never expand, and `cardAccessibleName` announces every member as "stack of N workflows". The cover is also on screen twice with ▸ open in the grid and ▸ closed here.

Worth a prop on `WorkflowCard` that suppresses the stack affordances — note `factChips` branches on `isStack` too, so simply rewriting `stack_size` to 1 for these rows would change which chips they show.

lindkvis (commented):
Reviewed `dafb6c5f` (origin/develop) against `088442bb` (PR head), `git diff origin/develop...`.

The shape is right: the flat index space, the id-based cursor, the computed column count instead of `auto-fill`, and `aria-owns` re-parenting the member rows are all the correct mechanisms, and the CSS is token-clean throughout (no hex, no raw shadow, no off-ramp font size). The tests assert real behaviour rather than presence. Backend side is fine — `GET /api/v1/workflows/cards` is already declared `_OWNER` in `ROUTE_POLICIES`, and this PR only adds a field to an existing response model, so no registry work is owed.

Five findings inline: one cursor bug the fixtures cannot reach, three store-lifecycle ones, and a dead control on every member row.

frontend/src/components/views/WorkflowsView.vue
lindkvis: Fixed in c0fa26bd. The stack's **own row** is now padded out to whole before the block is spliced in, so `rowEnd` is always a multiple of `cols` and `index % columns` keeps naming the column a row is drawn in:

```
a b c d | e f · · | e e1 e2 ·
```

`WorkflowsView.test.js` gains "a stack in the last, incomplete row keeps the panel reachable" — six cards over four columns with the stack at index 4, which is your case. Down from the stack card reaches the panel and Up from `e1` lands on `f` (column 1), not `b`. Reverting to `Math.min(cards.length, …)` fails it (`expected 'e2' to be 'e'`).

On the `verticalStop` fallback: I tried flipping it to scan back toward the list and could not get it to change any reachable behaviour once the head padding was in. The only trailing holes left are at the very end of the list (a stack opened in the grid's last row), where Down should do nothing anyway; a hole still *inside* the list wants the old forward scan, and flipping it broke the existing "Down from `d` skips the hole" assertion by landing on `b2` instead of `e`. So I left it alone rather than keep a change no test can see. Say the word if you want it flipped regardless.

frontend/src/stores/useWorkflowsStore.js
lindkvis: Fixed in c0fa26bd. `let epoch = 0`, bumped in `reset()`, and every write in `fetchCards` and `openStack` guarded with `if (mine !== epoch) return` — same shape as `useWorkflowShelfStore`.

`reset()` also clears `loading` now, which the epoch made necessary: the old fetch's `finally` belongs to the previous epoch and no longer clears it, and `fetchCards` returns early while it is set, so the new session would have sat on an empty grid.

New test in `useWorkflowsStore.test.js`, "drops a fetch and a member request that resolve after the reset", holding both the list and a member request open across `reset()`. Removing the `epoch += 1` fails it; so does dropping the `loading.value = false`.

frontend/src/stores/useWorkflowsStore.js
lindkvis: Fixed in c0fa26bd. Three states, as `loadSamples` has them: in flight in `inflight`, answered in `members`, failed in `membersFailed`. The error path no longer writes into `members`, so the "already have them" guard does not see it and a reopen asks again; `membersFailed` is cleared as the retry starts, not as it ends. `openMembers` already falls back to the cover, so the panel draws the same thing — it just stops being permanent.

New test "asks again after a failure instead of caching it as the answer": two requests fail, the stack is closed and reopened, two more go out and the members land. Putting `members.value = {…[coverKey]: [card]}` back in the catch fails it.

frontend/src/stores/useWorkflowsStore.js
lindkvis: Fixed in c0fa26bd, slightly differently from your suggestion: `inflight` was already tracking exactly the right thing, so it became a `ref(new Set())` and `membersLoading` is now `computed(() => inflight.value.has(openStackKey.value))` — one less piece of state rather than a condition on clearing it.

New test "reports the OPEN stack's loading, not whichever request finished first": open `the-stack`, then `few-but-loved`, release the first one's members only, assert `membersLoading` is still true. Restoring the shared boolean fails it (`expected false to be true`).

frontend/src/components/panels/StackPanel.vue
lindkvis: Fixed in c0fa26bd. `WorkflowCard` takes a `member` prop; `StackPanel` sets it on every row. It gates a new `stackMark` computed — the layered count, ▸ and the `wf-card--stack` edges — while `stack` itself is untouched, so `factChips` and the "differs by" label behave exactly as before, which is your point about rewriting `stack_size` to 1. `cardAccessibleName(card, { member })` drops the "stack of N workflows" clause and nothing else.

New test "has none of them on a member row either, but keeps its differences" asserts both halves: no ▸, no layered count, no `wf-card--stack`, no "stack of 6 workflows" in the name — and that "differs by: + face detailer" and the chips are still there. Reverting either half (the `stackMark` gate, or the name's `&& !member`) fails it.

PR #1435, opened 2026-09-19
Recipe tab: say why a run is refused, and use a picture as a workflow's input

lindkvis:
Finishes #1406.

Most of that issue — the Info | Recipe tabs and `OverlayRecipePanel.vue` —
already shipped on `develop` under its sibling #1313 (PR #1426). I did not
notice until the closing rebase and rebuilt this branch down to what #1406 still
wanted. That is why the diff is small and mostly deletion.

## What this adds

**The Recipe tab says why a run is refused.** `canGenerateVariants` was one
boolean folding together three unrelated facts, so an A1111 picture — whose
recipe the tab renders in full — was offered an enabled *Generate variants…*
that no ComfyUI could serve, and a share-link reader was told ComfyUI was not
connected on a machine where it was. Four reasons now, ranked by what the reader
can do about them: the picture (A1111/Forge, or a graph the server called
unrunnable, whose own `reason` supplies the words), then the session, then the
machine.

**`aria-disabled`, not `disabled`.** A natively-disabled button is out of the tab
order, so a keyboard reader could never reach the reason `aria-describedby`
points at. `AppButton` already inked both spellings and its CSS says why; the
panel just was not using it. The reason is also in prose above the buttons,
because a tooltip needs a hover or a focus.

**"Use as input for…"**, in the Recipe tab and the lightbox's right-click menu.
It closes the lightbox (the run panel it opens is the rail the lightbox covers)
and opens that panel on the picture. Hidden, not disabled, for a read-only
session or an unconfigured machine — the same treatment both context menus give
it.

**The lightbox's I2I menu is gone** (~480 lines). The run panel is a superset:
workflow picker, caption, LoRA swap and slot choice, stack toggle. The one thing
genuinely lost is the per-workflow caption remembered in `sessionStorage`.

## Two things found on the way

**A read-only guard lost its only implementation.** The I2I menu was one of three
`useLoraSwap` callers and the only one guarding the owner-only insertion read on
`isReadOnly`; deleting it would have removed the guard silently, since neither
survivor had copied it. It moved into the composable, where the next caller
cannot forget it — and its sibling `loadAdapters` (`GET /adapters`, also
owner-only) got the same guard, which nothing had.

**The rename reached two surfaces of three.** `SelectionMenu.vue` still said
*Edit with ComfyUI* for the identical `open-comfyui-panel` emit, with a different
icon, in a menu a user reaches seconds later. Both say the same thing now — and
an existing `not.toContain("Edit with ComfyUI")` assertion that the rename made
vacuous now asserts the real entry instead.

It rejected the first version of this branch and was right to: the read-only
reason was in the docstring and the docs but not in the code, and the new action
was gated in both context menus and ungated in the Recipe tab. Both fixed, both
now tested. Also from it: the selection convention below, the dead CSS, the
stale emits line, the `=== false` comment contradicted by the overlay's own
`available: data.available === true` (the overlay now passes the value through),
and the footer not fitting 288px.

Two I did not simply obey:

- **The selection.** It called the unconditional replace a defect. The fix is
  narrower than "do not replace": `useOverlayPictureAsInput` now follows the rule
  `handleImageContextMenu` and `handleFaceBboxContextMenu` already follow —
  narrow to the one picture only when it is not already selected. Acting on one
  picture of a selection built on purpose is not a reason to discard the rest.
- **The banner and Save.** Still absent, and my earlier reason for that was
  wrong: B6 shipped both the match rule (`saved_recipe_service.credit_by_recipe`)
  and `GET`/`POST /recipes`. What is actually missing is a frontend
  `api/recipes.js` and a decision on where the prompt/LoRA normalisation lives —
  re-deriving it in JS makes a second copy of a rule the backend owns, which is
  the hand-copy this repo has been bitten by. That belongs with the step that
  adds the client, not here.

## Checks

259 → 260 frontend suites, 4479 tests, eslint clean, build clean,
`render_backend_architecture.py --check` OK. Every new assertion was
mutation-checked — the guard removed, the reason branch deleted, the gate
dropped, the selection clobbered — red each time, then restored.

---

## Answering the review (d4d2e42d)

All six threads replied to and resolved. Two changed behaviour:

**The selection rule was borrowed from a premise that does not hold.** I copied
`handleImageContextMenu`'s "keep a selection the picture is already in" without
checking why it is right there — the grid selection is *on screen*.
`openOverlay` never touches `selectedImageIds`, so in the lightbox it is
invisible, and both ways into `useOverlayPictureAsInput` name one picture
explicitly. Keeping fifty unseen pictures selected would have handed
`WorkflowRunPanel` all fifty from a control whose tooltip says "this picture".
It now narrows unconditionally, and `ImageGridUseAsInput.test.js` flipped with
it — the case that pinned the old rule now pins the new one and says why.

**The panel now refuses the way the menus refuse.** I had claimed hiding was
"how both context menus treat the same action"; they disable. Rather than fix
the comment I fixed the code: `v-if="comfyuiConfigured"` and `aria-disabled`
for a read-only session, matching both copies. That exposed something the
hiding had concealed — on a read-only A1111 picture the two buttons are refused
for *different* reasons, and both pointed `aria-describedby` at one sentence.
Each now gets its own when they differ and they share one when they coincide.

The rest were comments and docs describing what the code does not do, plus a
6th cell in a 5-column table that rendered nowhere.

## The CI failure (cdbf4722)

`backend-windows shard 2` failed on two `tests/test_restore.py` drain tests.
This diff touches no backend code, but the failure is real and understood, not
a retry-until-green:

Both tests pause a request handler on `release_X.wait(10)` while the main
thread does work the test itself allows up to **20.2 s** for
(`close_started.wait(20)` plus a 0.2 s proof that no swap has happened) before
setting the event. The handler is less patient than the orchestration it waits
on. On Linux the gap is milliseconds; on a loaded Windows runner it is not.

**This exact defect was already fixed once in this file** — 807045a1,
"unstick the restore queue-gap wait", raised the sibling
`release_token_clear.wait(10)` to `120` with a comment noting a Windows runner
takes several times what Linux does. These two are the unfixed siblings, so
they get the same remedy.

Demonstrated rather than assumed: with a 12 s delay inserted before the release
— inside the 20.2 s the test permits — the old timeout fails with the exact CI
error and the new one passes.

**The wider pattern: I checked it, and there isn't one.** An earlier draft of
this description said ~12 other paused-handler waits shared "the same shape"
and wanted a sweep. That was a grep result, not a finding, and the sweep says
otherwise — none of the other 11 sites has this defect. The criterion that
tells them apart:

> It is not "a short release timeout". It is a handler paused on
> `release.wait(P)` while the main thread performs **another bounded wait,
> after the handler is already paused**, of length comparable to or greater
> than `P`, before releasing. The main thread's wait for the handler's own
> *entered* event does not count — it has already returned by the time the
> handler's clock starts mattering.

All 11 others are `after-pause = 0s`; four were read by hand to confirm the
scan. The restore pair were inverted because their orchestration holds a
*second* long wait, `close_started.wait(20)`, after the pause.

**What this fix makes worse, and is now #1441.** Both drain tests set the
release event *outside* any `try`, so an orchestration assertion that fires
leaves the paused thread blocked — holding the scratch `snapshot.sqlite` open,
which 807045a1's comment says breaks the next test's `rmtree` on Windows.
Raising the patience 10 s → 120 s lengthens that stall twelvefold. The trade is
still right — a correct test that stalls on failure beats a wrong test that
fails on success — but it is a real cost of this commit, not a neutral one, and
the `try`/`finally` that removes it is filed as #1441.


lindkvis (commented):
Reviewed `origin/develop` (dafb6c5) ...`pull/1435/head` (69e8b82), three-dot.

The refusal ladder checks out: the four `reason` values it maps are exactly the ones `GET /comfyui/pictures/{id}/recipe` produces, `no_prompt_chunk` is already filtered upstream so the tab is disabled rather than reasoned about, and `runReason` returns null only when `canGenerateVariants` is true — so there is no state left where the button is live and the run would fail. The I2I deletion is clean: every `.overlay-comfy-*` class kept still has a user in the Filters menu or `PluginParametersUI`, every deleted one has none, and the caption-focus Escape branch and `clearComfyuiCloseTimer` went with their state. `aria-disabled` + `aria-describedby` is the right call and `AppButton` really does ink both spellings the same. New CSS is all real tokens.

Six small findings below — one behavioural, the rest are comments and docs that describe something the code does not do.

frontend/src/components/views/ImageGrid.vue
```diff
@@ -7624,6 +7625,31 @@ function handleContextMenuOpenComfyuiPanel() {
...
+function useOverlayPictureAsInput(pictureId) {
+  const id = pictureId ?? overlayImageId.value;
+  if (id == null) return;
+  if (overlayOpen.value) closeOverlay();
+  // Narrowed to this picture only when it is not already in the selection -
+  // the same rule `handleImageContextMenu` and `handleFaceBboxContextMenu`
+  // follow. Acting on a picture the user had deliberately selected alongside
+  // others is not a reason to throw the others away, and there is no undo for
+  // a selection.
+  if (!selectedImageIds.value.includes(id)) {
```
lindkvis: The convention borrowed from `handleImageContextMenu` assumes the selection is **on screen**. In the lightbox it is not: `openOverlay` never touches `selectedImageIds`, so a reader who selected 50 pictures in the grid and then opened one of them gets *Use as input for…* sending all 50 to the run panel — `workflowRunStore.selectionIds` is the live grid selection and `WorkflowRunPanel` reads it at run time — while the entry's own tooltip says "Run a workflow with **this** picture as its input".

Both entrances to this function are per-picture (the recipe panel's button and `delegateWith('use-as-input', contextImage?.id)`), so narrowing unconditionally when the call carries an explicit id would match what the control promises.

frontend/src/components/views/OverlayRecipePanel.vue
```diff
@@ -233,22 +260,103 @@
...
+    return "This recipe cannot be run from here.";
+  }
+  return null;
 });
 
-const emit = defineEmits(["generate-variants"]);
+/**
+ * Whether to offer the picture as a workflow's input at all.
+ *
+ * Hidden rather than disabled, which is how both context menus treat the same
```
lindkvis: "which is how both context menus treat the same action" is not what the menus do — both render the entry and disable it:

- `ImageGridContextMenu.vue:176` (overlay mode): `:disabled="isReadOnly"`
- `ImageGridContextMenu.vue:509` (grid mode): `:disabled="!selectedImageIds.length || isReadOnly"`

So read-only gets a visible-but-dead entry in the menus and no entry at all here. Same claim is repeated in `docs/frontend_architecture.md`. Worth picking one treatment; if hiding is right, the menus should hide too.

frontend/src/components/views/OverlayRecipePanel.vue
```diff
@@ -233,22 +260,103 @@
...
+ * reading them: this machine has no ComfyUI to run anything on, this session
+ * may not run anything, or **this picture is not a thing ComfyUI can run** -
+ * which is the A1111 case the Recipe tab otherwise fills in completely. The
+ * graph reasons come from the recipe read's own `reason` so the words here
+ * cannot drift from the decision the server made.
+ */
+const RUN_REASONS = {
+  a1111:
+    "This picture was made in A1111 or Forge, so there is no ComfyUI graph " +
+    "here to run again. Use it as the input for a workflow instead.",
```
lindkvis: "Use it as the input for a workflow instead" points at a button that `canUseAsInput` hides when ComfyUI is unconfigured or the session is read-only — and the `a1111` branch is deliberately ranked *above* both of those checks, so those are precisely the readers who reach this sentence. On a share link or a machine with no ComfyUI the advice has nothing to act on.

Suggest making the second clause conditional on `canUseAsInput`, or dropping it from the constant and leaving the button to say it.

docs/design/notice-surface.md
```diff
@@ -341,7 +341,7 @@ shipped components had exactly that, and the fix would have broken them:
 | Site | Fill | Before | After the naive fix | Resolution |
 |---|---|---|---|---|
 | `ImageOverlay .overlay-comfy-warning` | `rgba(warning, .2)` on the lightbox | 11.38 / 12.99 | **1.41 / 1.33** | tint → `dark-surface-warning`, text → `on-dark-surface` (9.18 / 10.51) |
-| `ImageOverlay .overlay-comfy-error` | `rgba(error, .2)` on the lightbox | 12.92 / 13.89 | 12.92 / **1.24** | same (9.89 / 11.24) |
+| `ImageOverlay .overlay-comfy-error` | `rgba(error, .2)` on the lightbox | 12.92 / 13.89 | 12.92 / **1.24** | same (9.89 / 11.24) | *(the rule still stands; the class itself went with the lightbox's I2I menu in #1406)*
```
lindkvis: This adds a **6th cell to a 5-column table** (header is `Site | Fill | Before | After the naive fix | Resolution`). GFM truncates each row to the header's column count, so the parenthetical renders nowhere — it is only visible in the source.

Fold it into the Resolution cell, or put it as a sentence under the table.

docs/frontend_architecture.md
```diff
@@ -681,15 +681,21 @@ That makes the gesture deliberately a beat late, which is why the tile carries a
...
+
+  *Use as input for…* is **hidden**, not disabled, for a read-only session or an unconfigured machine — the same treatment both context menus give the same action, because it opens a run panel that could only refuse. The two buttons stack rather than sitting in a row: the pane is `--stats-panel-w` (288px) and these two labels are ~314px together, where the design's own "Run… / Save" fit.
```
lindkvis: "**replaces** the selection with that one picture" is not what it does — it narrows *only when the picture is not already selected*, which is the rule the PR's own test pins (`ImageGridUseAsInput.test.js`, "keeps a selection the picture is already part of"). A reader trusting this paragraph would take the other test for a bug.

changelog.d/1406-overlay-recipe-actions.md
```diff
@@ -0,0 +1,11 @@
+- The Recipe tab now says *why* a picture's recipe cannot be run again instead
+  of offering a button that would fail. **Generate variants…** stays in place,
+  disabled, with the reason beside it — and the reasons are told apart: a
+  picture made in A1111 or Forge has no ComfyUI graph to replay, a graph may
+  have no seed to change, ComfyUI may not be connected, and a shared link is a
+  read-only view. Each says its own thing rather than blaming the connection.
+- The Recipe tab and the right-click menu in the full-screen view gain **Use as
```
lindkvis, line 7: Two things:

- "which selects the picture" carries the same inaccuracy as the architecture doc — the selection is left alone when the picture is already in it.
- The fragment does not mention that the grid context menu and the selection menu entries were renamed from *Edit with ComfyUI* to *Use as input for…*. That is a user-visible label change on a surface people use daily, and the release notes are where it would be read.

frontend/src/components/views/ImageGrid.vue
lindkvis: Fixed in d4d2e42d. You are right, and the premise is the part I got wrong: I borrowed the rule from `handleImageContextMenu` without checking that what makes it right there holds here. It does not — `openOverlay` never touches `selectedImageIds`, so the selection is invisible in the lightbox, and both ways in name one picture explicitly.

`useOverlayPictureAsInput` now narrows unconditionally, and the comment says why the grid's rule does not carry over rather than claiming it does. `ImageGridUseAsInput.test.js` flipped with it: the case that asserted the selection was kept now asserts it is narrowed, and names the reason. I mutation-checked it by putting the guard back — red.

frontend/src/components/views/OverlayRecipePanel.vue
lindkvis: Fixed in d4d2e42d, by changing the code rather than the comment: the panel now does what the menus do. `v-if="comfyuiConfigured"` (no ComfyUI, nothing to be an input *for*), and `aria-disabled` with the reason for a read-only session, present but dead.

That surfaced something the hide had concealed — Run… and *Use as input for…* are refused for different reasons on a read-only A1111 picture, and both pointed `aria-describedby` at one sentence. Each now gets its own when they differ, and they share one when they coincide, so neither button is described by the other's problem. Two tests cover that split.

frontend/src/components/views/OverlayRecipePanel.vue
lindkvis: Fixed in d4d2e42d — dropped the second clause. The constant now stops at "there is no ComfyUI graph here to run again", and the button beside it says what can be done instead, when it can be. It was advice that pointed at a control the reader might not have, which is worse than no advice.

docs/design/notice-surface.md
lindkvis: Fixed in d4d2e42d — folded into the Resolution cell, so the row is five columns again. I checked the rendered count rather than the source this time.

docs/frontend_architecture.md
lindkvis: Fixed in d4d2e42d. The paragraph was describing the behaviour this PR has now changed anyway, so it says "narrows the selection to that one picture" and then states explicitly that it narrows *unconditionally*, which `handleImageContextMenu` deliberately does not — with the reason (the lightbox does not show the selection) so the next reader does not re-derive the wrong rule from the grid's.

changelog.d/1406-overlay-recipe-actions.md
lindkvis: Both fixed in d4d2e42d. "which selects the picture" is now accurate, since the selection is narrowed unconditionally after your first comment. And the rename has its own line:

> *Edit with ComfyUI* is now called **Use as input for…** in the grid's right-click menu and in the selection menu. Same action, clearer name.

The selection menu is in there because answering your other comment found it still said *Edit with ComfyUI* with a different icon for the identical emit.

PR #1437, opened 2026-09-19
Workflows & Recipes B7: one run route with preflight and every reason code

### LINKED ISSUE #1399: Workflows & Recipes B7: one run route with preflight and every reason code

## Depends on

|---|---|---|
| #1396 (B4) | yes | `git grep -q CHANGED_WORKFLOWS origin/develop -- pixlstash` |
| #1397 (B5) | yes | `git grep -q workflow_key origin/develop -- pixlstash/routes/comfyui.py` |
| #1398 (B6) | yes | `git grep -q '"/recipes"' origin/develop -- pixlstash/authz/registry.py` |

## What

- **Runnable-source resolver** for a workflow, in order:
  1. the linked imported file;
  2. the embedded API graph of this library's best kept picture;
  3. this library's best `workflow_recipe_instance.document`, with references resolved through `workflow_recipe_asset` (a forgotten name surfaces as a missing model);
  4. otherwise `no_runnable_source`.
- `POST /workflows/run` takes:
  - source `picture_ids | saved_recipe_id | workflow_key`;
  - target `workflow_key` (another stack member allowed);
  - `prompt`, `negative`, `loras`, `values` by parameter address;
  - `count`, `seed_mode new|keep|fixed`, `inputs`, `destination`.

  With several pictures and no target, the server groups them by each picture's recipe. Keep `MAX_RUNS_PER_REQUEST` and the `allow_unchecked` consent rule on every source. **New runs are not stacked by default.**
- `POST /workflows/run/preflight`: the same body as a dry run.
- Reason codes: `comfyui_not_configured`, `comfyui_unreachable`, `ui_format`, `missing_nodes[]`, `missing_models[{file, folder}]` (blocks the whole batch), `a1111`, `fixed_input_deleted`, `no_lora_loader`, `pixlstash_nodes`, `no_save_node`, `no_runnable_source`.
- LoRA slots go through the #1377 resolver (`detect_lora_targets` / `apply_adapter`, one slot = a node and a field), extended with strengths.
- Edited defaults are overrides applied at run time, never written into a graph.

## Acceptance

- Each reason code asserted for picture, saved-recipe and workflow sources. A missing model blocks a mixed batch. New runs are not stacked. `count` produces N submissions.
- **Policy `OWNER_ONLY`**, which narrows the shipped `PICTURE_SCOPED` run routes. **Chief-security-officer sign-off before merge** (new run surface, policy change).

lindkvis:
Closes #1399. Adds `POST /workflows/run` and `POST /workflows/run/preflight`.

## What

- **A runnable-source resolver** (`services/workflow_run_service.py`). A card is an identity derived from pictures, not a file, so "run this card" has to find something to submit: the linked imported file, else the best kept picture's embedded graph, else the best stored instance document with its asset references resolved, else `no_runnable_source`. A reference the hub can no longer name becomes a model no ComfyUI lists, so a forgotten name surfaces as a missing model.
- **Two routes, one body.** One source of `picture_ids | saved_recipe_id | workflow_key`, optional `target`, plus `prompt`, `negative`, `loras`, `values`, `count`, `seed_mode`, `destination`, `stack`, `allow_unchecked`. Several pictures and no target group by each picture's recipe.
- **Refusals are codes, not sentences** — all eleven from the issue. A batch mixes sources, and a panel grouping "these four are missing the same model" can't do that from prose.
- **New runs are not stacked by default**, unlike the shipped `run_recipe`. That one replays a picture's own graph, so its output is another take of it; this runs a card.
- **`OWNER_ONLY`**, narrowing the `PICTURE_SCOPED` run routes: a card resolves its graph from the whole library whatever the caller is scoped to.

## Not in here

**No `inputs` field.** A card's picture-input setup is read (`fixed_input_deleted`) but nothing fills one yet — that means uploading into ComfyUI's input folder, which is the i2i path `/comfyui/workflows/{name}/run` already owns. I had it as an accepted-and-ignored field and took it out: a caller would send a picture and get a run that never read it. Worth a follow-up.

## Adversarial pass

An independent reviewer was asked to argue for rejection. No secrets, no private information, nothing destructive. It found real defects and they are fixed in the second commit:

- `allow_unchecked` was inert — an unreachable ComfyUI blocked the batch before consent was read. Consent is now part of the blocking rule, and it reaches nothing else: a missing model is a fact that *was* established.
- The pre-flight could 400 where the run returned reasons, and could say `ok: true` for a body the run then rejected. The rule is now explicit and tested: a body that can't be interpreted against the card is a 4xx **on both**; everything else is a reason code.
- `seed_mode: "keep"` on a stored instance submitted seed 0 — `count: 3` would have produced three identical images. Refused with a message instead.
- A failure part way through left prompts queued in ComfyUI with no record returned. It now answers `status: "partial"` with the ids.
- A saved recipe's own LoRAs were dropped, and its `keep_seed` overrode an explicit `seed_mode`. Both fixed.
- Five tests asserted nothing (a 405 behind an `if status == 200`, monkeypatches on attributes that don't exist, Pydantic's validator standing in for the route's cap, all three "sources" resolving to the same tier). Rewritten, plus happy-path coverage for tiers 1 and 2 which had none.

Two findings I'm answering rather than acting on:

- **Nine private helpers imported from `routes/comfyui.py`.** Fair, but there's no cycle (`server.py` already imports from that module) and moving them out of a 3300-line route file is its own change, not this one's.
- **Pre-flight cost.** Real: tier 2 opens files. I cut the resolver depth from 20 to 5 and left a `ponytail:` note. A per-card "this picture carries no graph" cache is the fix if a panel ever calls this per keystroke.

## Size

~2700 lines against the ~600 guideline, and I don't think it splits usefully: roughly 750 production, 1400 tests, 400 docs, and the acceptance criteria ask for every reason code asserted across three source kinds. Say the word if you'd rather have the resolver and the routes as separate PRs.

## Checks

`ruff check` / `ruff format --check` clean. 173 tests in `test_workflows_api.py`, plus `test_architecture_guardrails.py`, `test_ci_shards.py`, `test_saved_recipes.py`, `test_workflow_cards.py`, `test_workflow_library.py` and `test_comfyui_recipe_route.py` — 598 passing locally. Fifteen mutation checks: every new assertion was shown failing with the thing it guards broken, and restored. One of them passed under mutation the first time and the test was rewritten until it didn't.

**Needs chief-security-officer sign-off before merge** (new run surface, policy change), per the issue.


lindkvis:
Reviewed `origin/develop...` the PR head (merge base `dafb6c5f`). One blocker, one should-fix, the rest is polish. The authz work, the reason-code contract and the dry-run/run agreement all hold up.

## Blocker — `read_kept_pixel_shas` raises on every match

`pixlstash/services/workflow_library_service.py`:

```python
return {
    sha
    for (sha,) in session.exec(select(Picture.pixel_sha).where(...)).all()
}
```

`select()` with one column is a `SelectOfScalar`, so `session.exec(...).all()` returns a list of **strings**, not 1-tuples. Unpacking `(sha,) = "aaa…"` raises `ValueError: too many values to unpack (expected 1)`. Reproduced against the real `Picture` model on sqlmodel 0.0.38.

`_fixed_input_reasons` → `read_kept_pixel_shas` → 500 on **both** `POST /workflows/run` and `POST /workflows/run/preflight`, for any card with a `fixed` picture input whose picture is **still in the library**. The feature is inverted: it only works when the picture is gone (empty result, nothing to unpack) — which is the one case the tests cover.

`best_picture_ids`, added in the same diff, gets it right, and so does every other single-column read in that file (`topology_picture_ids`, `variant_picture_ids`):

```python
return set(session.exec(select(Picture.pixel_sha).where(...)).all())
```

The fixture's pictures carry no `pixel_sha` at all, so the module cannot see this. Worth a test where a fixed input names a picture that exists — it should report no `fixed_input_deleted`, and it goes red on the current code.

## Should fix — `allow_unchecked` silently drops the run's LoRAs

`routes/workflows.py`, `_plan`:

```python
if body.loras and slots_in_graph and object_info is not None:
    found += _apply_loras(graph, body.loras, object_info)
```

Unreachable ComfyUI + `allow_unchecked: true`: `_apply_loras` never runs, `judge` reports only `comfyui_unreachable` (slots exist, so no `no_lora_loader`), `blocks_group` filters that out under consent, and the group submits. The caller asked for LoRA X, consented only to "run without inspecting", and gets whatever LoRA the stored graph named — no reason code, no log line.

Same class of silent-wrong-output the PR correctly refused for `seed_mode: "keep"` on a seedless instance. Either refuse it, or call `apply_adapter` with `{}` and let its `LookupError` become `missing_models`, which the `_apply_loras` except branch already does. No test covers consent + `loras`.

## Minor

- `resolve_source` takes `hub` and `library_uuid` and uses neither; same for `hub` on `_groups_for`.
- The `for` over `instance_documents` always returns on the first iteration — it is `[0]` wearing a loop.
- `stack_for_picture` is a write task called inside the per-run loop in `_submit_every`. It is idempotent, so hoist it above the loop.
- `model_copy(update=…)` in `_with_recipe` skips validation: a stored `seed` above `MAX_SEED_64` and a merged `values` list above `MAX_DEFAULTS` both bypass the declared field constraints.
- A saved recipe with an empty `workflow_key` yields a group with `runs: 0` and `reasons: []`, contradicting "reasons empty is the only thing that means this would run". One `no_runnable_source` in `_groups_for` closes it.

## Checked and satisfied

Both routes `OWNER_ONLY`, justified in the registry and the matrix, neither in `READ_SAFE_POST_PATHS`, and the negative test uses `assert_real_route` with a live positive control — so the 403s are refusals, not dead paths. Narrowing below the `PICTURE_SCOPED` `comfyui.py` routes is the right call. No path collision with any `{workflow_key}` POST. `_plan` is the single decision point for both handlers, so a 4xx on one is a 4xx on the other. Consent reaches only the two uninspectable codes. `deepcopy` per run, one `object_info` fetch per request, the lease kwargs and the partial-failure return all match the shipped `run_recipe`.

On your open questions: agreed the nine `routes/comfyui.py` imports are not this PR's job. On pre-flight cost — `fetch_object_info` is not cached either, so a panel calling preflight per keystroke hits ComfyUI each time; worth naming in the `ponytail:` note alongside the file reads.

lindkvis:
Thanks — the blocker was real and your diagnosis was exact. All of it is addressed in `65249aeb`, except two I'm answering rather than changing.

## Blocker — `read_kept_pixel_shas`

Reproduced before touching it, against the real `Picture` model on sqlmodel 0.0.38: `session.exec(select(Picture.pixel_sha)).all()` returns `['aaa111']`, and `(sha,) = "aaa111"` raises `ValueError: too many values to unpack (expected 1)`. Exactly as you described, including that the feature was inverted — it only worked when the picture was gone.

Fixed to `set(session.exec(...).all())`, with a comment saying why, since the wrong form reads as natural.

Your read on why the module couldn't see it was also right: the fixture's pictures carry no `pixel_sha`. The runnable card's picture now has one (`RUN_PIXEL_SHA`), and there's a test where a fixed input names a picture that **exists** and asserts no `fixed_input_deleted`. It goes red on the old code; the existing deleted-case test stays green on it, which is the whole problem in one line.

I grepped for siblings of the class — `for (…,) in …exec(` across `pixlstash/` — and this was the only one. `best_picture_ids` was already correct.

## Should fix — consent silently dropping the LoRAs

Agreed, and it's the same silent-wrong-output class as the `seed_mode: "keep"` refusal.

I took neither option exactly, because they differ per slot kind. A **digest** slot needs no `object_info` — that node resolves the file itself — so it can be honoured and now is. A **filename** slot is resolved against what that ComfyUI lists, and with no list the run genuinely cannot do what was asked; reporting `missing_models` there would assert the file is absent when we don't know. So that case is a 400 naming the slot. The `object_info is not None` guard in `_plan` is gone, so `_apply_loras` answers for the state itself rather than being skipped.

Test covers both halves: consent + LoRA refuses and submits nothing; consent alone still runs and the graph's own `lora_name` is untouched.

## Minor

- Unused `hub` / `library_uuid` on `resolve_source` and `hub` on `_groups_for` — removed.
- The `for` over `instance_documents` — now an index, with a line saying the caller hands them best-first.
- `stack_for_picture` hoisted above the per-run loop. Test asserts one call for `count: 3`.
- `model_copy(update=…)` — replaced with a `_revalidated` helper that goes through `model_validate`, so a stored value that breaks a field ceiling is a 422 naming the recipe instead of an assignment. Test uses a stored seed of `2**64`.
- Pre-flight cost note now names the uncached `fetch_object_info` alongside the file reads.

## One I'm answering instead

**The empty `workflow_key` case is unreachable.** `POST /recipes` declares `workflow_key: str = Field(min_length=1, …)`, so no recipe with an empty key can be created through the API. The reachable neighbour — a recipe naming a card this hub doesn't hold — already answered `no_runnable_source` via `find_card`, and there's now a test pinning it.

I still closed the class, but at the one place every keyless group converges in `_plan` rather than in the saved-recipe branch, so the next producer inherits it. Being straight about it: I mutated that fallback and the suite stayed green, because every producer today already attaches a reason. It's a guard on future code, not a live fix, and the comment says so.

## Checks

`ruff` clean. 178 in `test_workflows_api.py`, plus guardrails, ci_shards, workflow_library and saved_recipes — 547 locally. Every new assertion mutation-checked; the one that passed under mutation is the unreachable fallback above, and I've said so rather than dressing it up.

PR #1442, opened 2026-09-19
Release paused handlers in a finally, across all six sites

### LINKED ISSUE #1441: Restore drain tests: release the paused handler in a finally, not after the asserts
## What

The two restore admission-drain tests release their paused handler **outside any
`try`**, so an orchestration assertion that fires leaves that handler blocked
for its full timeout, holding the scratch `snapshot.sqlite` open. On Windows the
next test's `snapshots` `rmtree` then fails, so one genuine failure becomes two.

- `tests/test_restore.py::test_full_restore_drains_an_admitted_share_before_swap`
- `tests/test_restore.py::test_full_restore_drains_an_admitted_http_request_before_swap`

Both look like this:

```python
restore_thread.start()
assert close_started.wait(20), "restore never began the admission drain"   # can fail
assert server.auth.is_auth_closed_for_restore()                            # can fail
assert not swap_started.wait(0.2), "database swapped before ... drained"   # can fail

release_share.set()          # never reached if any assert above fires
share_thread.join(timeout=10)
```

**This is not hypothetical and it is not new.** 807045a1 ("unstick the restore
queue-gap wait") fixed exactly this in the sibling
`test_full_restore_closes_auth_before_swap_and_across_queue_gap`, and its
comment states the consequence:

> Inside the try, so a restore that never arrives is still released and joined:
> left blocked, it holds the scratch snapshot.sqlite open and the next test's
> snapshots rmtree fails on Windows.

The two drain tests are the siblings that fix did not reach.

## Why now

#1435 raised both handlers' patience from `wait(10)` to `wait(120)`, because
they were less patient than the orchestration that releases them — a real
inversion that reddened `backend-windows shard 2` on a frontend-only diff.

That fix is right, and it **makes this one more urgent**: the stall on a failed
assertion went from 10 s to 120 s. The trade was deliberate and is worth taking
— a correct test that stalls on failure beats a wrong test that fails on
success — but the stall should not exist at all.

## The fix

The shape 807045a1 already established: put the orchestration in a `try` and
the release in a `finally`, so a failing assertion still unblocks and joins the
paused thread before it propagates.

## What does NOT need fixing, so nobody re-derives it

A grep for `release*.wait(` finds 11 other sites across `tests/`. **None of them
has this defect**, and a naive scan says otherwise, so the criterion is worth
writing down:

> The defect is not "a short release timeout". It is a handler paused on
> `release.wait(P)` while the main thread performs **another bounded wait, after
> the handler is already paused**, of length comparable to or greater than `P`,
> before releasing. The main thread's wait for the handler's own *entered* event
> does not count — it has already returned by the time the handler's clock
> starts mattering.

By that criterion the 11 others are all `after-pause = 0s`: they wait for the
handshake, then do quick work (a couple of HTTP calls and asserts) and release.
Four were also read by hand to confirm the scan
(`test_library_switch.py:658`, `test_authz_library_pin.py:212`,
`test_libraries_routes.py:406`, `test_model_shelf_api.py:1159`).

The restore pair were the only inverted ones because their orchestration
contains a *second* long wait — `close_started.wait(20)` — that happens after
the pause.

## Optional, if it is judged worth the machinery

A guardrail could pin the criterion so the inverted shape cannot come back, in
the spirit of `tests/test_architecture_guardrails.py`. It would have to parse
the orchestration to be meaningful, which may cost more than it saves — the
criterion written down above may be enough.

lindkvis:
Closes #1441.

## What

A test handler parked on `release.wait(P)` was released **outside any `try`**, so an orchestration assertion that fired left it blocked for the full timeout. The orchestration now sits in a `try` and the release in a `finally`, the shape 807045a1 established.

## Six sites, not two

The issue named the two restore drain tests and stated a criterion clearing the other 11 `release*.wait(` sites. That criterion vets **timeout length** (the 10→120 s class of #1435); the defect here is **release-outside-try**, which is independent of P. Four siblings have it:

| file | pause | release |
|---|---|---|
| `tests/test_library_switch.py` | 658 | 693 |
| `tests/test_authz_library_pin.py` | 212 | 251 |
| `tests/test_libraries_routes.py` | 406 | 437 |
| `tests/test_model_shelf_api.py` | 1159 | 1191 |

Two of them carry a **larger blast radius than the reported bug**: the trailing `switch_to(original.uuid)` was also unguarded, so a failed assertion left every later test in the module running against the wrong library. That restore moves into a `finally` too. `test_libraries_routes.py:533` released correctly but joined outside, which leaves the switch still running into the next test; its joins move in as well.

Also: `restore_thread.join(timeout=30)` exactly equalled `_RESTORE_DRAIN_TIMEOUT_SECONDS = 30.0`, so a wedged restore returned from a silently-expired join. Raised to 90 s, past the drain budget with room for the swap.

## Measured

Mutating one orchestration assertion, with the handler paused:

| | before | after |
|---|---|---|
| `test_restore.py` share drain | 3 failed, 2 m 05 s wall | 1 failed, 5 s |
| `test_library_switch.py` | 2 failed + 7 errors | 1 failed, 41 passed |

Unmutated: 480 passed across the five files. Ruff format + check clean.

## Reviewer objections not taken

- **A guardrail pinning the shape.** It has to parse orchestration to mean anything, and the issue flags it as possibly costing more than it saves. Six sites now share one shape; a seventh is a review catch.
- **One shared context manager for all six.** The shapes differ — some join one thread, some two, two restore a library, one awaits a task. The helper plus its six call sites is longer than six local `try`/`finally` blocks.
- **The lost `authentication remains closed` diagnostic.** That line came from the drain timing out. With the release in the `finally` the drain no longer times out, so the condition it reported does not arise.
- **No new assertion guards the fix.** It is pure control flow; there is no assertion to add. The mutation evidence above is the check, run by hand.


lindkvis:
## Review — approve

Compared `origin/develop...` head `03222409` (merge base `1345783e`).

Verified: the 7 touched tests pass at head (30.5 s); ruff check + format clean;
mutating `assert not switch_done.is_set()` at `tests/test_library_switch.py:692`
fails in **5.65 s, 1 failed** — no 10 s park, no cascade into the rest of the class.

The shape is not an invention: the conditional restore and the
`if thread.ident is not None` guard already ship at
`tests/test_libraries_routes.py:384` and
`tests/test_library_destination_blocklist.py:528`.

Three notes, none blocking.

### 1. The sweep misses two more release-outside-try sites

"Six sites, not two" reads as exhaustive. Two others have the same defect:

| file | pause | release | park on failure |
|---|---|---|---|
| `tests/test_dedup_tier_service.py` | 1327 | 1352 | 2 s, plus a `TaskRunner` whose `stop()` thread is left unjoined |
| `tests/test_reference_folder_watcher.py` | 46 | 90, 115 | 5 s, daemon observer thread |

Short waits and small blast radius, so leaving them is defensible — but the
dedup one leaks a running `TaskRunner` into the next test, which is the residue
the moved joins exist to prevent. Fold them in, or say in the body that they
were seen and deliberately left.

The remaining five sites (`test_folder_structure_read.py:1032`,
`test_library_destination_blocklist.py:511`, `test_pictures_stream.py:218`,
`test_pipeline_stage_contracts.py:338`, `test_work_planner.py:132`) already
release in a `finally`.

### 2. `test_model_shelf_api.py` releases but does not await

`release.set()` moved into the `finally`, `_await_scan` did not. On a failure
the released `ModelFolderScanTask` runs on into the next test on the shared
`ModelFolderScanTask` progress lane — the same "released but unjoined" argument
given for the joins at `test_libraries_routes.py:541`. `_await_scan(...)` in the
`finally` closes it.

### 3. The outer restore `finally` can now raise where it previously did not

`test_library_switch.py:704` / `test_authz_library_pin.py:266`: if the inner
`switch_thread.join(timeout=20)` expires, the switch is still in flight when the
outer `finally` calls `active_library()` / `switch_to()`, which can raise
"already in progress" from the `finally`. Python chains it, so the original
assertion still prints, but the headline exception becomes the restore failure.
Previously that path just skipped the restore. Right trade — an unrestored
library is worse — worth knowing when triaging a red run.

### Observation

`restore_thread.join(timeout=90)` is right for the pass path (30 s
`_RESTORE_DRAIN_TIMEOUT_SECONDS` at `pixlstash/auth.py:543`, plus swap headroom).
On the failure path the `is_alive` assert the comment cites is never reached —
the orchestration exception propagates out of the `finally` first — so the 90 s
is pure wall-clock there, worst case ~120 s. Only reachable when the restore is
genuinely stuck. Leave it.

### The four objections

All answered correctly: no guardrail (it would have to parse orchestration), no
shared context manager (six differing shapes; the helper is longer than the six
blocks), the dropped `authentication remains closed` diagnostic (the condition
no longer arises), and no new assertion (pure control flow; the mutation run is
the check).

lindkvis:
Fixed in `4c1399a3`.

**1. The two remaining sites** now release in a `finally`:
`tests/test_dedup_tier_service.py` (plus `runner.stop()` on the branch where the
slice never began, since nothing else would) and both
`tests/test_reference_folder_watcher.py` tests.

**2. `test_model_shelf_api.py`**: `row = _await_scan(...)` moved into the
`finally` beside `release.set()`.

**3.** Left as designed — an unrestored library is worse than a chained
exception, and the alternatives either drop the restore or add machinery. The
`restore_thread.join(timeout=90)` comment is corrected to say the `is_alive`
assert is the *passing* path's reader.

### Correction to my own severity claims

I ran the same mutation before and after each fix, whole file:

| | pre-fix | post-fix |
|---|---|---|
| `test_dedup_tier_service.py` | 1 failed, 83 passed, 69.09 s | 1 failed, 83 passed, 69.27 s |
| `test_model_shelf_api.py` | 1 failed, 247 passed, 13.45 s | 1 failed, 247 passed, 13.36 s |
| `test_reference_folder_watcher.py` | 1 failed, 2 passed, 0.12 s | 1 failed, 2 passed, 0.12 s |

Identical. The leaked `TaskRunner` and the scan running into the next test are
both real in the code and neither is observable in these runs — the parks are
2 s and 5 s on daemon threads against per-test objects. So these three are shape
consistency, not the measurable win the original comment implied. Worth having
so the next assertion added to those blocks inherits the guarantee; not worth
claiming more than that.

Unmutated: 5 passed on the touched tests, plus the 7 from the original review.
Ruff format + check clean.

PR #1443, opened 2026-09-19
Compare release tag and pyproject version as PEP 440, not strings

lindkvis:
v1.12.0-dev.1 and v1.12.0-dev.2 both failed the Windows installer and Electron builds: the tag says `-dev.2`, pyproject says `.dev.2`, and the workflows compared them as raw strings. The format regex above the comparison accepts either separator, so the tag passed the format check and failed the equality check.

Both workflows now call `scripts/check_release_tag.py`, which applies the same regex and then compares PEP 440 normalised versions. Same accepted shapes, minus the separator trap.

Also: `\A`/`\Z` anchors rather than `^`/`$` (Python's `$` matches before a trailing newline, which the bash check rejected), and quoted `$env:` arguments on the pwsh side.

The inline logic had no behavioural test. It has 12 gated cases now, each mutation-checked.


.github/workflows/electron.yml
```diff
@@ -99,11 +99,7 @@ jobs:
             echo "Release ref/tag mismatch: $RELEASE_REF / $RELEASE_TAG" >&2
             exit 1
           fi
-          if [[ ! "$RELEASE_TAG" =~ ^v[0-9]+\.[0-9]+\.[0-9]+((a|b|rc)[0-9]+|[.-]dev[.-]?[0-9]+)?$ ]]; then
-            echo "Release tag has an unsupported format: $RELEASE_TAG" >&2
-            exit 1
-          fi
-          [ "$RELEASE_TAG" = "v$EXPECTED_VERSION" ] || { echo "tag/version mismatch"; exit 1; }
+          python scripts/check_release_tag.py "$RELEASE_TAG" "$EXPECTED_VERSION"
```
lindkvis, line 103: **The third call site still compares raw strings, and this PR makes that worse.**

`publish-pypi.yml` lines 42-53 run the identical logic: the same regex (which accepts `-dev`) and then `if tag != expected` where `expected = f"v{data['project']['version']}"`. With `pyproject.toml` at `1.12.0.dev.3`, a `v1.12.0-dev.3` tag now passes Electron and the Windows installer and still exits 1 in Publish PyPI - so the release gets installers attached and no wheel, which is a worse outcome than today's symmetric failure. Verified locally: the publish-pypi regex fullmatches the tag and the equality is False.

CLAUDE.md: *"Fix the class, not the reported instance. A report names one site; grep for the population and state the count before editing."* The population is three; two are converted.

scripts/check_release_tag.py
```diff
@@ -0,0 +1,55 @@
...
+release workflows used to compare the two as raw strings, which accepted the
+``-dev`` spelling in the format check and then rejected it in the equality
+check: v1.12.0-dev.1 and v1.12.0-dev.2 both failed there with the tag and the
+version naming the same release.
+"""
+
+import re
+import sys
+
+from packaging.version import InvalidVersion, Version
```
lindkvis: **The release gate's only check now depends on a package neither workflow installs.**

`packaging` is not in `pyproject.toml`'s dependencies and setup-python 3.12 ships pip only. It happens to be importable here solely because the *earlier* `Build wheel` step ran `pip install --upgrade pip build`, and `build` pulls it in transitively. Move this validation step ahead of the wheel build - the obvious fail-fast refactor for a gate - and the release dies with a `ModuleNotFoundError` traceback and exit 1, which reads exactly like a tag mismatch. `release-version.yml` already guards against this with an explicit `pip install --upgrade pip packaging`; either do the same in both workflows or add `packaging` to the project dependencies.

scripts/check_release_tag.py
```diff
@@ -0,0 +1,55 @@
...
+    """Return 0 if ``argv`` is a supported tag naming the expected version."""
+    if len(argv) != 2:
+        print("usage: check_release_tag.py <tag> <expected-version>", file=sys.stderr)
+        return 2
+    tag, expected = argv
+    if not TAG_PATTERN.match(tag):
+        print(f"Release tag has an unsupported format: {tag}", file=sys.stderr)
+        return 1
+    try:
+        if Version(tag[1:]) != Version(expected):
```
lindkvis: **A dropped invariant: `expected` is no longer shape-checked.**

The old `[ "$RELEASE_TAG" = "v$EXPECTED_VERSION" ]` transitively forced the pyproject version through the tag regex, so the string every artifact is named with was known well-formed. PEP 440 comparison is case- and separator-insensitive, so `1.12.0_DEV_2` in pyproject now matches a `v1.12.0-dev.2` tag and ships as `pixlstash-server-windows-1.12.0_DEV_2.exe` and as Inno Setup's `PIXLSTASH_VERSION`, while the wheel normalises to `1.12.0.dev2`. Given the repo already hand-writes non-canonical versions (`1.12.0.dev.2`), re-establish the guard:

```suggestion
    if not TAG_PATTERN.match(f"v{expected}"):
        print(
            f"pyproject version has an unsupported format: {expected}", file=sys.stderr
        )
        return 1
    try:
        if Version(tag[1:]) != Version(expected):
```

.github/workflows/windows-installer.yml
```diff
@@ -69,14 +69,8 @@ jobs:
...
         run: |
-          $tag = $env:RELEASE_TAG
-          $expected = "v$env:EXPECTED_VERSION"
-          if ($tag -notmatch '^v[0-9]+\.[0-9]+\.[0-9]+(?:(?:a|b|rc)[0-9]+|[.-]dev[.-]?[0-9]+)?$') {
-            Write-Error "Release tag has an unsupported format: $tag"
-          }
-          if ($tag -ne $expected) {
-            Write-Error "Tag $tag does not match pyproject version $expected"
-          }
+          python scripts/check_release_tag.py "$env:RELEASE_TAG" "$env:EXPECTED_VERSION"
```
lindkvis, line 73: **Nothing pins this call site.** `test_electron_apple_signing_requires_validated_release_tag` was updated to assert `scripts/check_release_tag.py` appears in electron.yml's step, but no test asserts it for windows-installer.yml - even though this file is in `PRIVILEGED_WORKFLOW_PATHS` (tests/test_ci_shards.py:85). A revert to the `-ne` string compare here, or a rename of the script, is silent. One assert in the same file covers it.

tests/test_ci_shards.py
```diff
@@ -404,8 +404,7 @@ def test_electron_apple_signing_requires_validated_release_tag():
     assert "validated_release_tag=false" in validation["run"]
     assert "validated_release_tag=true" in validation["run"]
     assert '"refs/tags/$RELEASE_TAG"' in validation["run"]
-    assert "^v[0-9]+" in validation["run"]
-    assert '"v$EXPECTED_VERSION"' in validation["run"]
+    assert "scripts/check_release_tag.py" in validation["run"]
```
lindkvis: **This is weaker than the two assertions it replaces and is mutation-passable.**

`'"v$EXPECTED_VERSION"' in validation["run"]` proved the comparison actually consumed the resolved version. A substring check on the script path passes for `python scripts/check_release_tag.py "$RELEASE_TAG" "$RELEASE_TAG"`, which accepts every well-formed tag. The `validation["env"]` assert above proves the variables are declared, not that they are used. Assert the whole invocation:

```suggestion
    assert (
        'python scripts/check_release_tag.py "$RELEASE_TAG" "$EXPECTED_VERSION"'
        in validation["run"]
    )
```

scripts/check_release_tag.py
```diff
@@ -0,0 +1,55 @@
...
+import re
+import sys
+
+from packaging.version import InvalidVersion, Version
+
+# ``\A``/``\Z`` rather than ``^``/``$``: Python's ``$`` also matches before a
+# trailing newline, which would accept a tag the shell checks this replaced
+# rejected.
+TAG_PATTERN = re.compile(
+    r"\Av[0-9]+\.[0-9]+\.[0-9]+((a|b|rc)[0-9]+|[.-]dev[.-]?[0-9]+)?\Z"
```
lindkvis, line 42: **This is the sixth copy of this regex, not the one copy it could have been.**

The same pattern lives in `docker-publish.yml:76`, `docker-publish.yml:200`, `release-version.yml:30`, `release-test-issues.yml:36` and `publish-pypi.yml:42`. Extracting the logic into a script was the moment to collapse them - a `--format-only` flag (or just calling this with the tag twice) would let every one of those steps call it. As it stands, adding a new accepted shape here leaves five call sites still rejecting it.

tests/test_ci_shards.py
```diff
@@ -1760,3 +1759,43 @@ def test_a_complete_map_warns_about_nothing():
...
+    tag, expected_version, accepted
+):
+    """A dev tag's separator must not decide whether a release builds.
+
+    ``v1.12.0-dev.1`` and ``v1.12.0-dev.2`` both failed the Windows installer
+    and Electron builds because the workflows compared the tag with the
+    pyproject version as raw strings, while the format check above it accepted
+    either separator. The two spellings are the same PEP 440 version.
+    """
+    result = subprocess.run(
```
lindkvis: **11 interpreter launches for 11 pure-function cases.**

This file already has the cheaper pattern 240 lines up - `test_duration_parser_reads_the_ids_pytest_actually_prints` does `sys.path.insert(0, str(REPO_ROOT / "scripts"))` and imports the function. Importing `main` and asserting `main([tag, expected]) == 0` gives identical coverage (the `__main__` guard is one line of `sys.exit`) at roughly a hundredth of the cost, and CLAUDE.md is explicit: *"The expensive thing in this suite is the environment rebuild, not the test."* Keep at most one subprocess case if you want the exit-code plumbing covered end to end.

scripts/check_release_tag.py
```diff
@@ -0,0 +1,55 @@
...
+        print(f"Release tag has an unsupported format: {tag}", file=sys.stderr)
+        return 1
+    try:
+        if Version(tag[1:]) != Version(expected):
+            print(
+                f"Tag {tag} does not match pyproject version {expected}",
+                file=sys.stderr,
+            )
+            return 1
+    except InvalidVersion as exc:
```
lindkvis: **The only untested branch, and the only one that can fire on a bad `pyproject.toml`.**

All 11 gated cases pass a valid `expected`, so nothing exercises `InvalidVersion` - which is reachable exactly when someone typos the version bump, i.e. the scenario this script exists for. CLAUDE.md asks for every new assertion to be shown failing; a case like `("v1.12.0", "not-a-version", False)` anchors it.

.github/workflows/windows-installer.yml
```diff
@@ -69,14 +69,8 @@ jobs:
...
+          if ($LASTEXITCODE -ne 0) { exit 1 }
```
lindkvis: Dead line: GitHub appends `if ((Test-Path -LiteralPath variable:\LASTEXITCODE)) { exit $LASTEXITCODE }` to every `pwsh` step, and python is the last command in this script, so a non-zero exit already fails the step. All this adds is flattening the usage exit code (2) to 1.

scripts/check_release_tag.py
```diff
@@ -0,0 +1,55 @@
...
+)
+
+
+def main(argv: list[str]) -> int:
+    """Return 0 if ``argv`` is a supported tag naming the expected version."""
+    if len(argv) != 2:
+        print("usage: check_release_tag.py <tag> <expected-version>", file=sys.stderr)
+        return 2
+    tag, expected = argv
+    if not TAG_PATTERN.match(tag):
```
lindkvis: `.match()` here is only a full-string check because of the `\Z` you added. A future edit that restores `^`/`$`-style anchors, or drops them as redundant-looking, silently turns this into a prefix match that accepts `v1.12.0-anything`. `TAG_PATTERN.fullmatch(tag)` is free and makes the anchors belt-and-braces rather than load-bearing.

scripts/check_release_tag.py
```diff
@@ -0,0 +1,55 @@
...
+        print(f"Release tag has an unsupported format: {tag}", file=sys.stderr)
```
lindkvis: House style for a failing release gate is a GitHub annotation, not bare stderr: `publish-pypi.yml` uses `print(f"::error::...")` so the reason shows on the run summary and the commit, rather than only in the step log. Same for the two messages below.

.github/workflows/electron.yml
lindkvis: Right, and this was the important one. The population is three and I had converted two; my grep used `EXPECTED_VERSION` and `pyproject`, and this site uses a local `expected` and says "project version", so it matched neither. I reproduced your scenario before fixing it.

`a7cfe690` routes publish-pypi through the same checker, keeping its ref check inline. `test_every_release_workflow_validates_the_tag_the_same_way` now pins all three call sites; reverting this one to a string compare turns it red.

scripts/check_release_tag.py
lindkvis: Confirmed, and worse than "would break under a refactor": publish-pypi validates *before* `pip install build`, so `packaging` was already absent there the moment I routed that site through the script. `a7cfe690` installs it explicitly in all three, and the new cross-workflow test asserts each one does.

scripts/check_release_tag.py
lindkvis: Correct, and I had not seen that the old string compare enforced this transitively. `Version()` treats `_` as a separator and is case-insensitive, so `1.12.0_DEV_2` compared equal to `v1.12.0-dev.2` while still naming the installer and `PIXLSTASH_VERSION`.

`a7cfe690` shape-checks `f"v{expected}"` through the same pattern. Three cases cover it; removing the check turns `[v1.12.0-dev.2-1.12.0_DEV_2]` red.

.github/workflows/windows-installer.yml
lindkvis: Agreed. `a7cfe690` pins it in `test_every_release_workflow_validates_the_tag_the_same_way`, which asserts the exact invocation for all three workflows rather than only electron.

tests/test_ci_shards.py
lindkvis: Fair, and your mutation is the one that matters. `a7cfe690` asserts the full invocation including the argument pair, so `"$RELEASE_TAG" "$RELEASE_TAG"` now fails both this test and the new cross-workflow one. I ran that mutation to confirm.

tests/test_ci_shards.py
lindkvis: Agreed, and CLAUDE.md is explicit about it. `a7cfe690` uses the in-process import already used by `test_duration_parser_reads_the_ids_pytest_actually_prints`; `main([tag, expected])` is identical coverage with no interpreter start-up.

.github/workflows/windows-installer.yml
lindkvis: Correct, GitHub appends the `exit $LASTEXITCODE` wrapper and python is the last command. Removed in `a7cfe690`, which also restores the distinction between exit 1 (bad tag) and exit 2 (bad invocation).

scripts/check_release_tag.py
lindkvis: Agreed. `a7cfe690` uses `fullmatch`. I kept `\A`/`\Z` alongside it deliberately so neither is load-bearing alone, with a comment saying why: Python `$` would also accept a trailing newline that the shell check this replaced rejected. The `v1.12.0\n` case covers it.

scripts/check_release_tag.py
lindkvis: Agreed, publish-pypi was already doing this. `a7cfe690` routes every failure through a `_fail()` helper that emits `::error::`.

scripts/check_release_tag.py
lindkvis: You are right that it was the only untested branch, but adding the case you suggested would have tested nothing: once `a7cfe690` shape-checks `expected` too, both operands must match the tag pattern before `Version()` is reached, and every shape that pattern admits parses. I enumerated them (all prefix/dev separator combinations, leading zeros, long numeric components) and found no input that reaches `InvalidVersion`.

So it was dead code, and `3a1110a1` deletes it rather than testing an unreachable path. Fail-closed is unchanged: if the pattern is ever loosened past PEP 440, the unhandled raise still exits non-zero.

scripts/check_release_tag.py
lindkvis: Half agreed, and I have taken the half I think is real.

`a7cfe690` removes publish-pypi's copy, so it is 4 files now rather than 6: docker-publish.yml (twice), release-test-issues.yml, release-version.yml and the script.

Where I push back is the failure you describe. None of those three workflows reads pyproject.toml — I checked — so none of them compares a tag against a version; they are format-only gates on a tag they have already been handed. The bug this PR fixes cannot occur in them, and collapsing them would mean giving the script a format-only mode for callers that are not currently broken. The three that did compare are now one implementation.

That leaves your real point standing: a newly accepted tag shape would still have to be added in four places. That is worth doing and I do not think it belongs in a release-blocking fix. Leaving this thread open for a human rather than resolving it.

PR #1445, opened 2026-09-19
Workflows & Recipes F3: the Workflow inspector tab

### LINKED ISSUE #1403: Workflows & Recipes F3: Workflow inspector tab

## Depends on

|---|---|---|
| #1402 (F1a) | yes | `git cat-file -e origin/develop:frontend/src/components/views/WorkflowsView.vue` |
| #1396 (B4) | yes | `git grep -q CHANGED_WORKFLOWS origin/develop -- pixlstash` |
| #1400 (B8) (only for the Insert loader entry) | no | `git grep -q insert-lora-loader origin/develop -- pixlstash/authz/registry.py` |

## What

Reshape `panels/WorkflowInspector.vue` into the **Workflow** tab on `AppInspector`. There is no Run tab.

- **Header:** title plus "In the X stack · N pictures", or "Showing the cover: X" when the stack itself is selected. The "N pictures" figure is a link (F7 wires the chip). Show "M deleted" when there are some.
- **Models, all editable:**
  - Checkpoint.
  - VAE ("From the checkpoint").
  - LoRA slots, **two lines per slot**: name + 28px strength, then a full-width `Segmented` Workflow/Recipe.
  - *Insert loader*, only if B8 shipped it.
- **Defaults = shown in Run:** each value with provenance ("from your best pictures" / "edited by you" + reset) and a pin. Value fields are `--control-h`, and the 96px label column is local.
- *All N parameters*, Notes, *Node names for ComfyUI users* disclosure. The input-mode UI moves here.
- **Footer:** **Run…**, plus ⋯ with Duplicate, Open in ComfyUI, Export workflow, Hide, and Delete (imported only). With several workflows selected, Run… stays visible but disabled (`aria-disabled` + `aria-describedby`), reason "Run one workflow at a time", and the inspector shows "N workflows selected" with *Stack together* and *Hide*.
- A selected member shows **that member** on this tab.
- Until F5 lands, Run… may open the existing run path.

## Acceptance

- Tests for provenance/reset, a mark flip calling the slots route, and the disabled multi-select Run.

lindkvis:
Closes #1403 (v1.12 Workflows & Recipes, F3).

Adds the **Workflow** tab to the Workflows grid's right rail on
`/workflows-next`: header, Models (checkpoint, VAE, LoRA slots with a
full-width Workflow/Recipe switch), Defaults with provenance, reset and a pin,
the *All N parameters* / Notes / *Node names* disclosures, and a footer with
Run… and ⋯. With several selected, Run… stays on screen and refuses
(`aria-disabled` + `aria-describedby`, "Run one workflow at a time") beside
*Stack together* and *Hide*. A selected stack member shows that member.

Backend, both needed by the above and both additive:

* `GET /workflows/cards/{key}` now returns `pins`. `PUT .../pins` shipped with
  nothing reading it back, so the pin could be set and never drawn. `null`
  (no choice) and `[]` (everything unpinned) stay different answers, and an
  unreadable row answers `null` rather than `[]`.
* Each entry of a card's `models`/`loras` carries `slot_label`. Without the
  address the client cannot say which slot it just marked.

## Two departures from the issue

**It is a new component, not a reshaping of `WorkflowInspector.vue`.** That
file still serves the shipped shelf on `/workflows`, and every step here is
required to leave the shelf working until F1b — which deletes
`WorkflowShelf.vue`, its store and its inspector together. Reshaping it in
place would have taken the shelf's rail out now for no gain.

**Six drawn items are not in this PR**, each because the data or the route is
absent, not to save work. Editable checkpoint/VAE (no route sets a card's
model file); LoRA strength (`_describe_slots` reads filenames, and
`FEATURED_NAMES` excludes `strength_model`); "M deleted" (no deleted count on
a card); the "N pictures" link (F7 wires it — the route exists but the picture
grid has no client-side `workflow_key` filter, so a link would land on an
unfiltered library); the picture-input UI (the stored overrides are readable,
but the list of picture inputs a card *has* is served only by the file-keyed
`GET /comfyui/workflows/{name}/inputs`); and ⋯ Duplicate / Open in ComfyUI /
Export workflow / Delete. The menu holds Hide/Unhide alone rather than showing
entries that would do nothing. All of it is written down in
`docs/frontend_architecture.md` §F3.

Run… opens the shipped run panel, which the issue permits until F5. It is not
scoped to the selected card, which is the honest state of the existing run
path rather than something this PR chose.

## Tests

22 frontend tests and two extended backend ones. Eighteen mutations were run
against the new logic; each turned a named assertion red and was restored.
Four assertions written first survived inversion and were rewritten to pair a
value to its own row rather than count across the panel.

## Where I disagree with the pre-push review

* `useWorkflowsStore.fetchCards`' early return while a fetch is in flight can
  swallow a refresh. Real, but it is the shipped store's behaviour from F1a and
  every caller has it; fixing it here would be an unrelated change to a
  function three other call sites share.
* "Export workflow has a route." `GET /workflows/recipes/{hash}/graph` exists
  but answers `runnable: false` by construction — the document has its
  parameters, seeds and prompts nulled and its assets named opaquely. Export
  has to write a file ComfyUI can open, so that read is not it.

## Size

~1500 lines against the brief's ~600. About 1100 of it is the new component,
its CSS and its tests; I would rather flag the overrun than drop the tests.


frontend/src/stores/useWorkflowsStore.js
```diff
@@ -199,6 +199,20 @@ export const useWorkflowsStore = defineStore("workflows", () => {
...
+   *
+   * For the one gesture that invalidates them wholesale: marking a LoRA slot
+   * re-keys every card of the topology, so the cards cached under a cover key
+   * can name workflows the hub no longer has. `fetchCards` does not clear
+   * them — it only re-reads the grid, which lists covers — so an open stack
+   * would keep drawing its pre-flip members until the session reset.
+   */
+  function forgetMembers() {
+    members.value = {};
+    membersFailed.value = new Set();
```
lindkvis, line 233: Two problems here, both reproduced against this head.

**1. An open stack panel ends up claiming its members could not be read.** Nothing re-opens the stack, so `openMembers` falls back to `[cover]`, `membersLoading` is `false`, and `StackPanel.vue` computes `missing = size - members.length` and renders "2 could not be read" — a failure message for a deliberate invalidation, with no way back except collapsing and re-expanding.

This lands on this PR's own headline flow: selecting a member *inside an open stack panel* is what the tab adds, and flipping that member's LoRA switch is the gesture it exists for. Worse when the flip re-keys the cover out of the grid: `openMembers` is then `[]` while `openStackKey` is still set, so the panel stops being drawn at all and the grid row keeps `aria-expanded="true"`.

**2. An in-flight `openStack` puts the stale members straight back.** `inflight` is left alone and `epoch` is not bumped, and `openStack`'s completion path guards on `mine === epoch` only — so a member read that was on the wire when the flip fired writes the *pre-flip* members back into `members`. That is exactly the stale state the docstring above says this function prevents. Expand a stack, call `forgetMembers()` before the member reads resolve, and `store.members` comes back holding the old cover key.

`reset()` below is the shape to copy — it bumps `epoch`, clears `inflight`, and sets `loading = false` because the old epoch's `finally` will not:

```js
function forgetMembers() {
  epoch += 1;
  loading.value = false;      // the old epoch's `finally` will not clear it
  members.value = {};
  inflight.value = new Set();
  membersFailed.value = new Set();
  closeStack();               // finding 1
}
```

`closeStack()` is the laziest answer to finding 1: the stack collapses, the false message never appears, and re-expanding fetches fresh members through the path that already exists. Re-opening it automatically is the bigger fix and needs a decision about a cover that is no longer a stack.

The `loading = false` line also closes the swallowed-refresh hole the PR body sets aside: every call site does `forgetMembers()` then `await store.fetchCards()`, and that call currently early-returns to nothing whenever a grid read is already in flight.

frontend/src/components/panels/WorkflowTab.vue
```diff
@@ -0,0 +1,873 @@
...
+    const body = await setWorkflowPins(key, pins);
+    if (stillOn(key)) detail.value = { ...detail.value, pins: body.pins ?? pins };
+  } catch (err) {
+    fail(err, "Could not change that pin.");
+  } finally {
+    busy.value = "";
+  }
+}
+
+async function saveNotes() {
```
lindkvis: This is the one write that does not `claim()`. Clicking a pin is what blurs the textarea, so "edit notes → click a pin" puts a PATCH and a PUT in flight together. The PATCH answers with the whole detail including the *pre-toggle* pins, so if it lands second the pin visibly flips back while the server holds the right value — precisely the failure `claim()`'s own docstring describes, with the notes save sitting just outside it.

frontend/src/components/panels/WorkflowTab.vue
```diff
@@ -0,0 +1,873 @@
...
+ * `allSettled`, and the grid is re-read whatever happened: a loop that threw
+ * on the first refusal left the ones before it hidden on the server, still
+ * drawn in the grid, still selected, under one sentence saying none of it
+ * worked. The message names how many actually went.
+ */
+async function hideSelected() {
+  if (!claim("hide")) return;
+  const keys = [...store.selectedKeys];
+  try {
+    const results = await Promise.allSettled(
```
lindkvis: N concurrent PATCHes, and each one runs a full `read_grid()` inside `_read_detail` — whose own docstring says the grid read is paid "once, on a gesture a person made, rather than per card". Each also `_announce`s separately.

There is no bulk-hide route, so the options are to sequence them or to add one; firing them concurrently is the one choice that maximises both grid reads and single-writer contention. Not blocking, and the `allSettled` handling itself is right.

frontend/src/components/panels/WorkflowTab.vue
```diff
@@ -0,0 +1,873 @@
...
+
+.wftab-chip--empty {
+  border-style: dashed;
+  background: transparent;
+  color: rgba(var(--v-theme-on-surface), var(--opacity-text-secondary));
+}
+
+/* A local track, like the 96px label column above: wide enough for a
+   strength ("0.85") and no wider. Not a spacing token used as a width. */
+.wftab-strength {
```
lindkvis: This draws a 56px bordered field showing `—` beside every LoRA, and the PR correctly explains no strength data exists (`FEATURED_NAMES` has no `strength_model`). An empty input-shaped box reads as broken rather than as "not recorded"; a plain caption, or nothing until the data lands, says the same thing without drawing a control that does nothing.

frontend/src/components/panels/WorkflowTab.vue
```diff
@@ -0,0 +1,873 @@
...
+}
+
+.wftab-value {
+  display: flex;
+  align-items: center;
+  min-height: var(--control-h);
+  padding: 0 var(--space-3);
+  border: 1px solid rgb(var(--v-theme-divider));
+  border-radius: var(--radius-sm);
+  background: rgba(var(--v-theme-on-surface), 0.04);
```
lindkvis: `rgba(var(--v-theme-on-surface), 0.04)` appears four times here and once in `WorkflowDefaultRow.vue`. The repo's surface-fill idiom is `0.06` / `0.08` (`ProjectFiles.vue`, `DedupPictureStrip.vue`, `ConnectAgentDialog.vue`, `StackExpansionStrip.vue` and others), so `0.04` is a new off-ramp value. Trivial, but it is the kind of drift the design manual exists to stop.

lindkvis (commented):
Reviewed `origin/develop...b5211586`. Backend is sound: no new route, so no `ROUTE_POLICIES` entry and no authz surface; `key_pins` keeps `null` and `[]` apart and type-checks outside the `try` (a stored scalar parses fine and then is not a pin list); `pins` lands in `_read_detail`, so the PATCH and the defaults PUT answer with it too and `detail.value = body` cannot silently drop it. `card_defaults` unions `seen | overrides`, so no edited value can sit outside `defaults` and be dropped by the whole-set reset. The tests are the good kind — paired to their own row rather than counted across the panel.

Green on the new frontend suite (22), the two extended backend tests, the architecture and CI-shard guardrails (202), `ruff check` / `ruff format --check`, and `npm run lint`.

**Two findings, both in `forgetMembers()`**, and one fix covers both. Both reproduced against this head with a throwaway store test. The rest of the comments are smaller notes, none blocking.

On the PR body's two disagreements: both hold. `fetchCards`' early return is shipped F1a behaviour shared by three other call sites; `GET /workflows/recipes/{hash}/graph` does answer `runnable: false` by construction, so it is not an export. Note that the fix below closes the `fetchCards` hole here for free.

Size is fine — the overrun is the component, its CSS and its tests, and the tests are the part worth having.

frontend/src/stores/useWorkflowsStore.js
lindkvis: Both right, both fixed in `5cb81f69`, and I took your shape almost verbatim.

`forgetMembers()` now bumps the epoch, clears `loading` and `inflight`, empties the maps and calls `closeStack()`. I confirmed finding 1 against `StackPanel.vue:129` — `missing = size - members.length` with `loading` false is exactly the "2 could not be read" path — and finding 2 is real for the reason you give: `openStack`'s completion guards only on `mine === epoch`, so clearing `inflight` alone would not have stopped the write-back.

Three tests in `useWorkflowsStore.test.js`, one per half plus the `loading` line:

- *collapses the open stack rather than leaving it reporting a failure* — `openStackKey` null and `openMembers` empty after the call.
- *drops a member read that was already on the wire, and can read again* — releases a member read after `forgetMembers()`, asserts `members` stays empty, then re-opens the stack and asserts the members that land are the fresh ones. That second half is what holds the `inflight` clear: without it the re-open returns early for ever, and asserting on `store.inflight` would not work anyway since the store does not expose it.
- *leaves a grid read able to start again* — the swallowed-refresh hole you point out. Mutating `loading.value = false` away turns it red.

Each of the three goes red under a targeted mutation (drop `closeStack()`, drop the epoch bump, drop the `loading` line) and green again on restore.

You are right that this closes the `fetchCards` early-return hole for the call sites in this component for free. I have left `fetchCards` itself alone, since the guard is shipped F1a behaviour that three other call sites share.

frontend/src/components/panels/WorkflowTab.vue
lindkvis: Right about the hole, and I took the invariant — but I queued the writes instead of refusing the second, because your own example is what argues against refusing.

Blur fires before click, so in "edit notes → click a pin" the notes save reaches `claim()` first and the pin click is the one refused. Adding `claim()` to `saveNotes` would have traded "the pin visibly flips back" for "the pin click does nothing at all, silently", which is not obviously the better failure.

So `claim()` is now `queueWrite(token, work)` in `5cb81f69`: writes run in turn on a promise chain, `busy` names only the link that is running, and each link catches its own so a failure cannot break the chain. Nothing is dropped, and nothing overlaps. Every handler is a `queueWrite` call now, `saveNotes` included.

New test, *does not let a notes save put the pre-toggle pins back*, drives your exact sequence: type, blur, click the pin. The mocked hub answers each read with the pins it holds at that moment, so ordering is observable — after the blur only the PATCH is in flight and `setWorkflowPins` has not been called; once it settles the pin write runs and `sampler_name` stays above the fold. Run together, the PATCH answers last carrying `pins: null`, the default set applies and the row drops back inside the fold while the hub holds it pinned. Mutating the chain back to `Promise.resolve().then(...)` turns this test and *runs one write at a time* red.

One note on the draft: `saveNotes` compares `notesDraft` against `detail.notes` at queue time rather than at run time, because the question is "did the person change anything", and by the time an earlier write has finished `detail` may already carry what they typed.

frontend/src/components/panels/WorkflowTab.vue
lindkvis: Agreed, and sequenced in `5cb81f69`. I had reached for `allSettled` to stop the loop aborting on the first refusal and took the concurrency along with it without asking whether I wanted it; I did not, for the reason you give — N PATCHes means N `read_grid()` calls and N announces against a single-writer hub, on a gesture that is not urgent.

It is a plain `for … await` again, but the `try/catch` is inside the loop rather than around it, so every card is still attempted and the message still names how many actually went. Adding a bulk route is a backend change I have left alone: it belongs with the other missing write routes rather than in this PR.

*says how many of a multi-select hide actually went* still holds it — mutating the body back to a loop that rethrows turns it red.

frontend/src/components/panels/WorkflowTab.vue
lindkvis: Agreed, and gone in `5cb81f69` — the element and its rule both. You are right that a bordered 56px box holding `—` reads as broken rather than as absent, and the honest state is that the card payload has no strength at all.

I considered a caption instead and dropped it: one muted "strength not recorded" per slot is noise on a workflow with four LoRAs, and the chip beside it already says which file is in the slot. The comment in the template names `_describe_slots` and `FEATURED_NAMES` so the next person knows what has to land before the field can exist, and `docs/frontend_architecture.md` §F3 says the same.

frontend/src/components/panels/WorkflowTab.vue
lindkvis: Fair, and fixed in `5cb81f69`. `0.04` was mine, not copied from anywhere — all four here and the one in `WorkflowDefaultRow.vue` are now `0.06`, which is what `ProjectFiles.vue` and the others use.

While in there I also replaced the one `0.08` — `.wftab-item:hover` — with `var(--hover-wash)`. That is the token for the case and I should have used it first time.

PR #1446, opened 2026-09-19
Closing a picture opened from a workflow tile goes back to the shelf

lindkvis:
Opening a picture from a "Made with it" tile in the workflow inspector took
you to All Pictures with the lightbox on top, and closing it left you there.
The lightbox is mounted inside `ImageGrid`, which the workflow library
replaces, so the tile has nowhere else to open a picture — but it can say
where to go back to.

## What changed

- The tile pushes `?from=<path>` alongside `?overlay=<id>`.
- `overlayCloseTarget` (`utils/overlayRoute.js`) is the one decision the
  lightbox close makes. It honours a same-document absolute path, carries
  `?token=` with it (a share session's credential, the rule
  `useAppNavigation.withShareToken` states), and otherwise drops `?overlay=`
  and stays put exactly as before.
- `closeOverlay(returnToOrigin = true)`: only the reader's own close goes
  back. The five closes the grid makes on its own behalf — reverse-image
  search, find-similar-faces, delete, scrapheap restore, use-as-input — and
  the route watcher pass `false`, because they close the lightbox *to show
  their result in the grid*, and a gridless screen would be a worse bug than
  the one being fixed.

Tests: the helper, the tile's push, and a mount of the real `ImageGrid` that
pins the close asking the helper. Each assertion was mutated to red and
restored. No changelog fragment: the Workflows view is unreleased, so this
never reached anybody.

## Answering the pre-merge review

Three of its findings I think are wrong, and say so here rather than in
silence:

1. **"`router.back()` does the whole job in one line."** It still needs a
   marker in the query to know the entry was pushed rather than replaced, so
   it saves nothing there — and it fails worse where it fails: on a reloaded
   or pasted `?overlay=` link, back() leaves the app entirely. The cost of
   the chosen shape is one inert Back press (history holds `/workflows`
   twice); I would rather pay that than a navigation off the app.
2. **"Four of seven helper tests guard a threat vue-router forecloses."**
   True that the catch-all would also swallow a foreign `?from=`. The target
   comes from the URL bar and four lines make it safe without depending on a
   redirect rule two files away.
3. **"Delete strands the reader on All Pictures."** Deliberate: the undo for
   the delete you just made is in the grid. Being teleported away from your
   own undo is worse than staying.

Its other findings are fixed in this diff: the dropped share token, the
untested seam (that mount test exists because of it), the "impractical to
mount" justification that was simply false, an overclaim about what comes
back — the selection and the Show filter survive in the store, the scroll
position does not — and one ambiguous term in the architecture doc.


lindkvis:
CI was red on `checks` → **Test Frontend Utilities**, with **264 files and 4529 tests all passing**. Vitest exited 1 on five unhandled `EnvironmentTeardownError: [vitest-worker]: Closing rpc while "onUserConsoleLog" was pending` — three from `ImageGridUseAsInput.test.js`, two from the mount test this PR adds.

**Cause.** jsdom has no canvas, so `getContext()` reports "Not implemented: HTMLCanvasElement's getContext() method…" through its virtual console and returns null. Three components call it (ImageGrid measures label widths, StatsSidebar draws sparklines, ImageOverlay encodes a PNG), so a run emitted ~67 of them. Vitest forwards worker console output to the main process over rpc, and a write still in flight when a fast worker's environment tears down is raised as an unhandled error that exits 1. It lands on whichever files finish quickest, which is why adding an eleventh fast ImageGrid-mounting file tipped it and why develop is green with the same latent race.

**Reproduced locally** by squeezing the two canvas-heavy files onto one core: `taskset -c 0 npx vitest run ImageGridOverlayCloseReturn ImageGridUseAsInput` failed 1 run in 6 with exit 1 and the same two teardown errors.

**Fix** (`frontend/src/testing/setup.js`): stub `HTMLCanvasElement.prototype.getContext` to return `null` — exactly what jsdom returns after reporting it, so every caller's `if (!ctx)` branch behaves as before and only the report is gone. It sits beside the ResizeObserver/IntersectionObserver stubs already there for the same reason. `setup.test.js` pins it on jsdom's virtual console and goes red when the stub is removed.

After: 0 of the 67 notices, 0 teardown errors, exit 0 on three full runs, and 12 squeezed runs of the pair without a single failure. No assertion was changed, loosened or skipped.

frontend/src/testing/setup.js
```diff
@@ -26,3 +26,28 @@ globalThis.IntersectionObserver = NoopObserver;
...
+// the gate goes red or green on the file list rather than on the code: #1446
+// failed with 264 files and 4529 tests all green, while the same tree passed
+// locally.
+//
+// Returning null is exactly what jsdom returns after reporting it, so every
+// caller's `if (!ctx)` branch behaves as before and only the report is gone. A
+// suite that needs a drawable context replaces this for its own lifetime, the
+// way `ImageOverlayMediaActions.test.js` already does.
+if (typeof HTMLCanvasElement !== "undefined") {
+  HTMLCanvasElement.prototype.getContext = function getContext() {
```
lindkvis, line 60: The mechanism is right — jsdom's report → vitest's `onUserConsoleLog` rpc → a write in flight at worker teardown → `EnvironmentTeardownError` and exit 1 with everything green. But this closes about 6% of it, and both the commit message and `setup.test.js` bill it as *the* fix.

Measured on this head, full suite:

```
$ npx vitest run --reporter=default | grep -cE "^(stderr|stdout) \|"
1012
```

1012 forwarded console writes remain **after** the stub. The 67 it removes were one producer among many, and not the largest:

```
69  src/components/settings/LibraryLayoutDialog.test.js
67  src/components/panels/TagPanelKeyboardTrap.test.js
56  src/components/views/ImageGridLibraryEmptyState.test.js
50  src/components/views/ImageOverlayTurnRefresh.test.js
38  src/components/views/ImageGridRotate.test.js
…
```

One test file on its own forwards more than the stub removed. The race is a function of total rpc traffic near teardown, so quieting one source makes the flake rarer and nothing else — as the comment's own "it lands on whichever files finish quickest" concedes.

Two consequences:

* **The class fix is a config line**, not a prototype patch: `test.silent: true` or `onConsoleLog: () => false` stops the forwarding entirely, which is the channel the error is raised on (upgrading past vitest's teardown-rpc bug is the other route). Either closes all 1079; this closes 67.
* **The bigger cost is the write-up.** `setup.test.js`'s header says this stub "is what reds the gate" and that without it "the flake would come back". With 1012 writes still going over that rpc, it will — and the next person will read that file, see the stub intact, and conclude the cause is somewhere else entirely.

Removing the report is worth doing on its own merits; it is genuinely dead noise. The ask is to stop billing it as the fix, and to put the config line in the same PR if the flake matters.

frontend/src/components/views/ImageGrid.vue
```diff
@@ -6501,11 +6516,17 @@ async function openOverlay(img) {
...
 }
 
-function closeOverlay() {
+// `returnToOrigin` false for every close this component makes on its own
+// behalf: a reverse-image search, a delete, a restore and "use as input" all
+// close the lightbox precisely so their result is visible in the grid behind
+// it, and honouring `?from=` would send that result to a screen with no grid.
+// The default is the reader's own close (the X, Escape, the `close` emit, which
+// carries no payload), which is the one that goes back where it came from.
+function closeOverlay(returnToOrigin = true) {
```
lindkvis, line 6553: The template binds this bare — `@close="closeOverlay"` at line 26 — so whatever the `close` emit carries becomes `returnToOrigin`.

Safe today: all three emit sites in `ImageOverlay.vue` (`:18`, `:2337`, `:2462`) are argument-free, so the default arm is taken, and the test here relies on that. But a default-`true` boolean bound bare to an event handler is the classic footgun: the day someone emits `close(reason)`, a truthy payload silently routes the reader to the shelf and a falsy one silently stops it, with nothing red anywhere.

`@close="closeOverlay()"` costs two characters and makes it impossible.

frontend/src/components/panels/WorkflowInspector.vue
```diff
@@ -522,10 +523,29 @@ function thumbUrl(id) {
...
+ *
+ * The path without its query, deliberately. `?topology=` is a one-shot
+ * instruction from a Recipe link that the shelf honours on mount: carrying it
+ * back would re-select whichever workflow that link named, over whichever one
+ * the reader has since chosen.
  */
 function openPicture(id) {
-  router.push({ name: "all-pictures", query: { overlay: String(id) } });
+  router.push({
+    name: "all-pictures",
```
lindkvis, line 546: Small correction to the PR body's "history holds `/workflows` twice — one inert Back press".

The tile `push`es and the close `replace`s, so it is one inert Back **per open/close cycle**, not one total: the push adds an entry, the replace turns that entry into `/workflows`, and the next picture does it again. After five pictures the reader presses Back five times to leave the shelf.

Still the right trade against `router.back()`, and the alternative is one word — `router.replace` here, which keeps history flat at the price of Back-from-the-lightbox going to whatever preceded the shelf rather than to the shelf. Worth stating the real number either way.

frontend/src/utils/overlayRoute.js
```diff
@@ -0,0 +1,41 @@
...
+// lightbox precisely to show their result in the grid behind it, and leaving for
+// the shelf would put that result on a screen that has no grid. Only the
+// reader's own close (the X, Escape) goes back. Either way `from` is dropped: it
+// is one close's instruction, not a sticky property of the route.
+export function overlayCloseTarget(query, returnToOrigin = true) {
+  const { overlay: _overlay, from, ...rest } = query || {};
+  if (
+    returnToOrigin &&
+    typeof from === "string" &&
+    from.startsWith("/") &&
```
lindkvis, line 46: This validates the shape of `from`, not the target. The router's catch-all is `redirect: "/"` (`router/index.js:74`), so `?from=/nonsense` passes both guards, resolves to nothing, and closes the lightbox onto the **home view** — not the "falls back to the plain close" this file's own doc comment promises two paragraphs up.

Only reachable by hand-editing the URL, so not urgent. `router.resolve(from).matched.length` in the caller would close it; if it is not worth closing, the comment should say the fallback is shape-only.

Separately, and just for the record: `?token=` is carried back out here but dropped on the way *in* — `openPicture` uses `router.push` directly rather than `pushAppRoute`/`withShareToken`. Moot today, since `/workflows` is owner-only (`useAppNavigation.js:417`), which also makes this `rest.token` branch unreachable from its only producer. Fine as defence; the pair is just inconsistent if a share-reachable `?from=` producer ever appears.

lindkvis (commented):
Reviewed `origin/develop...fbeb8841`. The routing change is right and the shape is good: `overlayCloseTarget` is one decision in one place, `returnToOrigin` cleanly separates the reader's close from the five the grid makes on its own behalf, and the seam is pinned twice — the helper alone, and a real mount of `ImageGrid` proving the close actually asks it.

`WorkflowInspector.openPicture` is the app's **only** navigation into `?overlay=`, so this fixes the whole class as it exists rather than the reported instance. (`WorkflowTab.vue` from #1445 will want the same `?from=` when F7 wires its "N pictures" link.) `_pushOverlayRoute` spreading the existing query means `?from=` survives arrow-key navigation inside the lightbox, so closing after browsing still returns to the shelf — good, and not obvious.

Green locally on the four touched suites (33) and `npm run lint`.

I agree with all three of the pushbacks in the body, including that `router.back()` fails worse on a pasted link.

**One finding, on the second commit**, plus four smaller notes. Nothing blocking the routing work.

frontend/src/components/views/ImageGrid.vue
lindkvis: Agreed, fixed in ab9ef7f9 — `@close="closeOverlay()"`, with the reason written above the element (a template cannot carry a comment between attributes).

I also found the footgun was not just unguarded but untestable as written: my mount test called `wrapper.vm.closeOverlay()` directly, so reverting the binding to bare left the suite green. There is now a case that emits through the real binding with a payload that would invert the decision:

```js
overlay.vm.$emit("close", false);
expect(replace).toHaveBeenCalledWith({ path: "/workflows", query: {} });
```

Bare binding → red; called → green. Thanks, this one was worth more than two characters.

frontend/src/utils/overlayRoute.js
lindkvis: Right, and worth closing rather than documenting — the failure mode is the exact landing this PR exists to prevent, so a comment saying "shape only" would have been a promise broken in the direction that matters. Fixed in ab9ef7f9.

One correction to the mechanism, though: **`matched.length` does not work.** The catch-all *is* a match, so it reports 1 for everything. Measured against a router built from this app's table:

```
/workflows     name: "workflows"  matched: 1  ["/workflows"]
/nonsense      name: undefined    matched: 1  ["/:pathMatch(.*)*"]   ← would pass the check
/training-runs name: undefined    matched: 1  ["/training-runs"]     (the published redirect)
```

The name is what separates them, and it happens to be exactly the right rule to want: the only unnamed routes in the table are the two redirects, so "`?from=` must name a destination" is both the test and the intent. `overlayCloseTarget` takes an `isDestination` predicate and `ImageGrid` passes `(path) => Boolean(router.resolve(path).name)`, wrapped in try/catch because a malformed percent-escape throws in the resolver rather than returning nothing — and logs when it does, since a link building `?from=` wrongly should not be hidden by the close quietly staying put. Pinned at both levels; both go red when the predicate is dropped.

On the token asymmetry: agreed it is inconsistent, and I am leaving it. `openPicture` carrying a token on an owner-only route would be dead code today, while the helper's branch sits at the shared seam where a future `?from=` producer arrives — that is the end worth defending in advance. If a share-reachable producer ever appears, the carry belongs in it.

frontend/src/components/panels/WorkflowInspector.vue
lindkvis: Correct, and the body now says so: one inert Back **per open/close cycle**, five pictures = five presses. Thanks for doing the arithmetic I waved at.

Keeping `push`. The accumulation is the price of the thing I actually want from it: with `push`, Back *from inside the lightbox* lands on the shelf, which is the same journey this PR is about — a reader who opened a picture from a tile and reaches for Back means "back to where I was", not "back to whatever preceded the shelf". `replace` would flatten history by making that gesture skip the shelf entirely.

Worth saying plainly that both options are worse than they look, and neither is free.

frontend/src/testing/setup.js
lindkvis: You are right, and it is worse than 6% — it still reds the gate, and it named my own file when it did. Reproduced your reading on this head before changing anything:

| measurement | result |
|---|---|
| forwarded console writes, full run | **1012** (your number exactly) |
| the two-file repro, 30 runs squeezed onto one core | **2 failures** |
| plain full suite, 3 runs | **1 failure**, both errors originating in `ImageGridOverlayCloseReturn.test.js` |

So the write-up was not just overclaiming, it was wrong in the direction that matters. Fixed in ab9ef7f9: `setup.js` now carries a `WHAT THIS IS NOT` section with the measured share and the repro command, and `setup.test.js` says the stub removes ~6% of the traffic the flake rides on rather than being the fix. The stub stays on the dead-noise merit you granted it.

**On the config line, I disagree, and I think the mechanism does not hold.** `silent` and `onConsoleLog` are both read in the main process; the worker installs its forwarding console unconditionally (`vitest/dist/chunks/base.*.js` → `createCustomConsole()`, which never consults `silent`), so the rpc call that raises the error has already happened by the time either is consulted. Measured rather than argued: **`--silent=true`, 30 squeezed runs → 1 failure, 3 teardown errors.** It hides the output and leaves the race.

**What I did instead**, bounded by the evidence: the gate failure named two files, and both now forward **zero** console writes. Their Vuetify "Failed to resolve component" warnings were an artefact of a deliberately shallow mount (`isCustomElement` is a compiler option and these templates were compiled by Vite, so it never applied) — dropped via a `warnHandler` that passes every *other* warning through. The `[perf]` line is real `markEnd` behaviour, spied out. Result on this tree:

* 8 full runs: **0 failures, 0 teardown errors**
* the 30-run squeeze that failed 2/30 this morning: **0/30**
* suite-wide writes: 1079 → 1012 (stub) → **965** (these two files)

**And here is the part I am not starting, which is why I am leaving this thread open.** 965 writes across 49 files remain, and the next-fastest noisy file inherits the trigger the moment something shifts — the whack-a-mole you predicted. The real close is the channel: a vitest upgrade past the teardown-rpc bug (a dependency decision, and vitest 5.0.0 is what this repo pins), or a suite-wide console cleanup across those 49 files. Either is a bigger piece of work than "closing the lightbox goes back to the shelf" and belongs to whoever owns that call, not to this card. Happy to open it as its own issue with these measurements if that is useful.

